PR #9, opened 2018-11-28
Add eslint flowtype

### LINKED ISSUE #1: Add flow and eslint

FilipMessa:
resolve #1


.eslintrc.js
```diff
@@ -0,0 +1,453 @@
+const OFF = 0;
+const WARN = 1;
+const ERROR = 2;
+
+module.exports = {
```
tbergq: How about using the new [@kiwicom/eslint-config](https://www.npmjs.com/package/@kiwicom/eslint-config) as base?

.flowconfig
```diff
@@ -0,0 +1,81 @@
...
+suppress_comment=\\(.\\|\n\\)*\\$FlowExpectedError: .+
+
+[lints]
+all=error
+sketchy-null=off
+untyped-import=off
+; unclear types are used by Relay Compiler
+unclear-type=off
+; untyped imports are used by Relay Compiler
+untyped-type-import=off
```
tbergq: could you also add 
`esproposal.optional_chaining=enable`
`esproposal.nullish_coalescing=enable`
`unnecessary-optional-chain=error`

see [mobile-flowconfig](https://github.com/kiwicom/mobile/blob/master/.flowconfig#L55)

pages/_document.js
```diff
@@ -9,31 +11,33 @@ const normalizeNextElements = `
...
-    return { ...page, styles: React.Children.toArray(styles) }
+      getStyleElement(),
+    ];
+    return { ...page, styles: React.Children.toArray(styles) };
   }
 
-  render () {
+  render() {
     return (
       <html style={{ height: '100%' }}>
```
tbergq: Is it possible to use Stylesheet.create here?

.eslintrc.js
FilipMessa: Great, It is finally deployed. Thanks, I will do that

.flowconfig
FilipMessa: Sure

pages/_document.js
FilipMessa: I don't know, let's try.

FilipMessa: Nope, I tried it but it not work. Probably because I tried to apply the react-native style to the DOM element. And I suppose that _document.js use just next.js and react-native don't use it.

PR #20, opened 2018-11-29
setup Test

### LINKED ISSUE #2: Setup Jest

FilipMessa:
closes #2


FilipMessa:
In my current implementation, it works on local, but for some unknown reason, it failed on CI.

tbergq:
CI should pass now 🤞

app/src/components/Text.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+import * as React from 'react';
```
tbergq: Maybe we should have components as a workspace

FilipMessa: We should mainly use the `universal-component` for components, but of course, we will need some margarita's specific. But doesn't this raise the complexity? 

I don't have many experiences with workspaces, only when their start using them on Balkan.

tbergq: Why would it raise complexity? The main benefit is I can easily import it like 
`import { Text } from '@kiwicom/margarita-shared';`

rather than

`import Text from '../../../../components/text/Text';`

FilipMessa: Thanks for the explanation, so yeah it makes sense, to do it as you suggest

FilipMessa: I will do it in this one, it should not be many changes so the code should be still readable for CR.

PR #29, opened 2018-12-03
Graphql: Set up http service

### LINKED ISSUE #11: Connect GraphQL to the Tequila API (HttpRequest service)

tbergq:
closes #11 

It works for web, but somehow times on mobile.

### DIFF
--- a/packages/graphql/index.js
+++ b/packages/graphql/index.js
@@ -18,7 +18,8 @@ const schema = new GraphQLSchema({
 
       hello: {
         type: GraphQLString,
-        resolve: async () => {
+
+        async resolve() {
           const test = await fetch(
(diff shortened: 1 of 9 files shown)


packages/graphql/index.js
```diff
@@ -9,6 +9,8 @@ import {
   GraphQLID,
 } from 'graphql';
 
+import { get } from './src/services/HttpRequest';
```
mrtnzlml: Promise yourself to never ever copy anything from our internal GraphQL. One week old project and you already copy-pasted our technical debt... 🙈🙂

Also, do you need anything from the repo? Let me know please and I can help you with that. Copy-paste doesn't sound like a way to go.

tbergq: In that case, please export these data fetching functions, so we can re use them here 🙂

PR #35, opened 2018-12-07
Graphql: Add flights query

tbergq:


packages/graphql/index.js
```diff
@@ -9,20 +9,21 @@ import {
...
 
 import fetch from './src/services/Fetch';
+import AllFlights from './src/queries/AllFlights';
+import createContext from './src/services/GraphQLContext';
 
 const schema = new GraphQLSchema({
   query: new GraphQLObjectType({
     name: 'RootQueryType',
     fields: {
+      allFlights: AllFlights,
```
michalsanger: I hope it's a right time to discuss some naming conventions. I suggest not to use prefix `all`, we stoped using it on graphql.kiwi.com it became something like "habit you can't explain"

packages/graphql/src/queries/AllFlights.js
```diff
@@ -0,0 +1,51 @@
...
+  +input: {|
+    +flyFrom: string,
+    +dateFrom: Date,
+    +dateTo?: Date,
+  |},
+  ...$Exact<ConnectionArguments>,
+|};
+
+export default {
+  name: 'AllFlights',
```
michalsanger: We should not forget we are not about flights anymore. This query can return also trains and buses, in the future also ferries etc.

Based on [our gossary](https://kiwi.wiki/kiwi/handbook/#/comms/glossary?id=itinerary-parts-examples) this query returns itineraries so shall we named it `itineraries` or `searchItineraries` maybe?

packages/graphql/index.js
FilipMessa: I agree with you prefix **all** is kind of confusing. And based naming conventions on kiwi's glossary make sense. I would like to share this naming convention as a part of our documentation.

tbergq: I agree, I will come up with a better name

packages/graphql/src/queries/AllFlights.js
tbergq: I think itineraries is fine, I will change it

tbergq:
@FilipMessa @michalsanger 🚀

packages/graphql/src/queries/Itineraries.js
```diff
@@ -0,0 +1,51 @@
...
+    },
+    ...connectionArgs,
+  },
+  type: ItinerariesConnection,
+  resolve: async (_: mixed, args: Args, { dataLoader }: GraphqlContextType) => {
+    const {
+      input: { flyFrom, dateFrom, dateTo },
+    } = args;
+    const flights = await dataLoader.itineraries.load({
+      flyFrom,
```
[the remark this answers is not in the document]
tbergq: Yeah, this is actually what the parameter is called when we pass it to tequila, but sure, we could change it too

tbergq:
@FilipMessa 🚀

FilipMessa:
There is still some naming with `flights` so because we assume that in futter it won't be only about flights, we should change all these names.
What do you think?

tbergq:
🙈 you are right, I missed this, I will change it.

tbergq:
@FilipMessa should be no flights left now

FilipMessa:
Thanks, good job.
But it hard to know if it works without Playground or some FE implementation now, so what about adding there some tests at least?

tbergq:
Yes, I wanted to add some tests, but we need some Testing tool, which I think should be another PR, this is big enough already. 

I have an implementation I used to test it which I did not commit. Do you want me to add that?

FilipMessa:
OK, sure make tests in another PR than, I believe you that it works so. I'll approve this PR.

scripts/createSchema.js
```diff
@@ -0,0 +1,11 @@
...
+
+import { schema } from '@kiwicom/margarita-graphql';
+import { printSchema } from 'graphql';
+import fs from 'fs';
+import path from 'path';
+
+fs.writeFileSync(
+  path.join(__dirname, '..', 'schema.graphql'),
+  printSchema(schema),
+);
```
mrtnzlml: 🤔 https://github.com/kiwicom/margarita/blob/06455a179940dc4688029d05a131701eb65bf504/scripts/relay-compiler.js#L20

The reason why I added it together with Relay Compiler is that it always fetches (reads) new schema before Relay works with it so you don't even have to think about such a step. This script should not be necessary.

tbergq: Yes, I was unaware of this when making the PR. I will remove it.

PR #36, opened 2018-12-08
graphql server setup

FilipMessa:


packages/graphql/src/Schema.js
```diff
@@ -0,0 +1,28 @@
...
+import { GraphQLSchema, GraphQLObjectType, GraphQLString } from 'graphql';
+import GlobalID from '@kiwicom/graphql-global-id';
+
+import fetch from './services/Fetch';
+
+const schema = new GraphQLSchema({
+  query: new GraphQLObjectType({
+    name: 'RootQuery',
+    fields: {
+      id: GlobalID(() => 'TODO: implement me'),
```
tbergq: do we really need an id field on the root query?

packages/graphql/src/services/Fetch.js
```diff
@@ -1,8 +1,10 @@
 // @flow
 
 import fetchWithRetries from '@mrtnzlml/fetch';
-import { API_KEY, NODE_ENV } from 'react-native-dotenv';
-import { stringify } from 'querystring';
+
+import type { HttpReqest } from './types';
```
tbergq: `HttpRequest` is the correct spelling, but maybe `HttpMethod` is better naming`

packages/relay/src/Environment.js
```diff
@@ -1,19 +1,7 @@
...
-      );
-    }
-    return inMemoryFetcher(request.text, variables);
-  };
-};
+import { createEnvironment, createNetworkFetcher } from '@mrtnzlml/relay';
 
 export default createEnvironment({
-  fetcherFn: createInMemoryFetcher(),
+  fetcherFn: createNetworkFetcher('http://127.0.0.1:4000'),
```
tbergq: Maybe this should point to an ENV variable instead? If we were to deploy this to a server, it would not work

scripts/relay-compiler.js
```diff
@@ -7,7 +7,7 @@ const path = require('path');
 const { printSchema } = require('graphql/utilities');
 const { spawnSync } = require('child_process');
 
-const { schema } = require('../packages/graphql/index');
+const schema = require('../packages/graphql/src/Schema');
```
tbergq: const { schema } = require ('@kiwicom/margarita-graphql')

README.md
```diff
@@ -27,7 +29,7 @@ The other available scripts are:
 - `yarn lint` to ensure your code is properly formatted;
 - `yarn flow` to ensure your code is respecting the types given to your vaiables;
 - `yarn test-ci` to ensure your code should pass the Continuous Integration (CI) tests;
-- `yarn relay` to get the latest schema of the [graphql server](packages/graphql) 
+- `yarn relay` to get the latest schema of the [graphql server](packages/graphql)
```
tbergq: does not yarn relay run the relay-compiler? 
yarn graphql to get the latest schema?

packages/graphql/index.js
```diff
@@ -1,42 +1,12 @@
...
+import schema from './src/Schema';
 
-const inMemoryFetcher = (source: string, variableValues: {||}) => {
-  const context = createContext();
-  return graphql(schema, source, undefined, context, variableValues);
-};
+const server = new ApolloServer({ schema, context: createContext() });
 
-export { schema, inMemoryFetcher };
+export { schema, server };
```
tbergq: I am not so sure we should export server from this file. No other app will use it. I think you should move the creation of the server to the server.js file. 

And this file only exports what other packages may use.

README.md
FilipMessa: I looked at it and it does both things, it generates new schema and then runs relay-compiler.
So It looks like we don't need the 'yarn schema' command.

packages/graphql/index.js
FilipMessa: make sense, I'll do how you suggested thanks.

FilipMessa: done

packages/graphql/server.js
```diff
@@ -0,0 +1,14 @@
...
+import { ApolloServer } from 'apollo-server';
+
+import createContext from './src/services/GraphQLContext';
+import schema from './src/Schema';
+
+const server = new ApolloServer({ schema, context: createContext() });
+
+server.listen().then(({ url, context }) => {
+  console.log(`🚀 Server ready at ${url}`); // eslint-disable-line no-console
+});
```
mrtnzlml: Why are you starting the server? Is there something wrong with the previous (literally) server-less approach? :-/

michalsanger: IMHO the ability to use tools like GraphiQL

mrtnzlml: GraphiQL doesn't need the network to get GraphQL schema. 🤔

tbergq: Well, out of curiosity, I would like to know how to use GraphiQL without a server running, because normally, I feed it an url. Not sure how to use it without an url to feed it.

FilipMessa: I tried it run with Playground React Component but I didn't find information in the documentation how to do it without the server.
So I decided to make graphql server. 

I from start expected that there will be the point when we need to split the app.

mrtnzlml: @tbergq You have to use `schema` property and not `fetcher` property.

> `schema`: a GraphQLSchema instance or `null` if one is not to be used. If `undefined` is provided, GraphiQL will send an introspection query using the fetcher to produce a schema.

https://github.com/graphql/graphiql#usage

> I from start expected that there will be the point when we need to split the app.

Yes, this is how I designed this _server-less_ GraphQL as well. Just a question - got the answer, let's move on... :)

PR #41, opened 2018-12-10
App workspace

tbergq:


tbergq:
I think @RobinCsl's approach is better, so I will close this one

PR #42, opened 2018-12-10
Add: Navigation 🚀

RobinCsl:
This is a proposition for our navigation and folder structure.

It is to be discussed whether we want to have a distinction between `apps` and `packages`, maybe it does not make sense.

The idea is to show screens in the platforms' expected "native" way:
- `react-navigation` screens for `native` app;
- file-based paths for Nextjs.

The views are abstracted inside `apps/core`. To navigate between screens, we would use the same API as `react-navigation`, i.e. we would inject/pass a `navigation` prop to components that need to be connected.
For Nextjs, we somehow pass a `navigation` prop for which we need to map the necessary bits. At minimum, we can have:
```
<Component navigation={{navigate: (url) => Router.push(`/${url}`)}} />
```

This will need to be abstracted further in a different PR.

To run the apps, you now need to:
- `cd apps/native; yarn ios`;
- `cd apps/web; yarn dev`;

- [x] Remove unnecessary stuff from workspace `package.json`
- [x] Add documentation


apps/web/pages/index.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import Router from 'next/router';
+import { Search } from '@kiwicom/margarita-core';
+
+export default () => (
+  <View>
+    <Search navigation={{ navigate: url => Router.push(`/${url}`) }} />
```
tbergq: it would be nice to have some more re-usable way of injecting this prop. 

`react-navigation` has the `withNavigation` HOC, so we can easily use navigation from anywhere without props-drilling the navigation down to the component that uses it. 

Would it be possible also a component that makes it work also for web?

Like withNavigation.web.js and withNavigation.native.js?

package.json
```diff
@@ -4,7 +4,7 @@
   "scripts": {
     "ios": "expo start --ios",
     "android": "expo start --android",
-    "dev": "BABEL_ENV=web next",
+    "dev": "BABEL_ENV=web next ./apps/web --config ./next.config.js",
```
tbergq: BABEL_ENV=web should no longer be necessary

apps/web/pages/index.js
RobinCsl: Yes, that was my plan eventually. I think I will add this in a different PR. What do you think?

tbergq: Sure, I am ok with that. We should create an issue for it.

RobinCsl: https://github.com/kiwicom/margarita/issues/50

RobinCsl:
@tbergq @FilipMessa Should be good to go now 🚀

tbergq:
@FilipMessa please review this task also. I think we should try to have it merged as soon as possible.

FilipMessa:
@tbergq It'll be done in a minute

PR #54, opened 2018-12-12
Simple SearchForm implementation

### LINKED ISSUE #45: Create the simple SearchFrom and connect it to the relay.
Prepare some simple the SearchForm with the list of results and connect it to the GraphQL with Relay.  Don't care about design in this step.

-  [x] - user can fill the form 
-  [x] - user can submit a search
-  [x] - user receives results on the new page.

JoseDuSV:
Simple search form implementation which redirects (with parameters) to results page and renders results in simple list.
Parameters passing part is temporary and will be updated after navigation expansion.

Closes #45


apps/core/Search.js
```diff
@@ -1,58 +1,82 @@
...
-import { StyleSheet, View, Button } from 'react-native';
-import { QueryRenderer, graphql } from '@kiwicom/margarita-relay';
-import { Text } from '@kiwicom/margarita-components';
+import { StyleSheet, View, Button, TextInput } from 'react-native';
 
 import Routes from './config/routes';
-import { type SearchQueryResponse } from './__generated__/SearchQuery.graphql';
 
 type Props = Object;
+type State = Object;
```
RobinCsl: We will need eventually to make these types more precise.

packages/graphql/src/types/output/Itinerary.js
```diff
@@ -11,6 +11,11 @@ export default new GraphQLObjectType({
   name: 'Itinerary',
   fields: {
     id: GlobalID(({ id }) => id),
+    price: { type: GraphQLInt },
```
RobinCsl: We will probably need to change this type for price as there would be currency added. @FilipMessa how do we want to deal with breaking changes in the GQL server? Do we want backwards compatibility? 
By using semver, we could always decide to do breaking changes and bump version?

apps/core/Results.js
```diff
@@ -0,0 +1,88 @@
...
+        style={styles.listItem}
+      >{`${flyFrom} / ${localDeparture} -> ${flyTo} / ${localArrival}  | Price: ${price}`}</Text>
+    );
+  };
+
+  keyExtractor = (item: Object, index: number) => item.node.id;
+
+  renderInner = (props: Object) => {
+    const { searchItineraries } = props;
+    const resultsLength = searchItineraries.edges.length;
```
tbergq: edges could be undefined, then this line would crash. Please use type generated by relay like [this](https://github.com/kiwicom/mobile/blob/a325684cc5e2c15dc24e00ecfb60d1d301003c7f/app/MMB/src/scenes/list/cityImage/FromToRow.js#L12)

apps/core/Search.js
```diff
@@ -1,58 +1,82 @@
...
-  goToResults = () => {
-    this.props.navigation.navigate(Routes.RESULTS);
+export default class Search extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      travelFrom: 'OSL',
+      travelTo: 'PRG',
+      dateFrom: '2018-12-30',
+      dateTo: '2018-12-31',
```
tbergq: Why are these values here?

packages/graphql/src/types/output/Itinerary.js
```diff
@@ -11,6 +11,11 @@ export default new GraphQLObjectType({
...
+    flyFrom: { type: GraphQLString },
+    flyTo: { type: GraphQLString },
+    localDeparture: { type: GraphQLString },
+    localArrival: { type: GraphQLString },
```
tbergq: maybe we want some reusable type like 
[RouteStop](https://gitlab.skypicker.com/graphql/graphql/blob/master/src/apps/flight/types/outputs/RouteStop.js)

packages/graphql/src/dataloaders/Itinerariesloader.js
```diff
@@ -23,11 +23,23 @@ export type ItinerariesSearchParameters = {|
...
 |};
 
 type ApiResponse = {|
   +data: $ReadOnlyArray<{|
     +id: string,
+    +airlines: Array<string>,
+    +price: number,
+    +flyFrom: string,
+    +flyTo: string,
+    +local_departure: string,
```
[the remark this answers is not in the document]
tbergq: Unfortunately yes

packages/graphql/src/types/output/Itinerary.js
tbergq: How about `@kiwicom/graphql-bc-checker`

apps/core/Search.js
JoseDuSV: Definitely, currently I treat it as simplified version which will be greatly expanded in next iteration. When most of the states and props will be expanded and precisely defined.

apps/web/pages/index.js
```diff
@@ -9,7 +9,16 @@ import { Layout } from '@kiwicom/margarita-components';
...
   <Layout>
     <View>
-      <Search navigation={{ navigate: url => Router.push(`/${url}`) }} />
+      <Search
+        navigation={{
+          navigate: (url, params) =>
+            // $FlowFixMe - wrong definition in flow-typed
+            Router.push({
+              pathname: `/${url}`,
+              query: params,
```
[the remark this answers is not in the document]
JoseDuSV: For now, I don't think so.

apps/core/Search.js
RobinCsl: It's probably because we don't have the PlacePicker and DatePicker yet, so at least we can see how the app works?

JoseDuSV: They are currently here as sample default values with plan to remove them in next iteration when adding proper inputs and look.
But I can definitely remove them even now.

apps/core/Results.js
```diff
@@ -0,0 +1,88 @@
...
+  resultItem = ({ item }: Object) => {
+    const { flyFrom, flyTo, localDeparture, localArrival, price } = item.node;
+    return (
+      <Text
+        style={styles.listItem}
+      >{`${flyFrom} / ${localDeparture} -> ${flyTo} / ${localArrival}  | Price: ${price}`}</Text>
+    );
+  };
+
+  keyExtractor = (item: Object, index: number) => item.node.id;
```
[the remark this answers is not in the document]
tbergq: That should be no question, `idx` will not report usage where not needed, while  optional-chaining does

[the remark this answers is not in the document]
FilipMessa: Good to know. I didn't yet work with `optional-chaining`.
So, in that case, it is clear choice.

packages/graphql/src/types/output/Itinerary.js
mrtnzlml: I would recommend you to not deal with breaking changes. It was in my opinion mistake to turn this GraphQL into the server because of issues like this and because of the fact you will have to deploy and keep it running (complexity of this approach is much higher). But honestly, you are your own consumers so please **do** breaking changes as needed. At least for now. 😊 

Eventually it would be great to use `@kiwicom/graphql-bc-checker` since it supports logging of breaking changes as well as detection of manual changes. No GitHub fancy integration though. At least not yet.

apps/core/Results.js
```diff
@@ -0,0 +1,88 @@
+// @flow
+
+import * as React from 'react';
+import { StyleSheet, View, FlatList, Text } from 'react-native';
+import { QueryRenderer, graphql } from '@kiwicom/margarita-relay';
+
+import EmptyResults from './EmptyResults';
+
+type Props = Object;
```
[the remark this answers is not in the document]
tbergq: > we should use generated flowtypes from realy

Absolutely, but type Props -> extends React.Component<Props> are the props that are being passed to the component from a parent component, and not from relay in this case. 

For the renderInner we should use generated types from relay

packages/graphql/src/types/output/Itinerary.js
tbergq: Yeah, for me the problem with the inMemory fetcher was the lack of graphiql, but now I know we can have even that. 

Another concern was the bundle size, but I don't think we need to care about it  in this project.

mrtnzlml: Not at this stage - make it work first. This is why it was like it was - just to iterate super fast without server needed (+ with the ability to very easily create a server from it). 🙃

FilipMessa: Yes, I understand your idea but we currently have the server and there is no reason to return back to the serverless state. 

> But honestly, you are your own consumers so please do breaking changes as needed. At least for now. 😊

this is good point  💯 , thanks @mrtnzlml

apps/core/Results.js
```diff
@@ -0,0 +1,121 @@
...
+      !searchItineraries ||
+      !searchItineraries.edges ||
+      searchItineraries.edges.length === 0
+    ) {
+      return <EmptyResults />;
+    }
+    return (
+      <>
+        <Text>{`Total results: ${searchItineraries.edges.length}`}</Text>
+        <FlatList
```
tbergq: I think we should move this to a separate component, so renderInner just returns something like `<SearchResultList data={props.searchItineraries} />`

apps/core/Search.js
JoseDuSV: For now I just at least updated state props and after we update navigation, I'll define even props more precisely in next iteration.

apps/core/Results.js
```diff
@@ -0,0 +1,121 @@
...
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+  },
+  listItem: {
+    fontSize: 12,
+    padding: 2,
+    margin: 2,
+    backgroundColor: '#dddddd',
```
tbergq: Also, this should probably be a value from orbit-tokens, possibly `$backgroundBody`

apps/core/Search.js
```diff
@@ -2,57 +2,87 @@
 
 import * as React from 'react';
 import { View, Button } from 'react-native';
```
tbergq: Please use button from universal-components instead

apps/core/Results.js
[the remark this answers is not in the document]
JoseDuSV: Updated. Components props are defined and generated types are used for renderInner

apps/core/Results.js
```diff
@@ -0,0 +1,88 @@
...
+import * as React from 'react';
+import { StyleSheet, View, FlatList, Text } from 'react-native';
+import { QueryRenderer, graphql } from '@kiwicom/margarita-relay';
+
+import EmptyResults from './EmptyResults';
+
+type Props = Object;
+
+export default class Results extends React.Component<Props> {
+  resultItem = ({ item }: Object) => {
```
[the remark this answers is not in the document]
JoseDuSV: Updated. Type defined.

apps/core/Results.js
JoseDuSV: Types and `EmptyResults` condition updated.

apps/core/Results.js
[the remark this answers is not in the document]
JoseDuSV: `optional-chaining` added.

apps/core/Results.js
JoseDuSV: Definitely. But I think breaking results to separate components and make results lits better in general is for separate issue.

tbergq: I strongly disagree. This component has way too many responsibilities, and one obvious flaw is the copying of the type `ItemProps`, what when this query changes, we have to maintain this type all the time. 

The main problem here is that this PR is way too big, and you should in the future try to have smaller PR.

apps/core/Search.js
JoseDuSV: Sure. I have a plan to swap it in next issue. Also with styling inputs according to final design and with implementation of trip and passengers selectors.

apps/core/Results.js
JoseDuSV: Ok, swapped for `orbit-tokens` value for now.

apps/core/scenes/results/Results.js
```diff
@@ -0,0 +1,66 @@
...
+|};
+
+type ResultsType = {|
+  +searchItineraries?: ResultsListType,
+|};
+
+export default class Results extends React.Component<Props> {
+  renderInner = (props: ResultsType) => {
+    const { searchItineraries } = props;
+    if (searchItineraries) {
```
tbergq: searchItineraries could be defined, but edges could be 0. This logic should be in the ResultsList component

apps/core/Results.js
JoseDuSV: You're right. The borderline for this issue was little bit unclear and it will be better to split big pieces like this into smaller ones next time. 
Results are now separated into smaller components.

apps/core/scenes/results/Results.js
JoseDuSV: Right, moved condition into ResultsList.

packages/graphql/src/types/output/Itinerary.js
FilipMessa: We handle this as the separate issue.

apps/core/scenes/results/Results.js
```diff
@@ -0,0 +1,62 @@
...
+  +dateTo: string,
+|};
+
+type ResultsType = {|
+  +searchItineraries: ResultsListType,
+|};
+
+export default class Results extends React.Component<Props> {
+  renderInner = (props: ResultsType) => {
+    const { searchItineraries } = props;
```
FilipMessa: You can handle it in the shortened version:
```
 renderInner = ( { searchItineraries }: ResultsType) => {
....
```
but it is just an idea., nothing big deal.

apps/mobile/src/screens/ResultsScreen.js
```diff
@@ -1,21 +1,16 @@
...
 
 import * as React from 'react';
-import {
-  AnywhereResults,
-  EmptyResults,
-  ResultsList,
-} from '@kiwicom/margarita-core';
+import { Results } from '@kiwicom/margarita-core';
 
 type Props = Object;
```
FilipMessa: I know that it is not part of your PR but is possible write this types properly?

JoseDuSV: Ok, types updated. So it's currently more clear which parameters component expects.
Later with navigation update this type will be probably updated.

FilipMessa: thanks

PR #62, opened 2018-12-13
Enable watch option for relay-compiler

tbergq:


FilipMessa:
Is this work for you? I try to run `yarn relay --watch` or and it always ends.`

tbergq:
Yes, it does work for me
screenshot 2018-12-13 at 13 40 44

Issue #50, opened 2018-12-11
Create withNavigation HoC to wrap Nextjs Router as react-navigation


RobinCsl:

RobinCsl:
Together with issue #84 and a proposed solution in #87, I believe that issue could be closed after the solution is merged.

PR #69, opened 2018-12-14
Add: margarita-navigation package

RobinCsl:
This is an attempt at #50, creating an HoC is difficult as `Platform.select` did not seem to work as intended in my previous attempt.
What do you think about this approach?

Basically, I make `Navigation` a non-exact object type which is compatible with the Navigation type from `react-navigation`.


tbergq:
> Platform.select did not seem to work

what about just using .web extension?

PR #76, opened 2018-12-18
Add: Trip type select

### LINKED ISSUE #31: Create the Select Component
## Select Component

- [ ] Option to set the default value
- [ ] Only one value can be choose
- [ ] A user can choose between one way or return flight type.
- [ ] One option should be always selected

*behavior*:
After the user clicks on the InputField, modal with options will be open (animation from the bottom - can be avoided in the first implementation).

screen shot 2018-12-04 at 13 40 02

JoseDuSV:
Trip type `Select` and `Modal` implementation
Closes #31


apps/core/scenes/search/Search.js
```diff
@@ -55,42 +78,114 @@ export default class Search extends React.Component<Props, State> {
...
+
+  handlePassengersPress = () => {
+    this.setState({ modalType: MODAL_TYPE.PASSENGERS });
+  };
+
+  handleModalClose = () => {
+    this.setState({ modalType: MODAL_TYPE.HIDDEN });
+  };
+
+  getModalContent = () => {
```
FilipMessa: what about `renderModalContent` ?

packages/components/src/modal/ModalContent.js
```diff
@@ -0,0 +1,65 @@
...
+
+import * as React from 'react';
+import { View, Dimensions, Platform } from 'react-native';
+import { StyleSheet } from '@kiwicom/universal-components';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import type { ModalContentProps } from './ModalTypes';
+import TouchableWithoutFeedback from '../TouchableWithoutFeedback';
+
+class ModalContent extends React.Component<ModalContentProps> {
```
FilipMessa: The **ModalContent** can be implemented like the functional component instead class.

packages/components/src/select/__tests__/Select.test.js
```diff
@@ -0,0 +1,30 @@
...
+    shallow(
+      <Select
+        optionsData={optionsData}
+        selectedType={Object.keys(optionsData)[0]}
+        onSelect={jest.fn()}
+        onClosePress={jest.fn()}
+      />,
+    ),
+  ).toMatchSnapshot();
+});
```
FilipMessa: Nice, but can you add also test for onSelect (callback)? Test if onSelect will be call after the click and if is called with the proper value.
Als0 simple test fo `onClosePress()`

FilipMessa:
And one small thing 
screen shot 2018-12-18 at 10 00 30 The icon inside the circle should be centered

anyway nice job thanks

apps/core/scenes/search/Search.js
JoseDuSV: Thanks, make sense, `render` is definitely more consistent.
Updated

packages/components/src/modal/ModalContent.js
JoseDuSV: Good catch. Updated

packages/components/src/select/__tests__/Select.test.js
JoseDuSV: Ok, both tests added.

apps/core/scenes/search/Search.js
```diff
@@ -55,42 +78,114 @@ export default class Search extends React.Component<Props, State> {
     console.warn('TODO'); // eslint-disable-line no-console
   };
 
+  handleTripTypeSelect = (type: string) => {
```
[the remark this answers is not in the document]
JoseDuSV: > And one small thing
> screen shot 2018-12-18 at 10 00 30 The icon inside the circle should be centered
> 
> anyway nice job thanks

It was caused by icon rendering update. For now I created separate issue for it https://github.com/kiwicom/universal-components/issues/136

Issue #79, opened 2018-12-19
Update documentation


FilipMessa:
How do set `environment` variables?
Where can find `apikey`?

PR #81, opened 2018-12-20
Update README.md and .env-sample (#79)

jvaclavik:


README.md
```diff
@@ -10,33 +11,57 @@ This project is a living example of a client app you could start building with T
 To get started, clone the repository, change directory and install the dependencies:
 
 ```shell
-git clone git@github.com:kiwicom/margarita.git
+https://github.com/kiwicom/margarita
```
JoseDuSV: Why not `git clone git@github.com:kiwicom/margarita.git` ?

jvaclavik: `https` doesn't need any authentication which seems to me better in this case. But it's not a big deal, we can keep the old one.

JoseDuSV: Ok, thanks, that makes sense. But then it should be at least `git clone https://github.com/kiwicom/margarita`

jvaclavik: I see, you are right! Thank you. It's copy-paste issue

jvaclavik: Now it should be fixed

PR #86, opened 2018-12-27
Improve: flags in test script

RobinCsl:
Following suggestion from https://github.com/kiwicom/margarita/pull/83#discussion_r244156449

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,8 @@
     "relay": "yarn babel-node --config=babel.config.js scripts/relay-compiler.js",
     "lint": "eslint . --report-unused-disable-directives",
     "flow": "flow --include-warnings",
-    "test": "node node_modules/jest/bin/jest.js --config=.jest.json --silent",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -12,8 +12,8 @@
     "relay": "yarn babel-node --config=babel.config.js scripts/relay-compiler.js",
     "lint": "eslint . --report-unused-disable-directives",
     "flow": "flow --include-warnings",
-    "test": "node node_modules/jest/bin/jest.js --config=.jest.json --silent",
-    "test-ci": "yarn flow && yarn lint && yarn test --runInBand",
+    "test": "node node_modules/jest/bin/jest.js --config=.jest.json --silent --ci --color",
+    "test-ci": "yarn flow && yarn lint && yarn test",
```
mrtnzlml: I am missing here `--maxWorkers` (because otherwise, you are going to have the same problem like before) and I believe the `--ci --color` should be only for CI. Also, I would remove the `--silent` option completely.

Issue #26, opened 2018-12-02
List of bookings results


FilipMessa:
# List of Results

- [x] Style one way flight (#90)
- [x] Style return flights (#90)
- [x] Update price structure  (#90)
from:
    ```
    ResultsListItem on Itinerary {
      currency
      price
      localDeparture
      ...
    }
    ```
    to:
    ```
    ResultsListItem on Itinerary {
      price {
        value
        currency
      }
      localDeparture
      ...
    }
    ```
- [x] Create timeline image in `ConnectionCard` as RN elements, not as image 
- [x] Style loading (#90)
- [x] 8) Polish styling of `ConnectionCard` in Universal components. Date should be bold, use different fonts, dotted line, and use cards with different background. (#90)
- [x] handle not found/error states (#126)
- [x] Set proper `locale` (e.g `en-US`) according to mobile device to price settings (`ResultsListItem.js`). Probably using by `Expo.Util.getCurrentLocaleAsync` (https://docs.expo.io/versions/v25.0.0/sdk/util)
- [ ] ~~Use `Card` component~~ (we don't need it in othere component for now)
- [ ] ~~Get name of the airline in `ResultsListItem.js`.~~ (not important, we don't show it anywhere)
- [ ] ~~`TripSegmentProps` and `CarrierData` should be exportable from Universal components~~ ( not part of universal-components anymore)
- [ ] ~~Separate `getDateTimeInterval` and `intervalToString` to utils~~ (should be refactored in #123 )
- [ ] ~~Handle cases when we don't have some data from GraphQL/REST server (`getRouteItem` in `RestultsListItem.js`)~~ (part of #123 )

- [x] Show and style header title (OSL -> PRG) on the web (#90)

<img width="341" alt="screen shot 2018-12-11 at 07 08 37" src="https://user-images.githubusercontent.com/6354326/49781603-b0f74080-fd13-11e8-8aa9-432aba7610a4.png">
<img width="329" alt="screen shot 2018-12-11 at 07 08 28" src="https://user-images.githubusercontent.com/6354326/49781604-b0f74080-fd13-11e8-91fe-fb16af13f967.png">

FilipMessa:
Wait until issue(create trip component) on universal will be done.https://github.com/kiwicom/universal-components/issues/92

PR #90, opened 2019-01-02
Feat: Implement ConnectedCard in flight results (#26)

jvaclavik:
screenshot 2019-01-07 at 15 57 23

### Done

- [x] Style one way flight
- [x] Style return flights
- [x] Update price structure 
from:
    ```
    ResultsListItem on Itinerary {
      currency
      price
      localDeparture
      ...
    }
    ```
    to:
    ```
    ResultsListItem on Itinerary {
      price {
        value
        currency
      }
      localDeparture
      ...
    }
    ```
- [x] Create timeline image in `ConnectionCard` as RN elements, not as image 
- [x] Style loading
- [x] 8) Polish styling of `ConnectionCard` in Universal components. Date should be bold, use different fonts, dotted line, and use cards with different background.


apps/core/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "dependencies": {
     "@kiwicom/orbit-design-tokens": "^0.2.5",
-    "@kiwicom/universal-components": "^0.0.6",
+    "@kiwicom/universal-components": "^0.0.7-alpha.5",
```
jvaclavik: `ConnectedCard` component is available from version `0.0.7`, but there is some bug with `DEVICE_LOCALE` in `node_modules/@kiwicom/universal-components/lib/native/utils/DateUtils/DateUtils.js`. This issue is also connected with test failures. 

What can we do with that?

jvaclavik: Fixed in new version `0.0.7`

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -11,51 +13,234 @@ type Props = {|
...
+      return null;
+    }
+    // @TODO polish awful condition if it's possible
+    return route.findIndex(
+      routeItem =>
+        routeItem &&
+        routes &&
+        routes.length === 2 &&
+        routes[1] &&
+        routeItem.flyFrom === routes[1][0],
```
tbergq: You can probably use optional chaining here
```
routeItem => routeItem?.flyFrom === routes?.[1]?.[0]
```

Should be the same (I think)

apps/graphql/src/types/output/Itinerary.js
```diff
@@ -1,21 +1,48 @@
...
+    cityTo: { type: GraphQLString },
+    flyFrom: { type: GraphQLString },
+    flyTo: { type: GraphQLString },
+    id: { type: GraphQLString },
+    localArrival: { type: GraphQLDateTime },
+    utcArrival: { type: GraphQLDateTime },
+    localDeparture: {
+      type: GraphQLDateTime,
+    },
+    utcDeparture: { type: GraphQLDateTime },
```
tbergq: There is quite a lot of from and to going on here. Maybe we should add
Arrival and Departure object. 

```
name: 'Route',
  fields: {
	arrival: {type: SomeReusableType },
	departure: {type: SomeRuesableType }
  }

name: 'SomeRuesableType', 
fields: {
	city: {type: string },
	localTime: {type: GraphqlDate},
	utcTime: {type: GraphqlDate},
}
```

Do you see the point?

package.json
```diff
@@ -25,6 +25,8 @@
     "@kiwicom/margarita-relay": "^0",
     "better-npm-run": "^0.1.1",
     "expo": "^31.0.6",
+    "graphql-iso-date": "^3.6.1",
+    "luxon": "^1.8.2",
```
tbergq: Why is this dependency moved here?

apps/graphql/src/types/output/Itinerary.js
```diff
@@ -1,21 +1,48 @@
...
+  },
+});
+
 export default new GraphQLObjectType({
   name: 'Itinerary',
   fields: {
+    currency: { type: GraphQLString },
```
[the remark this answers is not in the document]
RobinCsl: I would use `amount` instead of `value`.

RobinCsl:
Also, bear in mind that `ConnectionCard` is moved to Margarita, once that PR is merged:  https://github.com/kiwicom/margarita/pull/97

apps/core/scenes/results/ResultsListItem.js
jvaclavik: Wow 👍 what is this syntax? I haven't seen it before.

RobinCsl: It's called optional chaining
https://github.com/tc39/proposal-optional-chaining

apps/graphql/src/types/output/Itinerary.js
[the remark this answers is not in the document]
FilipMessa: definitely, `amount` is the better name.

package.json
jvaclavik: We use luxon in the `core` and also in the `graphql` directories. `graphql-iso-date` shouldn't be here, because it's only in `graphql`, so I will change it.

apps/graphql/src/dataloaders/Itinerariesloader.js
```diff
@@ -48,23 +80,24 @@ const parseDate = (date: Date) =>
...
-      returnTo: parseDate(input.returnDateTo),
+    ...(input.dateTo && {
+      returnTo: parseDate(input.dateTo),
     }),
     ...(input.passengers && {
       adults: input.passengers.adults ?? 0,
       children: input.passengers.children ?? 0,
       infants: input.passengers.infants ?? 0,
     }),
+    limit: 10, // @TODO: for testing purposes
```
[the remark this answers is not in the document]
jvaclavik: It's not important to have it here. I guess we can delete it completely and we can manage it when we will implement some pagination.

apps/graphql/src/types/output/Itinerary.js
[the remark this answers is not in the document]
jvaclavik: OK, from my perspective `amount` is better than `value`, but in this case we should also update it in `Price` from the `universal-components`

apps/graphql/src/types/output/Itinerary.js
jvaclavik: That's good idea, thanks!

jvaclavik: Done

jvaclavik:
This PR should be merged after the release of `universal-components@0.0.8` and update of `ConnectionCard` for getting proper `carrier` and `price`.

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,249 @@
...
+      const localArrival = DateTime.fromISO(route.arrival?.localTime);
+      const localDeparture = DateTime.fromISO(route.departure?.localTime);
+      const utcArrival = DateTime.fromISO(route.arrival?.utcTime);
+      const utcDeparture = DateTime.fromISO(route.departure?.utcTime);
+      return {
+        arrival: route.arrival?.city ?? '',
+        arrivalTime:
+          (localArrival && localArrival.toLocaleString(DateTime.TIME_SIMPLE)) ??
+          '',
+        carrier: { code: route.airline ?? '', name: '' }, // @TODO: get name of the airline
```
FilipMessa: Are there missing data for airline name in API or ?

apps/graphql/src/types/output/Itinerary.js
```diff
@@ -1,21 +1,57 @@
...
     id: GlobalID(({ id }) => id),
-    price: { type: GraphQLInt },
     flyFrom: { type: GraphQLString },
     flyTo: { type: GraphQLString },
     localDeparture: { type: GraphQLString },
     localArrival: { type: GraphQLString },
+    price: { type: priceType },
+    route: { type: new GraphQLList(routeType) },
+    routes: { type: new GraphQLList(new GraphQLList(GraphQLString)) },
     // TODO: Add fields as needed
```
FilipMessa: can you remove this comment pls? I think this is not necessary it is obviously

package.json
FilipMessa: I suppose that this is the right place for `luxon`, because we want to work with the same version in all workspaces.

apps/graphql/src/dataloaders/__tests__/Itinerariesloader.test.js
```diff
@@ -11,17 +11,19 @@ const searchParams = {
 it('parses parameters corectly', () => {
   expect(parseParameters(searchParams)).toMatchInlineSnapshot(`
 Object {
+  "curr": "EUR",
   "dateFrom": "01/01/2018",
   "dateTo": "01/01/2018",
   "flyFrom": "OSL",
+  "limit": 10,
```
FilipMessa: we don't need this `limit` parameter here now, right?

apps/graphql/src/dataloaders/__tests__/Itinerariesloader.test.js
```diff
@@ -30,12 +32,12 @@ Object {
...
   "adults": 2,
   "children": 0,
+  "curr": "EUR",
   "dateFrom": "01/01/2018",
   "dateTo": "01/01/2018",
   "flyFrom": "OSL",
   "infants": 0,
-  "returnFrom": "02/01/2018",
-  "returnTo": "10/01/2018",
+  "limit": 10,
```
FilipMessa: 🔝

apps/core/scenes/results/ResultsListItem.js
jvaclavik: Exactly

apps/core/scenes/results/Results.js
```diff
@@ -4,6 +4,8 @@ import * as React from 'react';
 import { View } from 'react-native';
 import { QueryRenderer, graphql } from '@kiwicom/margarita-relay';
 import { StyleSheet, Text } from '@kiwicom/universal-components';
+import { DateTime } from 'luxon';
```
RobinCsl: As mentioned in @JoseDuSV 's PR, we'll need to get rid of `luxon`; we will need to stop PRs using `luxon` after this one is landed.

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,246 @@
...
+      return route.map((routeItem: RouteItem) =>
+        this.getSanitizedRouteItem(routeItem),
+      );
+    }
+    return [];
+  };
+
+  getSector = (
+    route: Array<TripSectorWithId>, // rename to sanitizeRoute
+    demandedSector: number,
```
RobinCsl: What do you mean by `demanded`?

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,246 @@
...
+    sectorBorderIndex: ?number,
+  ): Array<TripSectorWithId> => {
+    if (
+      demandedSector === null ||
```
RobinCsl: Could you give names to these conditions for the logic to be easier to follow?

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,246 @@
...
+  };
+
+  getDurationBetweenSectors = (route: ?Route, sectorBorderIndex: ?number) => {
+    if (route && sectorBorderIndex != null && sectorBorderIndex > 0) {
+      const stopBeforeSector = route[sectorBorderIndex - 1];
+      const stopAfterSector = route[sectorBorderIndex];
+
+      const utcDeparture =
+        stopAfterSector &&
+        stopAfterSector.departure &&
```
RobinCsl: I guess you could shorten it to `stopAfterSector.?departure && ...`

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,246 @@
...
+        DateTime.fromISO(stopAfterSector.departure.utcTime);
+      const utcArrival =
+        stopBeforeSector &&
+        stopBeforeSector.arrival &&
```
RobinCsl: Same here.

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,246 @@
...
+    };
+
+    const duration =
+      this.intervalToString(
+        this.getDurationBetweenSectors(data.route, sectorBorderIndex),
+        'days',
+      ) ?? '';
+    const hasReturnFlight = sectorBorderIndex != null && sectorBorderIndex > 0;
+
+    // @TODO: use Card component
```
RobinCsl: Which Card component?

apps/graphql/src/dataloaders/Itinerariesloader.js
```diff
@@ -21,17 +21,51 @@ export type ItinerariesSearchParameters = {|
...
   +flyTo: string,
   +localDeparture: string,
   +localArrival: string,
+  +price: Price,
+  +route: Array<RouteItem>,
+  +routes: Array<Array<string>>,
 |};
 
+export type ApiRouteItem = {
+  +airline: string,
```
RobinCsl: Aren't we assuming here that the API response will never give null or undefined?

apps/graphql/src/queries/__tests__/Itineraries.test.js
```diff
@@ -24,21 +24,11 @@ it('works', async () => {
...
-          "node": Object {
-            "id": "SXRpbmVyYXJ5OjI=",
-          },
-        },
-      ],
-    },
+    "searchItineraries": null,
   },
+  "errors": Array [
+    [GraphQLError: Cannot read property 'map' of undefined],
```
RobinCsl: Is this error intentional?

apps/graphql/src/types/output/Itinerary.js
```diff
@@ -1,21 +1,56 @@
...
+  GraphQLList,
+} from 'graphql';
 import GlobalID from '@kiwicom/graphql-global-id';
+import { GraphQLDateTime } from 'graphql-iso-date';
 
 export type Itinerary = {|
   +id: string,
 |};
 
+const tripSectorType = new GraphQLObjectType({
```
RobinCsl: What is our convention for GraphQL type definitions? I think that since `GraphQLInt` is in PascalCase, I would like our defined types to follow the same convention.

apps/core/scenes/results/ResultsListItem.js
jvaclavik: It's only about styling. You can see that every connection has white background and two gray borders, so this should be `Card`.
screenshot 2019-01-11 at 16 10 34

apps/mobile/src/screens/ResultsScreen.js
```diff
@@ -11,18 +12,31 @@ type Props = {|
...
     |},
   |},
 |};
 
 export default class ResultsScreen extends React.Component<Props> {
+  static navigationOptions = ({ navigation }: any) => ({
+    headerTintColor: defaultTokens.paletteProductNormal,
+    headerTitleStyle: { color: defaultTokens.colorTextPrimary },
+    title: `${navigation.getParam('travelFrom')} ${
+      navigation.getParam('dateTo') ? '\u21C4' : '\u2192'
```
RobinCsl: We could use one of the components we created. I can't remember how it's called, but that should be possible.

package.json
RobinCsl: And that means it would be easy to replace it here with `date-fns` 😉

packages/components/src/ConnectionCard/ConnectionCard.js
```diff
@@ -50,18 +53,24 @@ export default function ConnectionCard({
...
   wayForth,
   wayBack,
-  // localizedPrice,
+  price,
   duration,
   badges,
 }: Props) {
+  const localizedPrice = price =>
+    price &&
+    new Intl.NumberFormat(price.locale, {
```
RobinCsl: We will need to create a Price component in shared that will render LocalizedPrice under the hood, and it will deal with the localization part.

RobinCsl: Besides, I don't know how this works on Android, I will need to test the app there (there usually are problems when it comes to Intl support, though it might have been fixed in a recent release of RN)

packages/components/src/ConnectionCard/__tests__/ConnectionCard.test.js
```diff
@@ -43,10 +43,14 @@ describe('ConnectionCard', () => {
       duration: '3h',
     },
   ];
-  // const localizedPrice = '€123,455.35';
-  const { getAllByType } = render(
+  const localizedPrice = {
```
RobinCsl: technically, this price is not localized

apps/core/scenes/results/ResultsListItem.js
jvaclavik: OK

apps/core/scenes/results/ResultsListItem.js
jvaclavik: It doesn't work. I guess it handles case only when it's `undefined` but not when it's `null`. Or something similar.

apps/mobile/src/screens/ResultsScreen.js
JoseDuSV: I think that in condition should be `returnDateFrom` instead of `dateTo` to display single arrow for `one way` trips.

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,246 @@
...
+    const hasSectorBorder = sectorBorderIndex != null && sectorBorderIndex > -1;
+    if (data == null) {
+      return null;
+    }
+    const { price } = data;
+    /**
+     * TODO: properly handle possible undefined props in final version of list
+     * (for example: should be entry with undefined price still displayed?)
+     *
+     * NOTE: ugly form of string creation below is used for now because
```
JoseDuSV: This NOTE is probably no longer relevant and can be removed.

apps/mobile/src/screens/ResultsScreen.js
jvaclavik: I agree

apps/core/scenes/results/ResultsListItem.js
jvaclavik: OK

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,246 @@
...
+     * using ?? inside template literals caused problems with code styling
+     * inside VS Code (this part will be probably completely replaced
+     * in next iteration, so it's temporary solution)
+     */
+    const badges = [
+      {
+        id: 1,
```
JoseDuSV: According to `TripSectorWithId` type `id` should be string.

jvaclavik: OK

FilipMessa:
Warnings from console 

screen shot 2019-01-12 at 15 45 32

apps/core/scenes/results/ResultsListItem.js
FilipMessa: I suppose that it should work with `undefined` and `null`
https://github.com/tc39/proposal-optional-chaining

packages/components/src/ConnectionCard/ConnectionCard.js
RobinCsl: Sure, but this PR does not include the patch, so merging it into `master` would break on Android. I will create a PR to add the patch of the Intl library, and then your PR should work nicely. 

screenshot_1547458823

package.json
jvaclavik: OK

apps/core/scenes/results/Results.js
jvaclavik: OK

apps/graphql/src/dataloaders/Itinerariesloader.js
jvaclavik: OK, fixed

jvaclavik:
> Warnings from console
> 
> screen shot 2019-01-12 at 15 45 32

Fixed

apps/graphql/src/queries/__tests__/Itineraries.test.js
jvaclavik: No :) fixed

apps/core/scenes/results/__generated__/ResultsListItem.graphql.js
```diff
@@ -11,60 +11,150 @@ import type { ConcreteFragment } from 'relay-runtime';
...
+      +city: ?string,
+      +cityCode: ?string,
+      +localTime: ?any,
+      +utcTime: ?any,
+    |},
+    +departure: ?{|
+      +city: ?string,
+      +cityCode: ?string,
+      +localTime: ?any,
+      +utcTime: ?any,
```
FilipMessa: these generated flow's types look useless. It looks like I can expect whatever back from graphql

apps/core/scenes/results/ResultsListItem.js
jvaclavik: I see, so there could be some other problem. 

screenshot 2019-01-14 at 14 17 19

packages/components/src/ConnectionCard/ConnectionCard.js
jvaclavik: Done

apps/core/scenes/results/Results.js
```diff
@@ -26,12 +30,30 @@ export default class Results extends React.Component<Props> {
...
+      travelFrom,
+      travelTo,
+      dateFrom,
+      dateTo,
+      returnDateFrom,
+      returnDateTo,
+    } = this.props;
+
+    // @TODO get from config
+    const dateFormat = 'Do MMMM';
```
RobinCsl: Together with the localization of prices, there was also a DateFormatter file: https://github.com/kiwicom/margarita/pull/117/files#diff-cbd47da69efb4e89e3d7340c962d5241

It might be useful here too?

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,195 @@
...
-const styles = StyleSheet.create({
-  listItem: {
-    fontSize: 12,
-    padding: 2,
-    margin: 2,
-    backgroundColor: defaultTokens.backgroundBody,
-  },
-});
+  // @TODO Handle cases when we don't have some data from GraphQL/REST server
+  getSanitizedRouteItem = (route: RouteItem) => {
```
RobinCsl: Could you replace `route` with `routeItem`, otherwise, it's confusing with the naming of the next function.

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,195 @@
...
+      departureTime:
+        (localDeparture && DateFNS.format(localDeparture, timeSimpleFormat)) ??
+        '',
+      duration: duration ?? '',
+      id: (route && route.id) ?? '',
+    };
+  };
+
+  getSectorBorderIndex = () => {
+    const { route, routes } = this.props.data;
```
RobinCsl: I'm confused about this naming too: is `routes` an array of `route`?

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,195 @@
...
+    if (route) {
+      return route.map((routeItem: RouteItem) =>
+        this.getSanitizedRouteItem(routeItem),
+      );
+    }
+    return [];
+  };
+
+  getSector = (
+    route: Array<TripSectorWithId>, // rename to sanitizeRoute
```
RobinCsl: Is this comment still relevant?

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,195 @@
...
+    demandedSector: number,
+    sectorBorderIndex: ?number,
```
RobinCsl: I don't understand that name.

Also, getSector is returning an array of TripSectorWithId, so maybe it should be renamed getSectors?

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,195 @@
...
+  render() {
+    const { data } = this.props;
+    const sanitizedRoute = this.getSanitizedRoute();
+    const sectorBorderIndex = this.getSectorBorderIndex();
+    const hasSectorBorder = sectorBorderIndex != null && sectorBorderIndex > -1;
+    if (data == null) {
+      return null;
+    }
+    const { price } = data;
+    const badges = [
```
RobinCsl: Add a `@TODO` comment?

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,195 @@
...
+        <ConnectionCard
+          wayForth={this.getSector(sanitizedRoute, 0, sectorBorderIndex)}
+          wayBack={
+            hasReturnFlight
+              ? this.getSector(sanitizedRoute, 1, sectorBorderIndex)
+              : undefined
+          }
+          badges={badges}
+          price={priceObject}
+          duration={hasSectorBorder ? duration : ''}
```
RobinCsl: Could you extract this logic before the return statement?

apps/graphql/src/dataloaders/Itinerariesloader.js
```diff
@@ -82,20 +118,43 @@ const fetchItineraries = async (
...
     localArrival: itinerary.local_arrival,
+    routes: itinerary.routes,
+    route:
+      itinerary.route &&
+      itinerary.route.map(routeItem => ({
+        airline: routeItem.airline ?? '',
+        arrival: {
+          city: routeItem.cityTo ?? '',
+          cityCode: routeItem.flyTo ?? '',
+          localTime: routeItem.local_arrival ?? new Date(),
```
RobinCsl: I am not sure `new Date()` is a good default value. Maybe `null` instead?

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,195 @@
...
+    const utcDeparture = this.parseDate(route?.departure?.utcTime);
+    const duration =
+      route &&
+      `${DateFNS.differenceInHours(
+        utcArrival,
+        utcDeparture,
+      )}h ${DateFNS.differenceInMinutes(utcArrival, utcDeparture) % 60}m`;
+
+    return {
+      arrival: (route && route.arrival?.city) ?? '',
```
RobinCsl: It looks like you're already sanitising your GraphQL server outputs, so maybe this is not necessary now.
My point before was that a GraphQL schema should probably contain nullable types, instead of `GraphqlString`, etc. in which case sanitising in the client makes sense. 
@FilipMessa @tbergq What are your thoughts about nullable types by default in GraphQL schemas?

packages/components/src/SearchParamsSummary/SearchParamsSummary.js
```diff
@@ -41,40 +41,38 @@ export default function SearchParamsSummary({
...
-                <AdaptableBadge
-                  style={styles.badge}
-                  textStyle={styles.badgeText}
-                  text={arrival.localizedDate}
-                />
-              </View>
-            )}
-          </View>
-        )}
+        {arrival != null &&
```
RobinCsl: There is a lot of logic here, and a ternary operator in JSX is not easy to parse. 
Would it make sense to create another component?

apps/core/scenes/results/ResultsListItem.js
jvaclavik: It is not. The naming is from REST API and it's not good. We should change it in the future according to @FilipMessa 's conventions. 

`routes` is array of sectors (like `[[OSL, PRG], [PRG, OSL]]`) and `route` is array with info of individual flights (like on `[0]` could be info about OSL -> BCN, on `[1]` could be info about BCN -> PRG, etc.).

apps/core/scenes/results/ResultsListItem.js
jvaclavik: OK

apps/core/scenes/results/ResultsListItem.js
jvaclavik: OK

apps/core/scenes/results/ResultsListItem.js
jvaclavik: Yes, I will change it

apps/core/scenes/results/ResultsListItem.js
jvaclavik: Well, I tried to draw it:
img_8763

`getSector` is fine, but `TripSectorWithId` should be renamed to `TripSegmentWithId` and `TripSector` component should be `TripSegment`.

apps/graphql/src/dataloaders/Itinerariesloader.js
jvaclavik: OK

packages/components/src/SearchParamsSummary/SearchParamsSummary.js
jvaclavik: I don't think it will be better with new component. What about something like this:
```
const isTripExists = arrival != null && departure != null;
...
{isTripExists && (tripType === 'OneWay' ? (...))}
```

apps/core/scenes/results/ResultsListItem.js
jvaclavik: OK

apps/core/scenes/results/Results.js
jvaclavik: Well, is it better in something to use it instead of `date-fns`? Because in this case we would need to take care about logic with ordinal number like (1st, 2nd, 3rd, 4th etc.)

apps/core/scenes/results/ResultsListItem.js
FilipMessa: It should be nullable, It can be annoying handle all this nullable stuff, but on the end, it pushes you as a developer thinking about this possibility. And, yes we are working here with the graphql as a proxy, but we still can rely on that all data will come. 
We should analyze what data are necessary to handle Bisnis logick on the FE, and base on that decided if we show rest of data or some error message. But it needs to analyze what data are important, so, for now, I suggest use "N/A" constant instead of empty string.

What do you think about that?

apps/core/scenes/results/Results.js
```diff
@@ -26,12 +30,30 @@ export default class Results extends React.Component<Props> {
...
+    const {
+      travelFrom,
+      travelTo,
+      dateFrom,
+      dateTo,
+      returnDateFrom,
+      returnDateTo,
+    } = this.props;
+
+    // @TODO get from config
```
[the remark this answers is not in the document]
jvaclavik: No I didn't :) During my work I spot many things which should be handled in the future. And I didn't want to leave them forgotten. I edited description of Trip list issue, and most of the TODOs are written there. 

Should I remove TODOs from the code? I wouldn't create issues for every TODO, because it could be messy. I propose to write all of them to the main issue and remove them from the code if it's a problem to keep them also in the code.

apps/core/scenes/results/ResultsListItem.js
jvaclavik: Done, renamed

apps/core/scenes/results/Results.js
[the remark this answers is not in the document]
FilipMessa: No, we can keep them in the code, I was just worried a lite bit 😨 😄

apps/core/scenes/results/ResultsListItem.js
jvaclavik: Sounds good. Done

apps/core/scenes/results/Results.js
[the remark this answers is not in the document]
jvaclavik: Well when this PR will be merged I'm going to fix some of the TODOs. But I can't do more in this PR. Even now it's terribly huge 😢

packages/components/src/SearchParamsSummary/SearchParamsSummary.js
jvaclavik: Well, `flow` is blind if I use this (^^^) notation with const.

jvaclavik: It should be now more clear: 
```
{arrival &&
          departure &&
          (tripType === 'OneWay' ? (
```

apps/graphql/src/dataloaders/Itinerariesloader.js
```diff
@@ -48,11 +84,10 @@ const parseDate = (date: Date) =>
   DateTime.fromJSDate(date, { zone: 'UTC' }).toFormat(dateFormat);
 
 export const parseParameters = (input: ItinerariesSearchParameters) => {
-  const dateTo = input.dateTo ?? input.dateFrom;
   const params = {
     flyFrom: input.travelFrom,
     dateFrom: parseDate(input.dateFrom),
-    dateTo: parseDate(dateTo),
+    dateTo: parseDate(input.dateFrom),
```
RobinCsl: Why is dateTo pointing to input.dateFrom?

packages/components/src/SearchParamsSummary/SearchParamsSummary.js
```diff
@@ -41,40 +41,38 @@ export default function SearchParamsSummary({
...
-          </View>
-        )}
+        {arrival &&
+          departure &&
+          (tripType === 'OneWay' ? (
+            <View>
+              <AdaptableBadge
+                style={styles.badge}
+                textStyle={styles.badgeText}
+                text={departure.localizedDate}
```
RobinCsl: This won't be displayed if there is no `arrival` though it should be possible, right?

apps/graphql/src/dataloaders/Itinerariesloader.js
jvaclavik: Because API is prepared for date intervals, not for specific date as we want right now. So we have to use `dateFrom` and `dateTo` as a departure and `returnFrom` and `returnTo` as an arrival.

apps/core/scenes/results/Results.js
```diff
@@ -13,25 +15,47 @@ type Props = {|
...
   +dateTo: string,
+  +returnDateFrom: string,
+  +returnDateTo: string,
 |};
 
 type ResultsType = {|
   +searchItineraries: ResultsListType,
 |};
 
+const statusBarHeight = 20; // @TODO add to orbit design tokens
```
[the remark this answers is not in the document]
jvaclavik: Yes I'm sure. 
I tested that on iPhone with rounded display (XS), also on the old one (SE) and also on Android.

We don't have to implement some checks if we use X iphones, because it is solved by using `SafeAreaView` (https://facebook.github.io/react-native/docs/safeareaview)
 in first child of Results screen.

Android also needs to define some top padding and status bar on Android has `20px`. I don't know why they use 0 in your link but when I used `0` padding it looks like this:
screenshot 2019-01-15 at 18 02 29

packages/components/src/SearchParamsSummary/SearchParamsSummary.js
jvaclavik: You are right. Done

apps/core/scenes/results/Results.js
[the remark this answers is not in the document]
tbergq: There are issues with `SafeAreaView` check [this](https://github.com/react-navigation/react-navigation/issues/4165)

According to issue the `SafeAreaView` from `react-navigation` is safe to use. But I have also had issues with it. It doesn't always work

apps/core/scenes/results/ResultsListItem.js
```diff
@@ -1,61 +1,200 @@
...
+    }
+    return null;
+  };
+
+  render() {
+    const { data } = this.props;
+    const sanitizedRoute = this.getSanitizedRoute();
+    const sectorBorderIndex = this.getSectorBorderIndex();
+    const hasSectorBorder = sectorBorderIndex != null && sectorBorderIndex > -1;
+    if (data == null) {
```
FilipMessa: Shouldn't be better if this condition will be imidiatly after `const { data } = this.props`?

jvaclavik: OK, done

PR #91, opened 2019-01-04
Move apikey to client

### LINKED ISSUE #88: Moved the Tequila's apikey from the GraphQL's env to FE application
The GraphQL server will take this from FE request header, and then he can call Tequila API.
Reason for that is that that on the end there should be one GraphQL server that can work with more apikeys (account).

**pls Update documentation also**

tbergq:
closes #88 

We are preparing for this graphql server to be moved public, then
every consumer should pass his own apikey to the graphql server


FilipMessa:
👍just one think can you pls update docs https://github.com/kiwicom/margarita#initial-configuration
thnx

tbergq:
Sure, sorry, forgot that part 🙈

PR #93, opened 2019-01-04
Feat: Adding SearchParamsSummary component

RobinCsl:
Summary: AdaptableBadge had to be added for this move to be possible; this is just a temporary fix and should eventually live inside @kiwicom/universal-components (it's very similar to current implementation of FilterButton there for instance, so refactor is necessary).
I left the stories file for posterity; we might want to add some Storybook-like capacity to the component library specific to margarita.

### DIFF
--- a/packages/components/src/AdaptableBadge/README.md
+++ b/packages/components/src/AdaptableBadge/README.md
@@ -1,0 +1,1 @@
+This is just a temporary fix; it should come from `universal-components`.
--- a/packages/components/index.js
+++ b/packages/components/index.js
@@ -9,3 +9,4 @@ export { default as Modal } from './src/modal/Modal';
 export {
   default as PassengersInputs,
 } from './src/passengersInputs/PassengersInputs';
+export { SearchParamsSummary } from './src/SearchParamsSummary';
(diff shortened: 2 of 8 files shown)


packages/components/src/AdaptableBadge/README.md
```diff
@@ -0,0 +1 @@
+This is just a temporary fix; it should come from `universal-components`.
```
tbergq: Maybe this could be a comment inside the component instead of a readme?

tbergq: Not really important though, the rest looks good

RobinCsl:
I was just thinking that maybe I should merge into a feature branch that would have everything done in it regarding the move + upgrading the library with a working version. WDYT?

FilipMessa:
> I was just thinking that maybe I should merge into a feature branch that would have everything done in it regarding the move + upgrading the library with a working version. WDYT?

TBH: I don't have a strong opinion on this, It OK but on the end, you probably must solve a lot of conflicts when you want to merge it into the master branch.

RobinCsl:
> We should probably just unify folders naming and structure.

I will do that in a separate PR.

People should use these moved components at their own risks! 😈 (until the situation is more stable)

PR #95, opened 2019-01-04
Feat: Adding PlaceCard and Shimmer components

RobinCsl:
Summary: Touchable was added but it should come from universal-components.
PlaceCard comes with its own Placeholder component, which relies on Shimmer component; it is a css-solution for web and a wrapper of `react-native-shimmer-placeholder` for RN.
Changed jest's preset to `jest-expo` as `react-native` was not enough.
Edited defaultSource and source props for Image component in react-native-web flow-typed types: their entries are now read-only.


packages/components/src/PlaceCard/Image/CityImage.js
```diff
@@ -0,0 +1,29 @@
...
+type Props = {|
+  +url: ?string,
+  +style?: StylePropType,
+|};
+
+export default function CityImage({ url, style }: Props) {
+  return (
+    <NetworkImage
+      source={{
+        uri: url || '',
```
tbergq: could we use ?? instead?

packages/components/src/PlaceCard/Image/ImageTypes.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+import type { StylePropType } from '@kiwicom/universal-components';
+
+export type Props = {|
+  +source: {|
+    uri: string,
+    height?: number,
+    width?: number,
```
tbergq: these props should also be read-only

packages/components/src/PlaceCard/PlaceCard.js
```diff
@@ -0,0 +1,112 @@
...
+// import BlackToAlpha from './assets/black-to-alpha-vertical.png'; @TODO
+import CityImage from './Image/CityImage';
+
+type Props = {|
+  +imageUrl?: string,
+  +price: string,
+  +place: string,
+  +country: string,
+  +onPress: () => void,
+  +disabled?: boolean,
```
tbergq: Not documented anywhere (yet) but isn't it nicer to keep mandatory props on top, and optional props at the bottom?

packages/components/src/PlaceCard/PlaceCard.stories.js
```diff
@@ -0,0 +1,83 @@
+/**
```
tbergq: Could we maybe just delete this file?

Though storybook is nice to have, we wanted fast iterations for this project, and storybook tends to make development a little bit slower.

tbergq: Also, you should be able to retrieve this file from universal-components git history if need be later.

packages/components/src/Placeholder/shimmer/index.native.js
```diff
@@ -0,0 +1,22 @@
+// @TODO see if it is still usable somehow: react-native-shimmer must be linked, so it will not work in Expo.
```
tbergq: check [this](https://github.com/tomzaku/react-native-shimmer-placeholder#for-who-using-expo) it is usable with expo

tbergq: 🙈 seems that is another package

packages/components/src/PlaceCard/PlaceCard.js
FilipMessa: It is a nice idea, I checked the `eslint-plugin-flowtype`, but I found only [sort-keys](https://github.com/gajus/eslint-plugin-flowtype#sort-keys) rule, which is not able to check this. 
So I don't see the big advantage from this if we can't ensure automatically check this rule.

packages/components/src/PlaceCard/PlaceCard.stories.js
FilipMessa: I thought about it, and I agree with you.
It's better don't add dead or commented code in project.
It should be there code only what we use, mostly.

packages/components/src/PlaceCard/Placeholder.js
```diff
@@ -0,0 +1,63 @@
+// @TODO inquire if this is still possible/usable
```
FilipMessa: What do you mean by this? thnx

packages/components/src/PlaceCard/Image/ImageTypes.js
RobinCsl: I am afraid this is not possible because of `react-native-web` flow-typed types:
https://github.com/kiwicom/margarita/blob/383ebc262ac22ce0132f1466344caacde604a0ad/flow-typed/npm/react-native-web_vx.x.x.js#L317
>    source?: string | {| uri: string, width?: number, height?: number |},

tbergq: 🙈

tbergq: But this is checked in to our git, so we could always change it to fit our way of using flow.

RobinCsl: But when we upgrade version, it might be painful. Well, I guess it could be painful any way.

packages/components/src/Placeholder/shimmer/index.native.js
RobinCsl: Thanks for suggestion!

packages/components/src/PlaceCard/Image/ImageTypes.js
RobinCsl: Done.

package.json
```diff
@@ -45,6 +45,7 @@
     "eslint-plugin-prettier": "^3.0.0",
     "flow-bin": "^0.88.0",
     "jest": "^23.6.0",
+    "jest-expo": "^32.0.0",
```
tbergq: should it not be in `/apps/mobile/package.json`?

packages/components/src/Shimmer/Shimmer.js
```diff
@@ -0,0 +1,52 @@
+// TODO: This should be usable, but we must have a look at native side.
```
tbergq: Is this comment still true?

package.json
RobinCsl: Well we use it in `.jest.json` at the root of the repo, so I thought it'd be more appropriate there.

packages/components/src/Shimmer/Shimmer.js
RobinCsl: No, it's a ghost from the past. Thanks!

packages/components/src/PlaceCard/Placeholder.js
```diff
@@ -0,0 +1,74 @@
+// @TODO see if it is still usable somehow: react-native-shimmer must be linked, so it will not work in Expo.
```
FilipMessa: What are your plans with this? 
I don't like to have not working code in the repo, because at some point we can easy forgotten about it, but if you have plans to look at it soon it's OK.

packages/components/src/PlaceCard/index.js
```diff
@@ -0,0 +1,4 @@
+// @flow
+
+export { default as PlaceCard } from './PlaceCard';
+// export { default as PlaceCardPlaceholder } from './Placeholder';
```
FilipMessa: this comment can be removed right?

PR #98, opened 2019-01-07
Chore: Improved file structure in packages/components/src

RobinCsl:
Summary: This is to be more consistent with @kiwicom/universal-components


RobinCsl:
Well, this was just to be more consistent with universal-components and @JoseDuSV mentioned that in a previous CR. 
@FilipMessa: what's your opinion?

tbergq:
I find it hard to find the correct files 🙈 
screenshot 2019-01-07 at 11 38 01

RobinCsl:
But if you look for something specific like `Button`, then you get the correct files, don't you?
In my opinion, `index.js` is so generic that it might not be the best search keyword anyway.

FilipMessa:
I understand your motivation but I also don't see the value of what `index.js` files brings when is used only for export the components.
Is there some added value, if we will do it this way with `index.js`?

JoseDuSV:
Personally I don't have a strong opinion about this as with `universal-components`.
I prefer current state without extra `index.js`.

But it will be reasonable to have it unified if we expect more components shuffling between repos in the future. On the other hand, now it's much clearer which component belongs to which project, so it probably won't happen that often.

RobinCsl:
I will close this PR and use the different convention of file structure that you guys prefer, and open a new PR for it.

FilipMessa:
Thanks 

> I will close this PR and use the different convention of file structure that you guys prefer, and open a new PR for it.

PR #99, opened 2019-01-08
Chore: Upgraded @kiwicom/universal-components to v0.0.8

RobinCsl:
Summary: I had to modify the webpack config for the Nextjs application to load the orbit-icons.ttf file. Basically, it copies the font files over to apps/web/static/fonts and they are then accessible; appending a style tag in _document.js makes the Icon component from '@kiwicom/universal-components' work without adding any CSS file.
I had to take into account some breaking changes for Button and RadioButton component: label prop is preferred if there is no need for styling; children prop is used for specific components and/or Text component with diffeerent styling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 # next.js
 #
 .next/
+apps/web/static/fonts/*
 
 # node.js
 #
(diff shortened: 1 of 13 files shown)


RobinCsl:
With the release of @kiwicom/universal-components@0.0.8, we can now upgrade the version and have margarita work correctly :)

PR #107, opened 2019-01-10
Feat: Search form UI

### LINKED ISSUE #23: Create the Search form UI
# The SearchForm component UI

https://www.figma.com/file/z0UzctMydWQXgxeZ00AwqG/Forms?node-id=704%3A20

- Without Tabs (Travel, Room, Cars)

The **SearchForm** will be composed from:

**Fields:**
- [ ] **From** the input field  - TextInput, should be able to accept a default value (optional)
- [ ] **To** input field same as for the **From** field, 

- [ ] **Departure** - for first iteration it should be just the native date picker, will be greate if for design (default state before click) we can reuse the TextInput (universal-component).

- [ ] **Return** - is optional field, otherwise same as for the **Departure**.

**Selectors**

- [ ] **type of the trip**  two possible values: *return* / *one way*, in future there will be more values. should be able to set a default value (one way)

- [ ] **passenger**   - count of the passenger, see design here 

- [ ] **search button** - The Button component (universal-components)

_________________________________________________________________________________________________________

search return -2

JoseDuSV:
Summary:
Updated `Search` form UI for all platforms. 
Related `Modal` component styles update.
Basic `DatePicker` implementation ( will be expanded in next PR with proper limits and rules according to selected values).

Added three basic related components:

`TripTypeButton`
 - Displays selected trip type with proper icon. 
 - Opens trip type `Select` modal.

`PassengersButton`
 - Displays number of passengers and bags. 
 - Opens `PassengersInputs` modal.

`SearchIllustration`
 - Form illustration image implementation slightly different for both platforms.
 
Closes #23


packages/components/src/tripTypeButton/__tests__/TripTypeButton.test.js
```diff
@@ -0,0 +1,25 @@
...
+const icon = 'flight-direct';
+const label = 'One way';
+const onPress = jest.fn();
+
+const { getByType } = render(
+  <TripTypeButton icon={icon} label={label} onPress={onPress} />,
+);
+
+it('should execute onPress method', () => {
+  fireEvent(getByType(TouchableWithoutFeedback), 'press');
```
FilipMessa: `expect(onPress).toHaveBeenCalled()`

FilipMessa:
Nice Job 👍

apps/core/scenes/search/Search.js
```diff
@@ -2,17 +2,22 @@
 
 import * as React from 'react';
 import { View } from 'react-native';
+import { DateTime } from 'luxon';
```
RobinCsl: We will need to discuss if we want to go with `luxon` or `date-fns`. (see https://bundlephobia.com/result?p=luxon@1.9.0)

apps/core/scenes/search/Search.js
```diff
@@ -58,50 +68,69 @@ type State = {|
...
+
+    const defaultDepartureDate = DateTime.local().toISODate();
+    const defaultReturnDate = DateTime.local()
+      .plus({ days: 2 })
+      .toISODate();
     this.state = {
       modalType: MODAL_TYPE.HIDDEN,
       travelFrom: 'OSL',
       travelTo: 'PRG',
-      dateFrom: '2018-12-30',
```
RobinCsl: At last! 😄 Thanks!

apps/core/scenes/search/Search.js
JoseDuSV: You are right. `date-fns` is part of the default kiwi stack.
Have to admit that I simply used `luxon` because it was already in dependencies and is used in graphql part.

packages/components/src/passengersButton/__tests__/PassengersButton.test.js
```diff
@@ -0,0 +1,31 @@
...
+const passengers = 2;
+const bags = 1;
+const onPress = jest.fn();
+
+const { getByType } = render(
+  <PassengersButton passengers={passengers} bags={bags} onPress={onPress} />,
+);
+
+it('should execute onPress method', () => {
+  fireEvent(getByType(TouchableWithoutFeedback), 'press');
```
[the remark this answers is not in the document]
JoseDuSV: You are right. Sorry for the confusion. It definitely make sense to test not just if event exist but also if it was actually fired.
I just checked the `react-native-testing-library` examples ( https://github.com/callstack/react-native-testing-library/blob/master/src/__tests__/fireEvent.test.js) and it's also tested this way there.

So I updated it for this PR, but we should later also unify this in other tests, mostly in `universal-components`.

Thanks!

packages/components/src/tripTypeButton/__tests__/TripTypeButton.test.js
JoseDuSV: `toHaveBeenCalled` added.

apps/core/scenes/search/Search.js
JoseDuSV: @RobinCsl Do you want to have this `luxon` / `date-fns` dilemma resolved before we merge this request or can I make it part of next issue that I have mentioned in PR description (where I plan to expand `DatePicker` implementation little bit, which will mean shuffling with this date related parts little bit)?

RobinCsl: Next PR is fine for me, as `luxon` is already used somewhere else.
However, we should move away from `luxon` towards `date-fns` as soon as possible, to avoid other `luxon` usage.

FilipMessa: definitely, I also prefer `date-fns`

packages/components/src/passengersButton/__tests__/PassengersButton.test.js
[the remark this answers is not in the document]
FilipMessa: 👍

PR #117, opened 2019-01-14
Feat: Added Date and Currency Formatters

RobinCsl:
Summary: Date and Currency Formatters, extracted from mobile repository, rely on Intl library, which needs to be patched for it to work on Android.
This lead to a rewrite of the other patch in apps/mobile in Node, to have all the patches in the same file.

### DIFF
--- a/apps/mobile/scripts/configureApplication.js
+++ b/apps/mobile/scripts/configureApplication.js
@@ -1,0 +1,199 @@
+// @flow strict
+
+const fs = require('fs');
+const path = require('path');
+const rimraf = require('rimraf');
+
+const log = message => console.log(`➤ ${message}`); // eslint-disable-line no-console
+
+log('Patching get-yarn-workspaces');
+/**
+ *
+ * It is necessary to patch this library as the index.js file from the currently published npm version (as of 11/12/2018) does not match the one on their GitHub repository.
+ * cf. https://github.com/viewstools/yarn-workspaces-cra-crna/issues/28
+ *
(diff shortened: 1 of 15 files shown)


apps/mobile/scripts/configureApplication.js
```diff
@@ -0,0 +1,190 @@
...
+        '..',
+        '..',
+        'node_modules',
+        'get-yarn-workspaces',
+        file,
+      ),
+    );
+  });
+
+  // fs.unlinkSync(
```
FilipMessa: Do we need this commented part?

packages/utils/localization/DateUtils.js
```diff
@@ -0,0 +1,132 @@
+// @flow strict
+
+// test immutability
```
FilipMessa: test immutability what does this mean? TODO?

apps/mobile/scripts/configureApplication.js
```diff
@@ -0,0 +1,190 @@
+// @flow strict
+
```
FilipMessa: Can you write some comment or add it to readme why we need this script, please? thanks

packages/utils/localization/DateUtils.js
RobinCsl: This was extracted as is from `mobile` repository.

apps/mobile/scripts/configureApplication.js
RobinCsl: Sure, I'll add some comments.

packages/utils/localization/DateUtils.js
RobinCsl: Added @TODO

Issue #118, opened 2019-01-14
Forbid to have departure date after the date of arrival


jvaclavik:
Right now you can set the date of departure after the date of arrival. When you do it the app crashes.

![simulator screen shot - iphone x - 2019-01-14 at 14 38 08](https://user-images.githubusercontent.com/858321/51116038-33d75380-180a-11e9-8525-ef3b91cd7db2.png)
![simulator screen shot - iphone x - 2019-01-14 at 14 38 12](https://user-images.githubusercontent.com/858321/51116042-35088080-180a-11e9-971b-36fa79235ebd.png)

JoseDuSV:
Closed by #119

PR #119, opened 2019-01-14
Fix: Added search dates limits

JoseDuSV:
Summary:
Added limits to `DatePicker` implementation in `Search` form. Smaller value than current day cannot be selected now. For `return` flights, `departure` cannot be bigger than `return` date.

Replaced `luxon` with `date-fns` in `Search` form.

Related to #118

### DIFF
--- a/apps/core/scenes/search/Search.js
+++ b/apps/core/scenes/search/Search.js
@@ -2,7 +2,6 @@
 
 import * as React from 'react';
 import { View } from 'react-native';
-import { DateTime } from 'luxon';
 import {
   StyleSheet,
   Button,
@@ -25,6 +24,7 @@ import {
   Routes,
   type Navigation,
 } from '@kiwicom/margarita-navigation';
+import * as DateFNS from 'date-fns';
 
 const TRIP_TYPE = {
   return: {
@@ -66,10 +66,10 @@ type State = {|
   modalType: $Keys<typeof MODAL_TYPE>,
   travelFrom: string,
   travelTo: string,
-  dateFrom: string,
-  dateTo: string,
-  returnFrom: string,
-  returnTo: string,
+  dateFrom: Date,
+  dateTo: Date,
+  returnDateFrom: Date,
+  returnDateTo: Date,
   tripType: string,
   datePickerVisible: boolean,
   datePickerDate: Date,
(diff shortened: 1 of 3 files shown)


apps/core/scenes/search/Search.js
```diff
@@ -294,14 +303,14 @@ class Search extends React.Component<Props, State> {
...
-              value={dateFrom}
+              value={DateFNS.format(dateFrom, submitDateFormat)}
             />
             {tripType === 'return' && (
               <TripInput
                 onPress={this.handleReturnDatePress}
                 label="Return"
                 icon={<Icon name="calendar" />}
-                value={returnFrom}
+                value={DateFNS.format(returnDateFrom, submitDateFormat)}
```
FilipMessa: what about DRY and make helper function for that something like `getFormatedDate(date)`?

apps/core/scenes/search/Search.js
```diff
@@ -81,21 +82,20 @@ class Search extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
 
-    const defaultDepartureDate = DateTime.local().toISODate();
-    const defaultReturnDate = DateTime.local()
-      .plus({ days: 2 })
-      .toISODate();
+    const defaultDepartureDate = new Date();
+    const defaultReturnDate = DateFNS.addDays(defaultDepartureDate, 2);
```
RobinCsl: If you look in the PR where I added Date and Currency Formatters, there is a similar function:
https://github.com/kiwicom/margarita/pull/117/files#diff-cbd47da69efb4e89e3d7340c962d5241

apps/core/scenes/search/Search.js
JoseDuSV: Ok, extracted.

apps/core/scenes/search/Search.js
JoseDuSV: Thanks, I've considered it and my problem is that it also converts date to UTC which is not that ideal, because currently we use local timezone in whole search form.
But maybe it will make sense to switch to UTC internally and just render localised strings for date buttons.
Next thing is that if we don't want to get rid of `date-fns` completely and replace it with our custom `DateUtils/Formatter` implementation I don't see much of the added value in duplicating what is already part of the `date-fns`.

RobinCsl: @tbergq What was the reasoning behind having a custom DateFormatter instead of using `date-fns` in `mobile`?

tbergq: The reason was that we was using `moment`, but moment was not good enough. 

The best we found for solving our problems was to use Javascript `Intl` package. If `date-fns` solves the same, we could always replace it in `mobile` in the future.

JoseDuSV: Thanks for explanation. I don't think we are currently facing any limitations with `date-fns`.
But if you think we can expect same problems or limitations as in `mobile` later, I can definitely switch to our custom solution, but I'll probably need to extend it to also handle non-utc dates.

PR #126, opened 2019-01-16
Style search error states

jvaclavik:
screenshot 2019-01-16 at 11 51 40

Btw you can notice that `Text` from `universal-components` with prop `weight="bold"` **is not bold** on iOS


packages/components/src/errorSearchIllustration/ErrorSearchIllustration.native.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+import React from 'react';
+import { Image } from 'react-native';
+import { StyleSheet } from '@kiwicom/universal-components';
+
+// $FlowFixMe - Cannot resolve module RelativeImageStub ( possible solution https://bit.ly/2BYfHRr )
+import ErrorSearchTravel from './assets/mobile/error-search-travel.png';
```
RobinCsl: If you look here, it's also possible to use different pixel density pictures: https://github.com/kiwicom/mobile/tree/master/app/MMB/src/scenes/tripServices/insurance/paymentScene/images

JoseDuSV: As I see there are already `@2x` and`@3x` versions inside assets subfolder, so I think it should be ok.

packages/components/src/ErrorInformation/index.js
```diff
@@ -0,0 +1,3 @@
+// @flow
+
+export { default as ErrorInformation } from './ErrorInformation';
```
JoseDuSV: This file is not necessary, after #98 discussion we settled on structure without extra `index.js` files.

packages/components/src/errorSearchIllustration/ErrorSearchIllustration.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+
+import React from 'react';
+
+const baseImagePath = './static/img/error-search-travel';
+
+export default function ErrorSearchIllustration() {
```
JoseDuSV: It definitely shouldn't be part of this PR. But we can probably create issue after merge to combine this component with `searchIllustration` to some more universal `illustration` component. Since most code is shared between them.

jvaclavik: I though about the same. We will create issue for it

packages/components/src/ErrorInformation/index.js
jvaclavik: OK, that's better :)

packages/components/src/errorSearchIllustration/ErrorSearchIllustration.js
FilipMessa: Where is the issue?

RobinCsl:
Please @jvaclavik, look at #132 
We shouldn't be creating components in a different package with dependencies in a different one. This is going to break whenever we (re)move pictures from `apps/web/static` by mistake, thinking it's not used anywhere in that folder, and all the more so when there is another solution by using Kiwi.com's CDN.

jvaclavik:
> Please @jvaclavik, look at #132
> We shouldn't be creating components in a different package with dependencies in a different one. This is going to break whenever we (re)move pictures from `apps/web/static` by mistake, thinking it's not used anywhere in that folder, and all the more so when there is another solution by using Kiwi.com's CDN.

OK, I understand, I didn't know that there is some CDN for it, I used the same approach as in search illustration and it worked :) 

But it would be much better to get it from CDN. How can I find URL of the image?

packages/components/src/errorSearchIllustration/ErrorSearchIllustration.js
jvaclavik: Probably nowhere when we will use images from CDN. It was about unifying illustration components.

jvaclavik:
@JoseDuSV It should be fixed now

packages/relay/src/QueryRenderer.js
```diff
@@ -22,9 +23,10 @@ export default class App extends React.Component<Props> {
   renderRelayContainer = ({ error, props }: ReadyState) => {
     if (error) {
       return (
-        <View>
-          <Text style={styles.text}>{error.message}</Text>
-        </View>
+        <IllustrationWithInformation
+          illustrationName="Error"
+          text={error.message}
```
JoseDuSV: In final version we will probably need some default error message instead of displaying exact error string (which can be quite confusing for user).
screenshot 2019-01-22 at 15 03 29

Or at least display some common "**Something went wrong**" header and then _error message_ underneath it with some small font and semi-transparent color.

But for now the current implementation is probably fine.

apps/core/scenes/results/EmptyResults.js
```diff
@@ -1,13 +1,13 @@
...
+import { IllustrationWithInformation } from '@kiwicom/margarita-components';
 
 export default function EmptyResults() {
   return (
-    <View>
-      <Text>EmptyResults</Text>
-    </View>
+    <IllustrationWithInformation
+      illustrationName="NoResults"
+      text="Sorry, we couldn't find that connection."
```
JoseDuSV: Maybe we can also add `Try selecting different dates or nearby places.` suggestion like in web version. But it's totally up to you, I'm ok even with current state.

screenshot 2019-01-22 at 14 15 58

packages/relay/src/QueryRenderer.js
jvaclavik: OK, done

apps/core/scenes/results/EmptyResults.js
jvaclavik: Done (I changed the text according to new one in Figma)

jvaclavik: screenshot 2019-01-22 at 16 06 40

packages/relay/src/QueryRenderer.js
jvaclavik: screenshot 2019-01-22 at 16 07 34

PR #127, opened 2019-01-16
Disable header on SearchScreen

jvaclavik:
screenshot 2019-01-16 at 12 25 35

left = before, right = after


apps/core/scenes/search/Search.js
```diff
@@ -82,6 +82,10 @@ const getFormatedDate = date => {
 };
 
 class Search extends React.Component<Props, State> {
+  static navigationOptions = {
```
JoseDuSV: What about moving this into `mobile/src/screens/SearchScreen.js` when it is related only to mobile and we still display some kind of header in web version, so it can be confusing.
Maybe we can disable header globally and turn it on only if needed/required.

jvaclavik: We don't want to disable header globally, because for example we use it in `Results` (but it's transparent).

I'm not sure about moving `navigationOptions` to the screen. At least if we want it we have to do it also for `ResultsScreen`. 

And there is probably some complication, I'm not sure if it's worth for spend more time on it:

```
Cannot cast object literal to NavigationRouteConfigMap because property header is missing in function type [1] but
exists in object literal [2] in property navigationOptions of property search.screen.

     apps/mobile/App.js
      20│   NavigationOptions,
      21│   NavigationProps,
      22│ > = createStackNavigator(
      23│   ({
      24│     [Routes.SEARCH]: {
      25│       screen: SearchScreen,
      26│     },
      27│     [Routes.RESULTS]: {
      28│       screen: ResultsScreen,
      29│     },
      30│     [Routes.PLACE_PICKER]: {
      31│       screen: PlacePickerScreen,
      32│     },
      33│   }: NavigationRouteConfigMap),
      34│ );
      35│
      36│ const AppContainer: NavigationContainer<

     flow-typed/npm/react-navigation_v3.x.x.js
 [1] 302│     navigationOptions?: NavigationScreenConfig<Options>;

     apps/mobile/src/screens/SearchScreen.js
 [2]  14│   static navigationOptions = {
      15│     header: null,
      16│   };
```

JoseDuSV: If I try something like below the header is gone and I don't see any error.
```

import { Search } from '@kiwicom/margarita-core';
import React, { Component } from 'react';

export default class SearchScreen extends Component {
  static navigationOptions = {
    header: null,
  };

  render() {
    return <Search />;
  }
}
```
It simply make more sense to me to keep this part in mobile code if it affects only this platform.
But if there is any other problem with this approach I definitely don't insist on it.

JoseDuSV: Oh, sorry. You're right.
And what about extending `mobile/App.js` like:
```
const AppNavigator: NavigationNavigator<
  NavigationState,
  NavigationOptions,
  NavigationProps,
> = createStackNavigator(
  ({
    [Routes.SEARCH]: {
      screen: SearchScreen,
      navigationOptions: ({ navigation }) => ({
        header: null,
      }),
    },
    [Routes.RESULTS]: {
      screen: ResultsScreen,
    },
    [Routes.PLACE_PICKER]: {
      screen: PlacePickerScreen,
    },
  }: NavigationRouteConfigMap),
);
```

RobinCsl:
I cannot reproduce from `master` what you solved. 🤔

FilipMessa:
> I cannot reproduce from `master` what you solved. 🤔

me either

jvaclavik:
> > I cannot reproduce from `master` what you solved. 🤔
> 
> me either

I see, looks like it has been already solved by another commit in that big PR. This solution is probably little bit better, but it's not a big deal. We can also decline it.

Issue #109, opened 2019-01-13
Deploy the GraphQL to the public.


FilipMessa:
We will have a presentation in February and we would like to have Graphql Playground available for the public. There should be a warning that it is still in progress and breaking changes are expected.

PR #129, opened 2019-01-16
Chore: Deployed Margarita's graphql and web apps on Netlify

RobinCsl:
Summary: After this PR is merged, Netlify will build both apps whenever a change occurs on the master branch.

apps/graphql/lambda is the folder that will be built by Netlify; so right now, only graphql.js would be built. Once build is successful, the Netlify Function is available at /.netlify/functions/<name_of_file> if <name_of_file>.js is in apps/graphql/lambda.

It will also deploy the static site from NextJs, which is built into 'apps/web/out'.

Because we are using `react-native-dotenv` in @kiwicom/margarita-relay, we need to explicitly create a .env file from the environment variables (which are already added to Netlify Deploy Settings); hence the 'echo GRAPHQL_URL=$GRAPHQL_URL > .env;' inside netlify.toml. That means we will need to be mindful of when we use more environment variables from import calls with `react-native-dotenv`.

This netlify.toml file is necessary to tell Netlify which build command to execute, and where to look for built versions of Netlify Functions. In our build command, there is 'yarn netlify-lambda:build', which stands for 'netlify-lambda build apps/graphql/lambda'. That build will compile everything from 'apps/graphql/lambda' into 'apps/graphql/netlify'.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,12 +6,17 @@
 #
 .next/
 apps/web/static/fonts/*
+apps/web/out/
 
 # node.js
 #
 node_modules/
 npm-debug.log
 yarn-error.log
+
+# apps/graphql
+#
+apps/graphql/netlify
 
 .env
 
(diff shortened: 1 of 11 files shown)


RobinCsl:
Closes https://github.com/kiwicom/margarita/issues/109

PR #132, opened 2019-01-17
Fix: Search Illustration is fetched from CDN

RobinCsl:
Summary: The way the component was built relied on logic that could break easily: if anyone (re)moved the files in apps/web/static, the component would break without any warning.
Orbit components' Illustration component fetches from Kiwi.com's image CDN, and I believe this is more robust and extensible.


RobinCsl:
@JoseDuSV I believe you cannot really do that. However, Orbit Components' Storybook makes it sort of possible https://kiwicom.github.io/orbit-components/?selectedKind=Illustration&selectedStory=Playground&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs
(Looking at the corresponding code on their repo can show you all the available names too)

PR #137, opened 2019-01-18
Create Illustration component

### LINKED ISSUE #135: Component for illustrations

jvaclavik:
Reusable component for Kiwi illustrations (closes #135).
I also re-add images illustrations for Boarding because they were in repo with incorrect sizes.

https://kiwicom.github.io/orbit-components/?selectedKind=Illustration&selectedStory=Playground&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs


packages/components/src/Illustration/IllustrationTypes.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+
+import { type StylePropType } from '@kiwicom/universal-components';
+
+export type IllustrationProps = {|
+  source: string,
```
JoseDuSV: What about using `name` instead of `source` to be consistent with Orbit ( https://orbit.kiwi/components/illustration/ ).

Also we can use union types similar to https://github.com/kiwicom/orbit-components/blob/master/src/Illustration/index.js.flow (but just use currently implemented illustrations, because of the mobile version).

jvaclavik: Good idea!

PR #140, opened 2019-01-18
[WIP] Create tokens for Margarita

### LINKED ISSUE #136: Move repeating style values to separate package
Will serve as future resource for Orbit design tokens.

As discussed in #140 we need separate `margarita-design-tokens` package to keep repeating style values which are not yet part of the `orbit-design-tokens`. To clean codebase from style related TODOs.

jvaclavik:
According to agreement with Orbit team we should create and maintain our design tokens before it will be integrated into Orbit design tokens.

Structure of the tokens should be flat and naming should be in the same way as in current default tokens: https://orbit.kiwi/design-tokens/

**When this PR will be merged we shouldn't use common values hardcoded inside of our components. Please keep it in mind in your future pull requests.**

(closes #136)

### DIFF
--- a/apps/core/scenes/results/Results.js
+++ b/apps/core/scenes/results/Results.js
@@ -6,6 +6,7 @@ import { QueryRenderer, graphql } from '@kiwicom/margarita-relay';
 import { StyleSheet } from '@kiwicom/universal-components';
 import * as DateFNS from 'date-fns';
 import { SearchParamsSummary } from '@kiwicom/margarita-components';
+import { mobileTokens } from '@kiwicom/margarita-utils';
 
 import type { ResultsList as ResultsListType } from './__generated__/ResultsList.graphql';
 import ResultsList from './ResultsList';
@@ -22,8 +23,6 @@ type Props = {|
 type ResultsType = {|
   +searchItineraries: ResultsListType,
 |};
-
-const statusBarHeight = 20; // @TODO add to orbit design tokens
 
 export default class Results extends React.Component<Props> {
   renderInner = (props: ResultsType) => {
@@ -84,7 +83,12 @@ export default class Results extends React.Component<Props> {
 const styles = StyleSheet.create({
   container: {
     flex: 1,
-    marginTop: statusBarHeight,
+    android: {
+      marginTop: mobileTokens.heightStatusBarAndroid,
+    },
+    ios: {
+      marginTop: mobileTokens.heightStatusBarIOS,
+    },
     web: {
       marginTop: 0,
     },
--- a/packages/components/src/PlaceCard/Placeholder.js
+++ b/packages/components/src/PlaceCard/Placeholder.js
@@ -4,6 +4,7 @@ import * as React from 'react';
 import { View } from 'react-native';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import { StyleSheet } from '@kiwicom/universal-components';
+import { tokens } from '@kiwicom/margarita-utils';
 
 import { Shimmer } from '../Shimmer';
 
@@ -12,22 +13,22 @@ const Placeholder = () => (
     <View style={styles.pricePlaceholder}>
       <Shimmer
         style={[styles.defaultPlaceholder, styles.large]}
-        width={80}
-        height={25}
+        width={tokens.widthShimmerSmall}
+        height={tokens.heightShimmerMedium}
       />
     </View>
     <View style={[styles.bottomContainer, styles.padding]}>
-      <View style={styles.placePlacholder}>
+      <View style={styles.placePlaceholder}>
         <Shimmer
-          width={200}
-          height={16}
+          width={tokens.widthShimmerLarge}
+          height={tokens.heightShimmerSmall}
           style={[styles.defaultPlaceholder, styles.small]}
         />
       </View>
       <View style={styles.countryPlaceholder}>
         <Shimmer
-          width={100}
-          height={16}
+          width={tokens.widthShimmerMedium}
+          height={tokens.heightShimmerSmall}
           style={[styles.defaultPlaceholder, styles.small]}
         />
       </View>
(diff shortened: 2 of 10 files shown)


packages/utils/theme/tokens.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
+export default {
+  heightPlaceCard: 152,
+  heightPlaceholder: 64,
+  heightSearchParamsSummary: 64,
```
JoseDuSV: Do we need separate tokens for `heightSearchParamsSummary`, `heightTripSegment`, `heightTimelineInformation` if this value seems to be used only once and is closely connected just to a specific component?

JoseDuSV: Or is your core idea just to collect all possible recurring values and sort everything later?

jvaclavik: Well for now we don't know if this value will be specific only for one component or it will be similar with something else. 

For example during this PR I find by accident that there are `PlaceCard` and `Placeholder` with the same `height`. I didn't know about it before but when I had two similar values in tokens, it was easy to find.

Btw Honza from Orbit wanted to collect in tokens every numbered value, but I think it's not so useful. We can start with something and we will see how it works in the future.

apps/core/scenes/results/Results.js
```diff
@@ -6,6 +6,7 @@ import { QueryRenderer, graphql } from '@kiwicom/margarita-relay';
 import { StyleSheet } from '@kiwicom/universal-components';
 import * as DateFNS from 'date-fns';
 import { SearchParamsSummary } from '@kiwicom/margarita-components';
+import { mobileTokens } from '@kiwicom/margarita-utils';
```
JoseDuSV: I definitely don't insist on it and it's just a quick idea, but what about moving it to separate package? Something like:
`import { mobileTokens } from '@kiwicom/margarita-design-tokens';`

packages/components/src/PlaceCard/Placeholder.js
```diff
@@ -37,28 +38,28 @@ const Placeholder = () => (
 
 const styles = StyleSheet.create({
   container: {
-    height: 152,
+    height: tokens.heightPlaceCard,
```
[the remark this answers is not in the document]
jvaclavik: I don't think so. Because it has the same height as `PlaceCard` and `Placeholder` is sub-component of `PlaceCard` (at least according to directory naming)

packages/utils/theme/tokens.js
JoseDuSV: Thanks for confirmation, that makes sense.
Just hoping that it will not bubble up that much, but we will see after a while ;) .

apps/core/scenes/results/Results.js
jvaclavik: Why not, but let's wait for notes from the others.

packages/utils/theme/tokens.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
+export default {
+  heightPlaceCard: 152,
+  heightPlaceholder: 64,
```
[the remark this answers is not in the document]
jvaclavik: True, deleted

packages/components/src/PlaceCard/Placeholder.js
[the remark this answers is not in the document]
JoseDuSV: Ok, I just thought so because of that unused `heightPlaceholder` value in tokens. Sorry

RobinCsl:
I don't know if all these values need to be abstracted away as design tokens. Some are just very specific to one component. It might be hard in the future to rename components and use a different token name if we ever need to. 
Also, I believe this job should partly fall on the designer for mobile components as well: he knows best what measures he wants and where.

TL;DR: I guess tokens about color, etc. make sense. But as for very specific sizes of components, I am not so sure. 

@tbergq What's your opinion in this matter?

JoseDuSV:
Again, just a quick idea :). What about slightly more non-destructive approach and instead of extracting values into separate files right away. We can just have some special todo comment for it like:
`// @TODO tokens` 
which we can place next to every style value we think that should be ideally later part of the tokens.
So we can easily filter them out later without worrying about naming and possible renaming.
This way we can also point out other token related problems and inconsistencies.

tbergq:
> @tbergq What's your opinion in this matter?

I believe something generic and reusable could be put into one file for re-use, but if I have my AwesomComponentWithHeight454, I don't see why I should put that into some token, if it only matters to that one component.

jvaclavik:
> > @tbergq What's your opinion in this matter?
> 
> I believe something generic and reusable could be put into one file for re-use, but if I have my AwesomComponentWithHeight454, I don't see why I should put that into some token, if it only matters to that one component.

We should use orbit design tokens (or margarita design tokens) for some important values which could be common in more components. For example, If you want to create `PlaceCard` it should have the same height as `Placeholder` for this card. Height for `SearchParamsSummary` could be the same as some different header which we don't have right now. Maybe it should have more generic names, but I think those values could be shared. 
How can we recognise important common values if we will not keep them out of the specific component?

Honza from orbit also told me, that the main goal of the tokens is to be able to re-style whole design by using different design tokens. In ideal case it should have the same keys with different (or the same) values across the platforms. And it also means sizes of the components. Maybe we could discuss this philosophical question also with them?


> Again, just a quick idea :). What about slightly more non-destructive approach and instead of extracting values into separate files right away. We can just have some special todo comment for it like:
> `// @TODO tokens`
> which we can place next to every style value we think that should be ideally later part of the tokens.
> So we can easily filter them out later without worrying about naming and possible renaming.
> This way we can also point out other token related problems and inconsistencies.

This is more about form of the data and I'm not sure if it helps to @tbergq 's and @RobinCsl 's notes. I'm not sure if it's better because if we want to create values in design tokens, we have to sooner or later send those values to orbit team. And it's easier to do it if you have it in one specific file.

FilipMessa:
So we still did not agree about this topic.
What I understand, is that sometimes in the **orbit-tokens** missing value what we need, and they exist in the mobile designs. Like different color for Badges then are on the web.  Or maby there are some the differences in accents colors.  But It should always related to the basic atomic components. We don't want to add every component styles value to a design-token ( `heightSearchParamsSummary`,  `heightTripSegmen`), we are not building the theme for the Margarita.

So my suggestion is if you realize that there is missing value in the Orbit design token add this value to let's say `design-token.js`  file. 
But it should be just valued relate to an atomic component, not values specific for bigger components,  at least for now.

What do you think?

JoseDuSV:
@FilipMessa I agree, it also makes sense to me just to keep global ( like **Color palette** from `orbit-design-tokens`) and atomic components related values.

jvaclavik:
> So we still did not agree about this topic.
> What I understand, is that sometimes in the **orbit-tokens** missing value what we need, and they exist in the mobile designs. Like different color for Badges then are on the web. Or maby there are some the differences in accents colors. But It should always related to the basic atomic components. We don't want to add every component styles value to a design-token ( `heightSearchParamsSummary`, `heightTripSegmen`), we are not building the theme for the Margarita.
> 
> So my suggestion is if you realize that there is missing value in the Orbit design token add this value to let's say `design-token.js` file.
> But it should be just valued relate to an atomic component, not values specific for bigger components, at least for now.
> 
> What do you think?
> @jvaclavik @tbergq @JoseDuSV @RobinCsl

OK, but it's not important right now, I will focus on it after Feb, 14th.

FilipMessa:
I suppose that we can close this PR.

Issue #123, opened 2019-01-15
GraphQL: Itinerary new schema design


FilipMessa:
```
Country: {
  "id": GlobalID,
  "name": string,
  "code": string,
  "slug": string,
  "flagURL": string,
}

Location: {
  "id": GlobalID,
  "locationId": 3-letter IATA code of airport or internal city code (can we take it from global?),
  "name": string,
  "timezone": GlobalID,
  "country": Country
}

Price: {
"currency": string,
"amount: number,
}

Date: {
 "local": local time,
  "utc": UTC date format,
}

Sector: {
 "id": GlobalID,
 "duration":  duration in minutes,
 "departureTime": Date,
 "arrivalTime": Date,
  "origin": Location, (flyFrom),
  "destination": Location, (flyTo),
 "segments" : Segment[],
 "connections" : Array<Array[Segment]>  // segments  ordered by date, if array length is more than one it is connection.  
}

Vehicle: {
  "type": bus | train | aircraft,
  "uniqueNo": flightNumber, or bus number,
}

Provider: {
"id": GlobalID,
"name": String, // RegioJet, Rayner, Emirates, etc….
}

Segment: {
  "id": GlobalID,
  "origin": Location,
  "destination": Location,
  "duration":  duration in minutes,
  "departureTime": Date,
  "arrivalTime": Date,
  "provider": Provider,
  "vehicle":  Vehicle,
 "connections": Segment[]  // the array of related Segments 
}

Itinerary: {
  "id": GlobalID,
  "type": "oneway" | "return" | "multicity",
  "price": Price,
  "origin" (travelFrom): Location,
  "destination" ("travelTo") Location,
  "startTime": Date,
  "endTime": Date,
  "sectors": Sector[],
}
```

discussion: https://skypicker.slack.com/archives/CEFT2NAE8/p1547407772002900

FilipMessa:
Done!

PR #141, opened 2019-01-21
Mock new GraphQL structure

jvaclavik:
New structure mocked on GraphQL server. There are no changes in previous structure, so nothing which have worked till now should be affected.

Issue and structure specification is here: #123

### Query example

**Query:**
```
query RootQuery($input: ItinerariesSearchInput!){
  searchItineraries (input: $input){
    edges{
      node{
        id
        type
        price {
          amount
          currency
        }
        origin {
          id
          locationId
          name
          timezone
          country {
            name
            code
          }
        }
        destination {
          id
          locationId
          name
          timezone
          country {
            name
            code
          }
        }
        startTime {
          local
          utc
        }
        endTime {
          local
          utc
        }
        sectors{
          arrivalTime{
            local
          }
          connections {
            origin {
              locationId
            }
            destination {
              locationId
            }
          }
          departureTime{
            local
          }
          destination{
            locationId
          }
          duration
          id
          origin{
            locationId
          }
          segments{
						provider{
              name
            }
            vehicle{
              type
            }
            duration
          }
        }
      }
    }
  } 
}
```

**Query variables**
```
{
  "input": {
    "travelFrom":"OSL",
    "travelTo": "PRG",
    "dateFrom": "2019-02-10",
    "dateTo": "2019-02-11"
  }
}
```

### DIFF
--- a/apps/graphql/src/dataloaders/Itinerariesloader.js
+++ b/apps/graphql/src/dataloaders/Itinerariesloader.js
@@ -7,149 +7,18 @@ import { OptimisticDataloader } from '@kiwicom/graphql-utils';
 
 import itineraryMock from '../mocks/itinerary';
 import fetch from '../services/Fetch';
-
-export type ItinerariesSearchParameters = {|
-  +travelFrom: string,
-  +dateFrom: Date,
-  +dateTo?: Date,
-  +travelTo?: string,
-  +returnDateFrom?: Date,
-  +returnDateTo?: Date,
-  +passengers?: {|
-    +adults?: number,
-    +children?: number,
-    +infants?: number,
-  |},
-|};
-
-export type TripSegment = {|
-  +city: string,
-  +cityCode: string,
-  +localTime: ?Date,
-  +utcTime: ?Date,
-|};
-export type RouteItem = {|
-  +airline: string,
-  +arrival: TripSegment,
-  +departure: TripSegment,
-  +id: string,
-|};
-
-// start - new structure
-
-export type Country = {|
-  +id: string,
-  +name: string,
-  +code: string,
-  +slug: string,
-  +flagURL: string,
-|};
-
-// @TODO rename PriceType
-export type Price = {|
-  +amount: number,
-  +currency: string,
-|};
-
-export type LocationType = {|
-  +id: string,
-  +locationId: string,
-  +name: string,
-  +timezone: string,
-  +country: Country,
-|};
-export type DateType = {|
-  +local: Date,
-  +utc: Date,
-|};
-export type ProviderType = {|
-  +id: string,
-  +name: string,
-|};
-export type VehicleType = {|
-  +type: string,
-  +uniqueNo: string,
-|};
-export type SegmentType = {|
-  +arrivalTime: DateType,
-  +departureTime: DateType,
-  +destination: LocationType,
-  +duration: number,
-  +id: string,
-  +origin: LocationType,
-  +provider: ProviderType,
-  +vehicle: VehicleType,
-|};
-export type SectorType = {|
-  +arrivalTime: DateType,
-  +connections: Array<SegmentType>,
-  +departureTime: DateType,
-  +destination: LocationType,
-  +duration: number,
-  +id: string,
-  +origin: LocationType,
-  +segments: Array<SegmentType>,
-|};
-
-export type newItinerariesStructure = {|
-  id: string,
-  +type: string,
-  +price: Price,
-  +origin: LocationType,
-  +destination: LocationType,
-  +startTime: DateType,
-  +endTime: DateType,
-  +sectors: Array<SectorType>,
-|};
-// end - new structure
-
-export type Itineraries = {|
-  +airlines: Array<string>,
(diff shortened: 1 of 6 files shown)


apps/graphql/src/mocks/itinerary.js
```diff
@@ -0,0 +1,108 @@
...
+const getSegment = (origin, destination, departureTime, arrivalTime) => ({
+  arrivalTime: getDate(2019, 3, 14, 10, 5),
+  departureTime: getDate(2019, 3, 11, 10, 5),
+  destination: destination,
+  duration: 100,
+  id: 'asdfg',
+  origin: origin,
+  provider: provider,
+  vehicle: vehicle,
+  // "connections": Segment[]  // the array of related Segments
```
jvaclavik: I'm not sure if we will need this.

apps/graphql/src/dataloaders/Itinerariesloader.js
```diff
@@ -34,11 +35,74 @@ export type RouteItem = {|
...
+
+export type Country = {|
+  +id: string,
+  +name: string,
+  +code: string,
+  +slug: string,
+  +flagURL: string,
+|};
+
+// @TODO rename PriceType
```
FilipMessa: Base on what you decided when using to `Type` prefix and when not? 
For example, `Country ` is without prefix and `LocationType` is with.
Is there some logic behind this?
Thanks

FilipMessa: What about creating a new file for types (ItinerariesloaderTypes.js)?
Because all these types are exported which means that they are reuse or will be reused on other places. 
Benefits will be:
1. smaller Itinerariesloader.js file - easiest to read
2. Clearer will be also when I will need to find some type. That I can look to files with `Types` suffix.

What do you think?

apps/graphql/src/mocks/itinerary.js
```diff
@@ -0,0 +1,108 @@
+// @flow
+
+const getCountry = () => ({
+  id: 'aaa',
+  name: 'Norway',
+  code: 'NO',
+  slug: '',
+  flagURL: '',
```
FilipMessa: pls, add some dummy date also here instead of the empty string if it is not necessary.
It will be easiest to recognize that everything is working properly.
It can be just something like that:
```
slug: 'slug',
flagURL: 'flagUrl',
```

apps/graphql/src/dataloaders/Itinerariesloader.js
```diff
@@ -34,11 +35,74 @@ export type RouteItem = {|
...
+export type DateType = {|
+  +local: Date,
+  +utc: Date,
+|};
+export type ProviderType = {|
+  +id: string,
+  +name: string,
+|};
+export type VehicleType = {|
+  +type: string,
```
FilipMessa: this can be `enum` or it is not a good idea? (// bus | train | aircraft,)

apps/graphql/src/mocks/itinerary.js
FilipMessa: I don't know probably we can avoid this for now, but I think will be useful know if the current segment is part of a connection. So maby ` "connections"?: Array<SegmentID>` will be enough for this information.

apps/graphql/src/queries/__tests__/Itineraries.test.js
```diff
@@ -28,12 +28,12 @@ Object {
...
         Object {
           "node": Object {
-            "id": "SXRpbmVyYXJ5OjE=",
+            "id": "SXRpbmVyYXJ5OmFzZGY=",
           },
         },
         Object {
           "node": Object {
-            "id": "SXRpbmVyYXJ5OjI=",
+            "id": "SXRpbmVyYXJ5OmFzZGY=",
```
FilipMessa: I not sure but, if it can make trouble in mocks but unique IDs are very important in GraphQL.

apps/graphql/src/dataloaders/Itinerariesloader.js
jvaclavik: I agree. 

Btw I used `Type` suffix because I have to distinguish normal `Date` from `Date` defined by me. I think we should add `Type` suffix everywhere for consistency. What do you think?

FilipMessa: Sure, if it will help then let's do it.
What do you think guys? @JoseDuSV @tbergq @RobinCsl

apps/graphql/src/mocks/itinerary.js
jvaclavik: Done

apps/graphql/src/dataloaders/Itinerariesloader.js
jvaclavik: Done

apps/graphql/src/dataloaders/Itinerariesloader.js
jvaclavik: I separated Types to different file and added suffixes to places where they missed

PR #145, opened 2019-01-24
Add bottom tab navigation, and navbar links for web

### LINKED ISSUE #144: Tab bar navigation
_As a user, I want to be able to switch between Search and Bookings (MMB) in the app._ 

Let's implement a simple tab bar navigation on the bottom of the screen. 

For the initial phase, there would be only two tabs: Search and Bookings. Please, align the icons properly :-) 

snimek obrazovky 2019-01-24 v 14 06 28

Design in Figma (for now, use only the Tab bar part): https://www.figma.com/file/GpBUlzILpGgkmmbv4oItD2ZD/Menu-(profile)?node-id=179%3A3011

tbergq:
closes #144

### DIFF
--- a/apps/web/components/Navbar.js
+++ b/apps/web/components/Navbar.js
@@ -6,24 +6,28 @@ import { StyleSheet } from '@kiwicom/universal-components';
 import Router from 'next/router';
 import { TouchableWithoutFeedback } from '@kiwicom/margarita-components';
 
-/**
- * This component should only be used by web.
- * react-navigation handles header for mobile
- */
+import NavbarLink from './NavbarLink';
+
 const goToHomePage = () => {
   Router.push('/');
 };
+
 export default function Navbar() {
   return (
     <View style={styles.navbar}>
-      <TouchableWithoutFeedback onPress={goToHomePage}>
-        <Image
-          style={styles.image}
-          source={{ uri: 'https://go.kiwi.com/logo.jpg' }}
-          accessibilityLabel="Logo"
-        />
-      </TouchableWithoutFeedback>
-      {/* TODO: Add more navigation links */}
+      <View style={styles.actionsLeft}>
+        <TouchableWithoutFeedback onPress={goToHomePage}>
+          <Image
+            style={styles.image}
+            source={{ uri: 'https://go.kiwi.com/logo.jpg' }}
+            accessibilityLabel="Logo"
+          />
+        </TouchableWithoutFeedback>
+        <NavbarLink label="Travel" route="/" />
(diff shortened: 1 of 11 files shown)


apps/web/components/Navbar.js
```diff
@@ -39,7 +43,7 @@ const styles = StyleSheet.create({
     },
     shadowRadius: 3,
     paddingHorizontal: 10,
-    justifyContent: 'flex-start',
+    justifyContent: 'space-between',
```
RobinCsl: I am not entirely sure about this layout, but I guess we would need designs to tell us what we should do.

tbergq: It is how the web navbar is
screenshot 2019-01-25 at 10 04 11

apps/web/components/NavbarLink.js
```diff
@@ -0,0 +1,66 @@
...
+|};
+
+export default class NavbarLink extends React.Component<Props, State> {
+  state = {
+    isActive: false,
+  };
+
+  componentDidMount() {
+    if (Router.route === this.props.route) {
+      this.onMouseEnter();
```
FilipMessa: Just one think this naming `this.onMouseEnter();` is confusing a little bit.
probably just will be more clearly: 
```
 this.setState({ isActive: true });
```

apps/web/components/NavbarLink.js
```diff
@@ -0,0 +1,66 @@
...
+  };
+
+  onPress = () => {
+    Router.push(this.props.route);
+  };
+
+  render() {
+    return (
+      <TouchableWithoutFeedback
+        onMouseEnter={this.onMouseEnter}
```
JoseDuSV: Can we please use `Hoverable` from `universal-components` instead?
```
<Hoverable
   onMouseEnter={this.onMouseEnter}
   onMouseLeave={this.onMouseLeave}
>
```

It disables hover in web version on touch devices. Where `onMouseLeave` is not called and situations like below when component is stuck in hovered mode can happen.

screenshot 2019-01-25 at 10 06 57

apps/web/components/NavbarLink.js
tbergq: Yes, agreed, however there is an eslint-rule saying `no-set-state` in componentDidMount. 

Ideally, we should have done this in the constructor, however, `next/router` can not be accessed on the server, and constructor is executed also on server.

so then we would need to change it to: 

```
this.setState({ isActive: true }); // eslint-disable-line react/no-did-update-set-state
```

PR #155, opened 2019-01-27
Chore: update kiwi dependencies

FilipMessa:
- @kiwicom/graphql-global-id to 0.4.1
- @kiwicom/graphql-utils to 0.3.1
- @mrtnzlml/fetch to  2.0.0
- @kiwicom/eslint-config  to v2.5.0


apps/mobile/index.js
```diff
@@ -32,11 +32,15 @@ class App extends React.Component<Props, State> {
...
 
-  async componentDidMount() {
-    await this.loadFonts();
+  afterFontLoaded = () => {
     this.setState({ fontsLoaded: true }, () => {
       Expo.SplashScreen.hide();
     });
+  };
+
+  async componentDidMount() {
```
tbergq: Maybe I am just damaged from having used `airbnb-eslint-config` in the past, but my OCD tells me `componenentDidMount` should be the first function defined an a class.

tbergq: And, there is an eslint-rule for it, `react/sort-comp`, if we want to enforce it 😊

FilipMessa: 👍 thanks I will check that rule

FilipMessa: done

PR #158, opened 2019-01-28
Move state from Search.js to context

tbergq:


apps/core/scenes/search/SearchContext.js
```diff
@@ -0,0 +1,142 @@
...
+        dateTo: date,
+        returnDateFrom: returnDate,
+        returnDateTo: returnDate,
+      };
+    });
+  };
+
+  setReturnDate = (date: Date) => {
+    this.setState(prevState => {
+      const departureDate = DateFNS.min(prevState.returnDateFrom, date);
```
JoseDuSV: It's originally my mistake but it should be `dateFrom` here
```
const departureDate = DateFNS.min(prevState.dateFrom, date);
```

tbergq: 🚀

apps/core/scenes/search/Datepickers.js
```diff
@@ -17,11 +19,10 @@ type Props = {|
 
 type State = {|
   isDatePickerVisible: boolean,
-  datePickerDate: Date,
   selectDate: $Keys<typeof DATEPICKER_MODE> | null,
 |};
 
-const getFormatedDate = date => {
+export const formatDate = (date: Date) => {
```
RobinCsl: Should this be here?

apps/core/scenes/search/Placepickers.js
```diff
@@ -9,36 +9,39 @@ import {
...
-      <TripInput
-        onPress={props.handlePlacePress}
-        label="To"
-        icon={<Icon name="airplane-landing" />}
-        value={props.travelTo}
-      />
-    </View>
-  );
-}
+const handlePlacePress = () => {
```
RobinCsl: Should this be as a property of class Placpickers instead? Or will it not be re-created on ervery re-render?

apps/core/scenes/search/Search.js
```diff
@@ -14,74 +14,39 @@ import {
...
 
+import {
+  withSearchContext,
+  type SearchContextState,
+  type TripTypes,
+  type ModalTypes,
+} from './SearchContext';
 import Placepickers from './Placepickers';
-import Datepickers from './Datepickers';
+import Datepickers, { formatDate } from './Datepickers';
```
RobinCsl: It's a smell for me that we have to import `formatDate` from `Datepickers`. I believe it should be extracted in a new file.

packages/utils/index.js
```diff
@@ -8,3 +8,10 @@ export { formatPrice } from './localization/CurrencyFormatter';
 export type { Price } from './localization/CurrencyFormatter';
 export { default as DateFormatter } from './localization/DateFormatter';
 export { default as DateUtils } from './localization/DateUtils';
+
+// src
+
+export { default as withContext } from './src/withContext/withContext';
+
+// other
+export const noop = () => {};
```
RobinCsl: Is that really necessary?

apps/core/scenes/search/Placepickers.js
tbergq: This is defined in the file, so this function is created only once. 

When we get a proper placepicker, I assume we will wrap this component in a `withNavigation` and handle the on click even in a `class Placepickers`

packages/utils/index.js
tbergq: Well, I expect we will have more contexts, and instead of defining this once in every context, I think it is nice to just import it.

PR #166, opened 2019-01-29
FIX: fix yarn lock

FilipMessa:
This MR fix CI error:

`error https://registry.yarnpkg.com/rxjs/-/rxjs-6.3.3.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "ENOENT: no such file or directory, chmod '/home/circleci/.cache/yarn/v4/npm-rxjs-6.3.3-3c6a7fa420e844a81390fb1158a9ec614f4bad55/node_modules/rxjs/scheduler/queue.js.map'"`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,12 +2,12 @@
 # yarn lockfile v1
 
 
-"@apollographql/apollo-tools@^0.2.6":
-  version "0.2.9"
-  resolved "https://registry.yarnpkg.com/@apollographql/apollo-tools/-/apollo-tools-0.2.9.tgz#1e20999d11728ef47f8f812f2be0426b5dde1a51"
-  integrity sha512-AEIQwPkS0QLbkpb6WyRhV4aOMxuErasp47ABv5niDKOasQH8mrD8JSGKJAHuQxVe4kB8DE9sLRoc5qeQ0KFCHA==
+"@apollographql/apollo-tools@^0.3.0":
+  version "0.3.0"
+  resolved "https://registry.yarnpkg.com/@apollographql/apollo-tools/-/apollo-tools-0.3.0.tgz#45e29010cd1c4ffb0d359d49ad243eba7ea13cba"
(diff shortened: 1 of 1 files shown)


tbergq:
Any reason why the `yarn.lock` is changing? Could it be from different node-versions? If so, maybe we should set engine in package.json?

FilipMessa:
@tbergq 

I found this as a solution here 
https://github.com/yarnpkg/yarn/issues/6312#

tbergq:
Yes, but checking out this branch and doing `yarn install` with node 11 undo the change in yarn.lock

FilipMessa:
Hm 🤔 I will try, but Yes definitely we shoul add an engine into the `package.json`

tbergq:
Happens also on node 10.13 for me, so maybe it is related to yarn version?

FilipMessa:
so something like this?

```
"engines": {
    "node": ">=11.0.0",
    "yarn": ">=1.13.0"
}
```

tbergq:
The CI runs on node 10, so then the config should be upgraded as well.

tbergq:
And even with node `11.6.0` and yarn `1.13.0` My lock file changes when doing yarn on this branch

tbergq:
Hmm, well, my machine adds some more changes to yarn.lock, I will try commit them on top of this, then see what happens

tbergq:
🙈 my machine does not want me to work on margarita

tbergq:
Removing my commit again since problem seems to be on my machine

FilipMessa:
> 🙈 my machine does not want me to work on margarita
Your computer doesn't believe in the RNW? 💔

tbergq:
Seems @RobinCsl also get's another `yarn.lock` so it is not just me, right @RobinCsl ?

JoseDuSV:
Same here, differences:
screenshot 2019-01-29 at 14 28 41

tbergq:
how about #170 in favor of this one?

FilipMessa:
I close this because it should be solved by #170

tbergq:
>I close this because it should be solved by #170

it was not 🙁

PR #170, opened 2019-01-29
Replace concurrently with custom script for runDev

tbergq:
Since rxjs keeps failing on CI and `concurrently` is the one adding rxjs


FilipMessa:
Looks good to me but can you add Description into the commit message why we are do this. It can be same message as you add to the PR `Since rxjs keeps failing on CI and concurrently is the one adding rxjs`

tbergq:
@FilipMessa 🚀

tbergq:
Seems to be also other dependencies adding rxjs 🤕

PR #172, opened 2019-01-29
Add customer bookings query

tbergq:
This will return mocked data for all bookings.
This query will be used to show list of user bookings

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,12 +2,12 @@
 # yarn lockfile v1
 
 
-"@apollographql/apollo-tools@^0.2.6":
-  version "0.2.9"
-  resolved "https://registry.yarnpkg.com/@apollographql/apollo-tools/-/apollo-tools-0.2.9.tgz#1e20999d11728ef47f8f812f2be0426b5dde1a51"
-  integrity sha512-AEIQwPkS0QLbkpb6WyRhV4aOMxuErasp47ABv5niDKOasQH8mrD8JSGKJAHuQxVe4kB8DE9sLRoc5qeQ0KFCHA==
+"@apollographql/apollo-tools@^0.3.0":
+  version "0.3.0"
+  resolved "https://registry.yarnpkg.com/@apollographql/apollo-tools/-/apollo-tools-0.3.0.tgz#45e29010cd1c4ffb0d359d49ad243eba7ea13cba"
+  integrity sha512-Tg0NVtCFHMQkcSX/dqT0m+BNnK9/lbjo4YFNX9W5g3EwczlC0edrleUM/dC4wXw71DwGrGwFiZxWLxqY1ocU5A==
   dependencies:
-    apollo-env "0.2.5"
+    apollo-env "0.3.0"
 
 "@apollographql/graphql-playground-html@^1.6.6":
   version "1.6.6"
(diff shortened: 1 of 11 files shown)


FilipMessa:
I confused with all these changes in `yarn.lock`

tbergq:
> I am confused with all these changes in yarn.lock

You and me both

FilipMessa:
But basically, these changes in `yarn.lock` just updated the dependencies so it should be not any problem.

apps/graphql/src/queries/__tests__/CustomerBookings.test.js
```diff
@@ -0,0 +1,9 @@
+// @flow
```
mrtnzlml: Martín happy. 👏 The test file shouldn't be named like this though. It implies it's only for `customerBookings` query and people tend to copy-paste this test later on. It's better to do something like `queries.test.js` because these five lines can test every query imaginable in fact (just create new fixture file). No need to write this JS test ever again. That's the whole point of this utility... :)

tbergq: I will rename it

yarn.lock
```diff
@@ -9795,13 +9813,13 @@ react-native-linear-gradient@2.4.0:
   dependencies:
     prop-types "^15.5.10"
 
-react-native-maps@expo/react-native-maps#v0.22.1-exp.0:
+"react-native-maps@github:expo/react-native-maps#v0.22.1-exp.0":
   version "0.22.1"
   resolved "https://codeload.github.com/expo/react-native-maps/tar.gz/e6f98ff7272e5d0a7fe974a41f28593af2d77bb2"
 
-"react-native-multi-slider@https://github.com/kiwicom/react-native-multi-slider.git#beb71bf5eb210fed393cb29b18032105b725611e":
+"react-native-multi-slider@git+https://github.com/kiwicom/react-native-multi-slider.git#beb71bf5eb210fed393cb29b18032105b725611e":
```
JoseDuSV: screenshot 2019-01-31 at 10 15 44

After running `yarn install` I see same `yarn.lock` difference that we've tried to solve yesterday.
We can probably merge it and try to solve it in different PR.
Or maybe can you try to edit it manually and run `yarn install` if those `git+` and `github:` will return?

tbergq: > Or maybe can you try to edit it manually

That sounds like a really bad idea.

Also, techincally, `git +` is correct [when installing npm dependency from git](https://stackoverflow.com/questions/17509669/how-to-install-an-npm-package-from-github-directly#answer-17509764)

JoseDuSV: Ok, I was mostly just curious if `yarn install` on your machine will put it back again.
Sadly if I delete my `yarn.lock` and try to create fresh one `git+` is correctly there in mentioned dependencies, but I'm getting even more version differences.

tbergq: this whole `yarn.lock` is confusing 😞 

I did `rm yarn.lock && yarn install` and I had changes to yarn lock. I repeated and then did just `yarn install` and now it does not change. So I pushed the latest version

JoseDuSV: Now if I run `yarn install` I see same three differences.
But if I delete `yarn.lock` the fresh one is ok without any difference. So the "solution" can be to force everyone with this problem to create fresh `yarn.lock` so we can move forward. But it seems to be little bit fragile and I'm slightly worried that it will somehow return again later.

tbergq: Well, the current `yarn.lock` is not working well on the CI, there is a lot of failures without reason. On this branch, the CI has not yet failed once.

FilipMessa: So I vote for merge this branch.

JoseDuSV: I'm also for merging of this branch. But main problem right now is that if I create fresh project from this branch and run `yarn install` I'll get those three differences. And we cannot ask everyone who will try to run this project just to delete `yarn.lock` first.
Damn 😄 , this whole problem doesn't really make sense, when correct file is only created when there is no current one.

FilipMessa: let's add "yarn.lock" to git ignore :D

FilipMessa: We are not only one with this problem https://github.com/yarnpkg/yarn/issues/4379#issuecomment-334948680

FilipMessa: So I think that solution is we can add  `--frozen-lockfile` `true` to `.yarnrc`
But as I think it is not necessary for now and If yarn.lock is updated with just yarn install but the app works it is OK.

FilipMessa: But we should change CI settings to `yarn install --frozen-lockfile`

FilipMessa: What do you think? https://github.com/kiwicom/margarita/pull/176

FilipMessa: @tbergq can you rebase it pls?

tbergq: 🚀

PR #177, opened 2019-01-31
Chore: CircleCI autodeploys GraphQL server to AWS Lambda

RobinCsl:
Summary: Netlify functions were not enough for us, so CircleCI should now deploy our GraphQL server to AWS Lambda when we merge into master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,9 +21,64 @@ jobs:
             - node_modules
           key: v1-dependencies-{{ checksum "yarn.lock" }}
       - run: yarn test-ci
-
+  build-graphql:
+    docker:
+      - image: circleci/node:10
+    working_directory: ~/margarita
+    steps:
+      - checkout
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -13,17 +13,78 @@ jobs:
...
+      steps:
+        - checkout
+        - retore_cache:
+            keys:
+              - v1-dependencies-{{ checksum "yarn.lock" }}
+              - v1-dependencies-
+        - run:
+          name: Install Serverless CLI and dependencies
+          command: |
+            sudo npm i -g serverless
```
tbergq: should we use `yarn global add` instead?

RobinCsl: I took what was written on the CircleCI docs/tutorial. Do you think it makes any difference?

tbergq: Probably not

PR #187, opened 2019-02-01
Add arrival, departure, destinationImageUrl and passenger count

tbergq:
These are the fields necessary for creating the MMB bookingslist

### DIFF
--- a/apps/graphql/src/types/output/RouteStop.js
+++ b/apps/graphql/src/types/output/RouteStop.js
@@ -1,0 +1,21 @@
+// @flow
+
+import { GraphQLObjectType, GraphQLString } from 'graphql';
+
+import GraphQLSegmentTime from './SegmentTime';
+
+export default new GraphQLObjectType({
+  name: 'RouteStop',
+  description: 'Departure or arrival for a segment',
+  fields: {
+    cityName: {
+      type: GraphQLString,
+    },
+    cityId: {
+      type: GraphQLString,
+    },
+    time: {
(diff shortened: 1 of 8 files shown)


apps/graphql/src/types/output/RouteStop.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import { GraphQLObjectType, GraphQLString } from 'graphql';
+
+import GraphQLSegmentTime from './SegmentTime';
+
+export default new GraphQLObjectType({
+  name: 'RouteStop',
+  description: 'Departure or arrival for a segment',
+  fields: {
```
FilipMessa: And what about `city: Location` instead of `cityName` and `cityId`?

tbergq: But not sure it should be the same DataType, since we do not have all the Data that is specified on `Location` this is what data we have:

```
"where": {
            "city_id": "paris_fr",
            "country": "FR",
            "code": "CDG",
            "name": "Paris"
          }
```

FilipMessa: 🤔 we can fetch missing data with resolvers, but will we need it?
So probably will best for know keep with your first Idea, and later we can create the city field when we realize that we need it.

tbergq: > but will we need it?

Probably not

> later we can create the city field when we realize that we need it.

Sounds like the best way to go forward yes 😊

PR #190, opened 2019-02-04
MMB: Add booking list, show cityImage with status and id

tbergq:
For next iterations:
show From -> to, date, passengers
Improve layout for web


tbergq:
Yeah, there should also be a title there, future trips. It will be in one of the next PRs

PR #192, opened 2019-02-04
MMB: Add destination, arrival date and passenger count to booking image

tbergq:


RobinCsl:
Why was it not merged?

Issue #186, opened 2019-02-01
Refreshing the results page should hydrate state


RobinCsl:
Currently, refreshing after a successful display of results leads to `Something went wrong` being displayed and the header shows no locations and `Invalid Date`.

RobinCsl:
After a bit of digging, it works nicely in dev environment, and when deploying it as a static application on Netlify, it breaks.
First answers to be found in that thread: 
https://github.com/zeit/next.js/issues/4663

PR #193, opened 2019-02-04
Fix: (Partial) Hydrate props from query params

RobinCsl:
Summary: When going to /results page from a built nextjs application, getInitialProps and the likes are ignored. That meant that refreshing /results would return Error because the query params were not taken into account.

I used the solution from this comment https://github.com/zeit/next.js/issues/4804#issuecomment-451558569, and it works pretty well.

Caveat is that we show for a brief moment a sort of broken state. That should be improved.

refresh

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,3 +3,5 @@
 **/datasets/**
 **/__generated__/**
 apps/graphql/lambda/dist
+apps/web/out
+apps/web/.next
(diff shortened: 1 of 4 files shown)


RobinCsl:
Fixes #186

RobinCsl:
Was approved by @JoseDuSV, merging.

PR #194, opened 2019-02-04
MMB: Add navigation to booking-detail screen

tbergq:
Fix flow types in withNavigation HOC

### DIFF
--- a/packages/navigation/src/withNavigation.js
+++ b/packages/navigation/src/withNavigation.js
@@ -8,14 +8,16 @@ import type { Navigation, Route } from './types';
 const navigate = (url: Route, params?: Object) =>
   Router.push({ pathname: `/${url}`, query: params });
 
-type Props = {};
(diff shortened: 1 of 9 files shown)


packages/navigation/src/withNavigation.js
```diff
@@ -8,14 +8,16 @@ import type { Navigation, Route } from './types';
...
+type Props = {
+  navigation: Navigation | void,
+};
 
-export const withNavigation = (
-  Component: React.ComponentType<EnhancedProps>,
-): React.ComponentType<Props> =>
-  class extends React.Component<Props> {
+export const withNavigation = <PassedProps: {}>(
+  WrappedComponent: React.ComponentType<PassedProps>,
```
RobinCsl: Thanks for improving this, that seems more like `flow`'s way of doing it 👍

tbergq: 👍

PR #197, opened 2019-02-04
Map ItineraryCard to new GraphQL structure

jvaclavik:
**Done**
- Rename `ConnectionCard` to `ItineraryCard`
- Refactor `ItineraryCard` according to new structure
- `ItineraryCard` with collapsed segments
- Add multiple carrier's logos in card
- Fixing types, lint
- Fixing unused fields in relay


packages/components/package.json
```diff
@@ -6,7 +6,8 @@
   "dependencies": {
     "@kiwicom/margarita-utils": "^0",
     "@kiwicom/universal-components": "0.0.11",
-    "react-native-shimmer-placeholder": "https://github.com/tomzaku/react-native-shimmer-placeholder.git#expo"
+    "react-native-shimmer-placeholder": "https://github.com/tomzaku/react-native-shimmer-placeholder.git#expo",
+    "shortid": "^2.2.14"
```
RobinCsl: Why do we need this? Would there be a way to do without this dependency?

packages/components/src/ItineraryCard/TripSector.js
```diff
@@ -0,0 +1,158 @@
...
+const timeSimpleFormat = 'H:mm';
+const dateFormat = 'ddd D MMM';
+
+const getFormattedDate = (time, format = timeSimpleFormat) =>
+  DateFNS.format(DateFNS.parse(time), format);
+
+const getDuration = durationInMinutes => {
+  return (
+    durationInMinutes &&
+    `${Math.floor(durationInMinutes / 60)}h ${durationInMinutes % 60}m`
```
RobinCsl: This is not going to be translation friendly. Can DateFNS deal with it somehow?

packages/components/index.js
```diff
@@ -27,11 +27,7 @@ export {
 export { PlaceCard } from './src/PlaceCard';
 export { BlackToAlpha };
 export { Shimmer } from './src/Shimmer';
-export { ConnectionCard } from './src/ConnectionCard';
-export type {
-  TripSegmentWithId,
-} from './src/ConnectionCard/ConnectionCardTypes';
-
+export { ItineraryCard, ItineraryCardType } from './src/ItineraryCard';
```
JoseDuSV: `ItineraryCardType` export can be probably removed, it doesn't seem to be used anywhere and is currently causing warning below.

screenshot 2019-02-07 at 16 20 11

packages/components/src/ItineraryCard/TripSector.js
```diff
@@ -0,0 +1,158 @@
...
+  );
+};
+
+const mapTransporters = segments => {
```
JoseDuSV: Can you please move all the helpers and parsers like `mapTransporters`, `getDuration`, `getFormattedDate` into separate file ( for example `TripSectorHelpers`) so I can easily reuse them inside web version.

packages/components/src/ItineraryCard/TripSector.js
jvaclavik: True, I will check it

packages/components/src/ItineraryCard/TripSector.js
jvaclavik: Good idea. I will do it

packages/components/package.json
FilipMessa: The current use case of this is just for the key in `map`.
And we should not generate the Id on GraphQL, because it later can lead to misunderstanding.

packages/components/index.js
jvaclavik: OK, removed

scripts/relay-compiler.js
```diff
@@ -21,7 +21,7 @@ fs.writeFileSync(path.join(__dirname, '..', 'schema.graphql'), clientSchema);
 
 _x('yarn', [
   'relay-compiler',
-  '--src=./apps',
+  '--src=./',
```
tbergq: Why this change?

packages/components/src/ItineraryCard/TripSector.js
jvaclavik: Fixed

packages/components/src/ItineraryCard/TripSector.js
jvaclavik: Fixed. I also had to create new fragment for `transporters`

scripts/relay-compiler.js
jvaclavik: Relic from times when fragment container was in `margarita-components`. 

Fixed :-)

apps/core/components/ItineraryCard/ItineraryCard.js
```diff
@@ -0,0 +1,112 @@
...
+      return (
+        <ItineraryCardRow>
+          <TripSector data={item} />
+        </ItineraryCardRow>
+      );
+    }
+    return null;
+  };
+
+  keyExtractor = () => shortid.generate();
```
tbergq: Please use the id field for this purpose.

apps/core/components/ItineraryCard/TripSector.js
```diff
@@ -3,62 +3,52 @@
...
-  carrier,
-  departure,
-  departureTime,
-  duration,
-  tripDate,
-}: Props) {
+function TripSector({ data }: Props) {
+  const firstSegment = data?.segments?.[0];
+  const lastSegment =
+    data && data.segments && data.segments[data.segments.length - 1];
```
tbergq: please use optionalChaining together with last from rambda here. 

```
const segments = data?.segments ?? [];
const firstSegment = first(segments);
const lastSegment = last(segments);
```

apps/core/components/ItineraryCard/TripSector.js
```diff
@@ -67,6 +57,31 @@ export default function TripSegment({
...
+  graphql`
+    fragment TripSector on Sector {
+      duration
+      segments {
+        arrivalTime {
+          local
+        }
+        departureTime {
+          local
+        }
```
tbergq: You could use another fragment for this, so you don't have to repeat this logic

apps/core/components/ItineraryCard/TripSectorHelpers.js
```diff
@@ -0,0 +1,19 @@
...
+  time: ?string,
+  format: ?string = timeSimpleFormat,
+) => DateFNS.format(DateFNS.parse(time), format);
+
+export const getDuration = (durationInMinutes: ?number) => {
+  if (!durationInMinutes) {
+    return null;
+  }
+  const date = new Date(0, 0, 0, 0, durationInMinutes);
+  return durationInMinutes && DateFNS.format(date, durationFormat);
```
tbergq: do you need durationInMinutoes here, you already checked in the if that it is not falsy

apps/core/components/ItineraryCard/ItineraryCard.js
jvaclavik: But server is not getting you any ID of sector

packages/components/package.json
tbergq: We should generate some ID field for this on the server. We should not use shortID. It beats the purpose of the [key](https://reactjs.org/docs/lists-and-keys.html#keys) prop.

apps/core/components/ItineraryCard/ItineraryCard.js
tbergq: Then we should create an ID for sector

mrtnzlml: You should use stable keys for this purpose. It's not complicated to add a new unique and stable ID. You could as well replace this with `keyExtractor = (_, index) => index;` which does the same thing but without any external library.

apps/core/components/ItineraryCard/TripSector.js
jvaclavik: Nice, it's much better. Thanks!

apps/core/components/ItineraryCard/TripSectorHelpers.js
jvaclavik: You are right. Fixed

apps/core/components/ItineraryCard/TripSector.js
jvaclavik: Could you explain it more, please? I don't know what exactly you mean.

packages/components/package.json
jvaclavik: So, ID is now sector index which is get from GraphQL, is it OK?

tbergq: Anything that uniquely identifies the sector is useable

apps/core/components/ItineraryCard/TripSector.js
tbergq: I can show you on monday 😊

apps/graphql/src/dataloaders/itinerariesHelpers.js
```diff
@@ -61,10 +62,11 @@ export const mapSectors = (
...
 
         // Second map is used for mapping proper GraphQL structure
-        .map(sector => {
-          const firstSegment = sector[0];
-          const lastSegment = sector[sector.length - 1];
+        .map((sector, key) => {
+          const firstSegment = head(sector);
+          const lastSegment = last(sector);
           return {
+            id: `${key}`,
```
tbergq: I am not convinced this will be unique.

apps/core/components/ItineraryCard/ItineraryCard.js
FilipMessa: So is it OK generate ID on the server if there is no exist ID? 
I think that it can lead to confusion because it creates the expectation that you can ask for a particular sector by ID, but it is no truth.

mrtnzlml: Should be fine, yes. ID doesn't say anything about fetching. It just uniquely represents the type (exactly what you need here). Later you could use it for fetching without BC breaks if there is going to be such a query.

As I said - if you don't want it's fine to use just `keyExtractor = (_, index) => index` - it's better than generating random keys every time. Keep it simple.

FilipMessa: @mrtnzlml 
thanks for the explanation, 👍 make it more sense to me now.

PR #200, opened 2019-02-05
Chore: increase --maxWorkers=3

FilipMessa:
Probably should solve failing CI. https://circleci.com/gh/kiwicom/margarita/773

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "lint": "eslint . --report-unused-disable-directives",
     "flow": "flow --include-warnings",
     "test": "node node_modules/jest/bin/jest.js --config=.jest.json",
-    "test-ci": "yarn flow && yarn lint && yarn test --ci --color --maxWorkers=2",
+    "test-ci": "yarn flow && yarn lint && yarn test --ci --color --maxWorkers=3",
     "server": "yarn workspace @kiwicom/margarita-graphql start",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "lint": "eslint . --report-unused-disable-directives",
     "flow": "flow --include-warnings",
     "test": "node node_modules/jest/bin/jest.js --config=.jest.json",
-    "test-ci": "yarn flow && yarn lint && yarn test --ci --color --maxWorkers=2",
+    "test-ci": "yarn flow && yarn lint && yarn test --ci --color --maxWorkers=3",
```
RobinCsl: I am not sure why this would solve anything: --maxWorkers option only applies to `yarn test`, and the issue in CI is spawning the flow server.

FilipMessa:
@RobinCsl I didn't realize that, but strange is when I change this number It is a pass, and after it passes, even on another branch. So maybe something with the cache.

tbergq:
> So maybe something with the cache

I think it is a random-factor. When it fails on a branch, you can hit rerun until it passes.

RobinCsl:
Yes, I believe the cache has an issue. Also, after restarting the CI test a few times, it ended working. So I think CI is not super stable right now.

FilipMessa:
OK it was just experiment so I will close this PR

PR #203, opened 2019-02-05
Fix react-relay flow types

tbergq:


RobinCsl:
Could you add to your commit message a summary to explain that you updated the types with flow-typed and manually added two things at the end of the file?

tbergq:
@RobinCsl 🚀

PR #204, opened 2019-02-05
Fix: CircleCI should run `build-and-deploy-graphql` after `test` is successful

RobinCsl:
~* Summary: We are already running the test when reviewing the PR, so it is not necessary to re-run the test after we merged into `master`; we can go straight to `build-and-deploy-graphql`.~


RobinCsl:
@tbergq 🚀

PR #211, opened 2019-02-05
Feat: SearchForm web layout

### LINKED ISSUE #147: Search form layout on web
_As a user, I wan to see a search form on the website similar to the one I'm used to from Kiwi.com._

Let's improve the experience on the web and use the same layout as we know it from Kiwi.com. 

snimek obrazovky 2019-01-22 v 13 44 05

If it's too difficult to customize the code of the one `way/return` and `passengers` components, feel free to use the ones we already have:

snimek obrazovky 2019-01-25 v 9 30 53

The `origin` and `destination` components would allow just simple text input (as we currently have it), there would be no tags or labels we know from the Kiwi.com website. Also the calendar will be very simple, basically the one we already use on Margarita. The only real change brought by this issue should be the improved layout.

JoseDuSV:
Summary:
Updated search form styles and components structure for responsive web layout.

Added web version of `PassengersButton`.

Added `TripTypeSwitch` component which replaces `TripTypeButton` in web version.

Added `LayoutContext` which provide `layout` prop for responsive layouts.

Closes #147 

resize-480

### DIFF
--- a/apps/core/scenes/search/Datepickers.js
+++ b/apps/core/scenes/search/Datepickers.js
@@ -1,14 +1,24 @@
 // @flow
 
 import * as React from 'react';
-import { View } from 'react-native';
+import { View, Platform } from 'react-native';
 import { TripInput } from '@kiwicom/margarita-components';
 import { Icon, DatePicker, StyleSheet } from '@kiwicom/universal-components';
 import { format, startOfDay } from 'date-fns';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import {
+  withLayoutContext,
+  LAYOUT,
+  type LayoutContextState,
+} from '@kiwicom/margarita-utils';
 
-import { withSearchContext, type SearchContextState } from './SearchContext';
+import {
+  withSearchContext,
+  type TripTypes,
+  type SearchContextState,
+} from './SearchContext';
 import { DATE_FORMAT } from './SearchConstants';
+import PickersWrapper from './PickersWrapper';
 
 type Props = {|
   +tripType: string,
@@ -16,6 +26,8 @@ type Props = {|
   +returnDateFrom: Date,
   +setDepartureDate: Date => void,
   +setReturnDate: Date => void,
+  +setTripType: TripTypes => void,
+  +layout: number,
 |};
 
 type State = {|
@@ -55,12 +67,23 @@ class Datepickers extends React.Component<Props, State> {
   };
 
   handleDateChange = (date: Date) => {
-    switch (this.state.selectDate) {
+    const { selectDate } = this.state;
+    const {
+      tripType,
+      setDepartureDate,
+      setReturnDate,
+      setTripType,
+    } = this.props;
+
+    switch (selectDate) {
       case DATEPICKER_MODE.DEPARTURE:
-        this.props.setDepartureDate(date);
+        setDepartureDate(date);
         break;
       case DATEPICKER_MODE.RETURN:
-        this.props.setReturnDate(date);
+        setReturnDate(date);
+        if (tripType === 'oneWay') {
+          setTripType('return');
+        }
         break;
       default:
         break;
@@ -69,27 +92,36 @@ class Datepickers extends React.Component<Props, State> {
   };
 
   render() {
-    const { tripType, dateFrom, returnDateFrom } = this.props;
+    const { tripType, dateFrom, returnDateFrom, layout } = this.props;
+    const rowLayout = layout >= LAYOUT.largeMobile;
+    const showReturnInput = tripType === 'return' || Platform.OS === 'web';
+    const returnType = tripType === 'return';
     const datePickerDate =
       this.state.selectDate === DATEPICKER_MODE.DEPARTURE
         ? dateFrom
         : returnDateFrom;
+
     return (
       <>
-        <TripInput
-          onPress={this.handleDepartureDatePress}
-          label="Departure"
-          icon={<Icon name="calendar" />}
-          value={format(dateFrom, DATE_FORMAT)}
-        />
-        {tripType === 'return' && (
+        <PickersWrapper layout={layout}>
           <TripInput
-            onPress={this.handleReturnDatePress}
-            label="Return"
+            style={rowLayout && styles.rowInput}
+            onPress={this.handleDepartureDatePress}
+            label="Departure"
             icon={<Icon name="calendar" />}
-            value={format(returnDateFrom, DATE_FORMAT)}
+            value={format(dateFrom, DATE_FORMAT)}
           />
-        )}
+          {showReturnInput && (
+            <TripInput
+              onPress={this.handleReturnDatePress}
+              label={returnType ? 'Return' : ''}
+              icon={<Icon name="calendar" />}
+              value={
+                returnType ? format(returnDateFrom, DATE_FORMAT) : 'No return'
+              }
+            />
+          )}
+        </PickersWrapper>
         <View style={this.state.isDatePickerVisible && styles.picker}>
           <DatePicker
             isVisible={this.state.isDatePickerVisible}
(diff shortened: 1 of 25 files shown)


apps/core/scenes/search/Datepickers.js
```diff
@@ -69,27 +88,35 @@ class Datepickers extends React.Component<Props, State> {
...
-            label="Return"
+            style={rowLayout && styles.rowInput}
+            onPress={this.handleDepartureDatePress}
+            label="Departure"
             icon={<Icon name="calendar" />}
-            value={format(returnDateFrom, DATE_FORMAT)}
+            value={format(dateFrom, DATE_FORMAT)}
           />
-        )}
+          {(returnType || Platform.OS === 'web') && (
```
tbergq: Maybe you could do this check before the return, it would be easier to read. 

`const showReturnInput = tripType === 'return' || Platform.OS === 'web'`

apps/core/scenes/search/PickersWrapper.js
```diff
@@ -0,0 +1,47 @@
...
+export default function PickersWrapper({ children, layout }: Props) {
+  const desktopLayout = layout >= LAYOUT.desktop;
+  const rowLayout = layout >= LAYOUT.largeMobile;
+
+  return (
+    <View
+      style={[
+        styles.container,
+        desktopLayout && styles.desktopContainer,
+        rowLayout && styles.rowContainer,
```
tbergq: Should we think about some more reuasable way of handeling this?

apps/core/scenes/search/Search.js
```diff
@@ -80,36 +87,69 @@ class Search extends React.Component<Props> {
...
+                adults={adults}
+                infants={infants}
                 bags={bags}
               />
             </View>
 
-            <Placepickers />
-            <Datepickers />
+            <View style={desktopLayout && styles.inputsDesktop}>
+              <Placepickers layout={layout} />
```
tbergq: Also why is this prop-drilled down to the component, can it not import the layout there?

apps/core/scenes/search/Search.js
```diff
@@ -80,36 +87,69 @@ class Search extends React.Component<Props> {
...
+                  selectedType={tripType}
+                  onSelect={this.handleTripTypeSelect}
+                />
+              ) : (
+                <TripTypeButton
+                  onPress={this.handleTripTypePress}
+                  icon={TRIP_TYPE[tripType].icon}
+                  label={TRIP_TYPE[tripType].label}
+                />
+              )}
```
tbergq: Maybe we should consider extracting this to a separate component. This component is already too large, and we should not stick more logic in here

JoseDuSV: Good point, definitely, I'll do it.

packages/utils/layoutContext/LayoutContextHelpers.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+import { LAYOUT } from './LayoutConstants';
+
+// $FlowFixMe Object.values() returns mixed[] when it should return number[] https://github.com/facebook/flow/issues/2221
+const layoutValues: Array<number> = Object.values(LAYOUT).sort((a, b) => a - b);
+
+export const getLayout = (windowWidth: number): number => {
```
FilipMessa: I think that it will be better if getLayout will be a pure function. And what about use reduce instead of for each? But it is just an idea. 
something like this:
```
layoutValues.reduce((result, value) => {
  if(windowWidth >= value) {
     return result = value
  }
   return result
}, LAYOUT.largeDesktop  )
```

apps/core/scenes/search/PickersWrapper.js
tbergq: I mean some kind of `media-query` wrapper [react-native-web-docs](https://github.com/necolas/react-native-web/blob/master/docs/guides/style.md#what-about-media-queries)

packages/utils/layoutContext/LayoutContextHelpers.js
JoseDuSV: Sure, updated to reduce. Thanks

apps/core/scenes/search/Search.js
JoseDuSV: `TripType` and `PassengersButton` section extracted into separate component. So search form is way much lighter now.

apps/core/scenes/search/Search.js
JoseDuSV: Prop drilling removed.

apps/core/scenes/search/Datepickers.js
JoseDuSV: Updated. It's definitely more readable. Thanks

apps/core/scenes/search/PickersWrapper.js
JoseDuSV: @tbergq I've been thinking about something like
`<Media query={{ minWidth: LAYOUT.desktop }} style={styles.desktopContainer}>`
I'll try to review it and turn it into something more reusable. But this PR is starting to be quite big, so if you don't mind I'll create new issue for it.

tbergq: Yes that makes sense. New PR, and we can come back and improve this

apps/core/scenes/search/SearchFormModes.js
```diff
@@ -0,0 +1,125 @@
...
+
+type Props = {
+  +tripType: string,
+  +adults: number,
+  +infants: number,
+  +bags: number,
+  +setTripType: TripTypes => void,
+  +setModalType: ModalTypes => void,
+  +layout: number,
+};
```
tbergq: Make the type exact please

JoseDuSV: 👍 sure, updated

packages/components/src/tripTypeSwitch/TripTypeSwitchOption.js
```diff
@@ -0,0 +1,47 @@
...
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import TouchableWithoutFeedback from '../TouchableWithoutFeedback';
+
+type Props = {
+  +type: string,
+  +label: string,
+  +selected: boolean,
+  +onSelect: (type: string) => void,
+};
```
tbergq: Should be exact

JoseDuSV: Thanks, updated

tbergq:
LGTM, but please someone else have a look, since it is quite big 🙈

apps/core/scenes/search/SearchFormModes.js
```diff
@@ -0,0 +1,125 @@
...
+  +adults: number,
+  +infants: number,
+  +bags: number,
+  +setTripType: TripTypes => void,
+  +setModalType: ModalTypes => void,
+  +layout: number,
+|};
+
+class SearchFormModes extends React.Component<Props> {
+  handleTripTypeSelect = (type: string) => {
```
FilipMessa: Do we need to use the switch for something?
Why not just the `if` condition?

packages/components/src/passengersButton/PassengersButton.js
```diff
@@ -6,43 +6,43 @@ import { StyleSheet, Text, Icon } from '@kiwicom/universal-components';
...
+          size="small"
+          color={defaultTokens.paletteInkNormal}
+          name="chevron-down"
+        />
+      </View>
+    </TouchableWithoutFeedback>
+  );
 }
 
+// todo - values from tokens
```
FilipMessa: It looks like this todo is done.

apps/core/scenes/search/SearchFormModes.js
JoseDuSV: Not really. Maybe only thing is that this condition can grow if we add more trip types later.
But for now `if` is definitely shorter and we can refactor it later if it grow bigger with other types.

packages/components/src/passengersButton/PassengersButton.js
JoseDuSV: Sure, removed. Thanks

PR #212, opened 2019-02-05
MMB: Add title to booking list, add better styling for web

tbergq:


apps/core/scenes/bookings/Booking.js
```diff
@@ -21,6 +25,20 @@ type Props = {|
   +navigation: Navigation,
 |};
 
+type TouchWrapperProps = {|
+  +children: React.Node,
+  +onPress: () => void,
+|};
+
+const TouchWrapper = (props: TouchWrapperProps) => {
```
FilipMessa: > // We are not able to style Touchable properly for web, since it does not apply style prop to parent element

In this case shouldn't be this in `@kiwicom/margarita-components`?

tbergq: Why,I don't expect we need to this this anywhere else. And we should change the Touchable component in `universal-components`

RobinCsl: >Why,I don't expect we need to this this anywhere else. And we should change the Touchable component in `universal-components`

Shall we create an issue for it?

tbergq: Yes, please. I was planning to do so 😄

RobinCsl: https://github.com/kiwicom/universal-components/issues/233

PR #215, opened 2019-02-06
MMB: Add booking status header

### LINKED ISSUE #160: MMB: Booking status header
screenshot 2019-01-28 at 14 42 10

Possible status': 
`CLOSED` -> `paletteRedNormal`
`CONFIRMED` -> `paletteGreenNormal` (display travelled if past booking) 
`REFUNDED` -> `paletteInkDark`

tbergq:
closes #160

### DIFF
--- a/apps/core/scenes/bookingDetail/tripDetails/Header.js
+++ b/apps/core/scenes/bookingDetail/tripDetails/Header.js
@@ -1,0 +1,57 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { Text, StyleSheet } from '@kiwicom/universal-components';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import { graphql, createFragmentContainer } from '@kiwicom/margarita-relay';
+
+import type { Header as BookingType } from './__generated__/Header.graphql';
+
+type Props = {|
+  +data: ?BookingType,
+|};
+
+const Header = (props: Props) => {
+  const bookingId = props.data?.bookingId ?? '';
+  return (
+    <View style={[styles.row, styles.container]}>
+      <View style={styles.row}>
+        <View style={styles.circle} />
+        <Text type="success" size="small">
+          {props.data?.status}
+        </Text>
+      </View>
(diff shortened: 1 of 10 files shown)


tbergq:
Tests pass locally 🙈

RobinCsl:
On CI too :trollface:

tbergq:
Yes, after I hit retry

apps/web/pages/mmb/booking-detail.js
```diff
@@ -1,11 +1,31 @@
...
+    <View style={styles.wrapper}>
+      <BookingDetail bookingId={props.router.query.id} />
+    </View>
   </Layout>
 );
+
+const styles = StyleSheet.create({
+  wrapper: {
+    marginTop: defaultTokens.spaceXSmall,
+    marginHorizontal: '25%',
```
RobinCsl: I don't know how this is going to look on small screens?

RobinCsl: And no background color on the web version?

apps/core/scenes/bookingDetail/tripDetails/Header.js
```diff
@@ -0,0 +1,57 @@
...
+      </Text>
+    </View>
+  );
+};
+
+const styles = StyleSheet.create({
+  container: {
+    justifyContent: 'space-between',
+  },
+  circle: {
```
RobinCsl: Should we create a component for that?

apps/web/pages/mmb/booking-detail.js
tbergq: Web MMB has white background, so no need to set one. We should probably improve this yes when we have some general wrapper for media queries.

this is how it will look like on iPhone SE: 

screenshot 2019-02-06 at 12 21 03

apps/core/scenes/bookingDetail/tripDetails/Header.js
tbergq: I don't think we will need it anywhere else, but if we ever do, we can extract it.

PR #221, opened 2019-02-06
Add modal for the place pickers

### LINKED ISSUE #220: Add Modals for PlacePickers (FROM, TO)

FilipMessa:
close #220


apps/core/scenes/search/Placepickers.js
```diff
@@ -9,39 +9,49 @@ import {
...
+        icon={<Icon name="airplane-takeoff" />}
+        value={props.travelFrom}
+      />
+      <TouchableWithoutFeedback onPress={props.handlePlaceSwitchPress}>
+        <View style={styles.placeSwitch}>
+          <Icon name="replace" color="#7F91A8" />
+        </View>
+      </TouchableWithoutFeedback>
+      <TripInput
+        onPress={createHandler(props.setModalType, MODAL_TYPE.DESTINATION)}
```
tbergq: Why this approach over just creating a class and using the class method which will already have access to the `setModalType`

apps/core/scenes/search/SearchModal.js
```diff
@@ -65,6 +70,20 @@ class SearchModal extends React.Component<Props> {
...
         )}
+        {[MODAL_TYPE.ORIGIN, MODAL_TYPE.DESTINATION].includes(
+          this.props.modalType,
+        ) && (
+          <View
+            style={{
+              margin: 25,
+              flexDirection: 'row',
+              alignItems: 'center',
+            }}
```
tbergq: Please create a style with `StyleSheet.create` instead

FilipMessa: this is just temporary so I will change it in next PR

tbergq: ok

apps/core/scenes/search/Placepickers.js
FilipMessa: You mean create a class from PlacePicker? 
And then inside create something like `this.handleFromClick` and `this.handleToClick`.
Do I understand right?

tbergq: Yes

tbergq:
```
class PlacePickers extends React.Component<Props> {
	onPress = () => {this.props.setModalType(type)}
	render() {
		return ()
	}
}
```

FilipMessa: OK it is a good idea so I will change it

FilipMessa: Done

PR #222, opened 2019-02-06
MMB: Add TripInfo component with FromTo row

tbergq:
Move FromTo to components folder in order to
re-use it for TripInfo
Add separator component


packages/components/src/separator/__tests__/Separator.test.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+import * as React from 'react';
+import { shallow } from 'react-native-testing-library';
```
FilipMessa: Why you use `shallow` and not `render`? Is small and atomic component som my point of view render make more sense.

tbergq: 🚀

PR #226, opened 2019-02-07
Chore: Upgrading @kiwicom/universal-components v0.0.11

RobinCsl:
Summary: `release-it` config was broken in @kiwicom/universal-components and v0.0.10 effectively shipped an old build... Some global exports were missing too.


packages/components/src/TimelineInformation/__tests__/__snapshots__/TimelineInformation.test.js.snap
```diff
@@ -58,21 +58,12 @@ exports[`TimelineInformation should match snapshot diff 1`] = `
...
-            "fontSize": 14,
-          },
           Object {
             "fontSize": 14,
           },
           Object {
             "color": "#46B655",
           },
-          Object {
-            "textAlign": "left",
```
RobinCsl: Hopefully that won't be a problem?

PR #228, opened 2019-02-07
Graphql: Refactor type CustomerBooking to BookingInterface

tbergq:
The reason for is to prepare for Different BookingTypes, OneWay, Return and Multicity.
Extends mocked data to cover all possible booking types

And yes it is Big, but it is mostly auto generated files, and tests, and copy paste.


apps/graphql/src/dataloaders/BookingsLoader.js
```diff
@@ -22,10 +22,12 @@ type BookingApiResult = {|
   +flights: $ReadOnlyArray<{|
     +arrival: ApiRouteStop,
     +departure: ApiRouteStop,
+    +return: number,
```
RobinCsl: The API documentation does not say anything about this? Why do we need it?

apps/graphql/src/dataloaders/BookingsLoader.js
```diff
@@ -57,9 +60,26 @@ const sanitizeBookings = (
     arrival: sanitizeRouteStop(last(booking.flights).arrival),
     departure: sanitizeRouteStop(head(booking.flights).departure),
     passengerCount: booking.passengers.length,
+    type: detectType(booking),
   }));
 };
 
+const detectType = (booking: BookingApiResult) => {
+  if (Array.isArray(booking.flights)) {
+    const returnFlight = booking.flights.find(flight => flight.return === 1);
```
RobinCsl: Okay, I just find `return` of type number to be confusing.

apps/graphql/src/queries/__tests__/__snapshots__/queries.test.js.snap
```diff
@@ -92,74 +92,74 @@ fragment RouteStopFragment on RouteStop {
       "edges": [
         {
           "node": {
-            "id": "Q3VzdG9tZXJCb29raW5nOjI3MDcyNTE=",
-            "status": "confirmed",
+            "id": "Qm9va2luZ09uZVdheToxNjQ2MzQ0Nw==",
```
RobinCsl: Are these mocks auto-updated? Where do they come from?

apps/graphql/src/types/output/BookingInterface.js
```diff
@@ -0,0 +1,63 @@
...
+    type: GraphQLRouteStop,
+  },
+  departure: {
+    type: GraphQLRouteStop,
+  },
+  destinationImageUrl: {
+    type: GraphQLString,
+    args: {
+      dimensions: {
+        type: new GraphQLEnumType({
```
RobinCsl: Should extract this GraphQL type to its own file?

apps/graphql/src/types/output/BookingMulticity.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+
+import { GraphQLObjectType } from 'graphql';
+
+import BookingInterface, { commonFields } from './BookingInterface';
+import type { Booking } from '../../dataloaders/BookingsLoader';
+
+const name = 'BookingMulticity';
```
RobinCsl: maybe `NAME`?

apps/graphql/src/types/output/BookingOneWay.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import { GraphQLObjectType } from 'graphql';
+
+import BookingInterface, { commonFields } from './BookingInterface';
+
+const name = 'BookingOneWay';
```
RobinCsl: Idem

apps/graphql/src/types/output/BookingReturn.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+
+import { GraphQLObjectType } from 'graphql';
+
+import BookingInterface, { commonFields } from './BookingInterface';
+import type { Booking } from '../../dataloaders/BookingsLoader';
+
+const name = 'BookingReturn';
```
RobinCsl: Idem

schema.graphql
```diff
@@ -27,26 +26,89 @@ type CustomerBooking {
   departure: RouteStop
   destinationImageUrl(dimensions: BookingDestinationImageDimensions = _600x600): String
   passengerCount: Int
+
+  """OneWay, multicity or return"""
```
RobinCsl: The casing is weird, is there a way to fix it?

apps/graphql/src/dataloaders/BookingsLoader.js
tbergq: Yeah, could be derived from `C` or something 🤷‍♂️

apps/graphql/src/queries/__tests__/__snapshots__/queries.test.js.snap
tbergq: I just added all bookings from the `reactnative@kiwi.com` account, since we have better data there. More bookings, and different types. The old set just contained 3 bookings, all OneWay

apps/graphql/src/types/output/BookingInterface.js
tbergq: It is quite specific for just this one field, but if you think it is better, I can extract it.

RobinCsl: Do you think we won't need to re-use it? If that's the case, then I'm fine with leaving this at it is.

schema.graphql
tbergq: Yes, change the description field of the type 😄

apps/graphql/src/types/output/BookingInterface.js
tbergq: > won't need to re-use it

definitely not

tbergq:
@RobinCsl 🚀

PR #232, opened 2019-02-07
MMB: Add component for showing correct trip type icon

tbergq:
Multicity, return and one way should all have different icon


tbergq:
@RobinCsl @KarinaDavtyan @FilipMessa please review, flow types are ok, just CI that is 🐛 y

tbergq:
> Maybe return a question mark icon instead of null?

No, I think it is better to skip the icon in case of some error. I don't expect that it should happen anyway

PR #235, opened 2019-02-08
Add the simple place picker

FilipMessa:
It looks big but most changes are just about splitting components to smaller.
And because of that, I decided to move it to `components` directory in the next PR.

- Removed PlacePicer scene
- Splited the PlacePicker Component to smaller
- small changes in search context state
- improved style

___________________________________________________________________
For next PRs
- move PlacePicker to the `components` folder
- add bunch of the tests
- implement OptionPicker from the universal components

**PLS SQUASH AND MERGE**


apps/core/scenes/search/PlacePicker/PlacePickerContent.js
```diff
@@ -2,117 +2,133 @@
...
+    setModalType(MODAL_TYPE.HIDDEN);
+  };
+
+  render() {
     return (
       <View style={styles.container}>
         <View>
           <Text weight="bold">{this.getLabel()}:</Text>
           <TextInput
+            autofoucs
```
jvaclavik: `autofoucs` -> `autofocus`

apps/core/scenes/search/PlacePicker/PlacePickerRefetchContainer.js
```diff
@@ -0,0 +1,63 @@
...
+
+// This rule is disabled because this is just temporary the PlacePicker component.
+/* eslint-disable relay/unused-fields */
+
+import {
+  graphql,
+  createRefetchContainer,
+  type RelayRefetchProp,
+} from '@kiwicom/margarita-relay';
+import { debounce } from '@kiwicom/margarita-utils';
```
jvaclavik: Why don't you import `lodash.debounce` directly here?

apps/core/scenes/search/PlacePicker/PlacePickerRefetchContainer.js
```diff
@@ -0,0 +1,63 @@
...
+
+type Props = {
+  locations: ?PlacePickerRefetchContainerType,
+  relay: RelayRefetchProp,
+};
+
+class PlacePickerRefetchContainer extends React.Component<Props> {
+  handleChangeText = debounce((text: string) => {
+    this.props.relay.refetch({ input: { term: text } });
+  }, 250);
```
jvaclavik: Just an idea – we could have numbers like this one in some config. Maybe some day we will need to create another debounced user input.

apps/core/scenes/search/PlacePicker/PlacePickerContent.js
FilipMessa: 🙈

apps/core/scenes/search/PlacePicker/PlacePickerRefetchContainer.js
FilipMessa: Nothing special but utils looks like a good place for it, and if we'll decide change lodash debounce implementation to something else it will be easy, change it just here, of cores the interface must be same.

apps/core/scenes/search/PlacePicker/PlacePickerRefetchContainer.js
FilipMessa: Good idea.

apps/core/scenes/search/PlacePicker/PlaceItem.js
```diff
@@ -0,0 +1,53 @@
...
+
+type Props = {|
+  +onPress: Location => void,
+  +location: ?Location,
+|};
+
+class PlaceItem extends React.Component<Props> {
+  handlePress = () => {
+    const { location, onPress } = this.props;
+    location && onPress(location); // eslint-disable-line babel/no-unused-expressions
```
tbergq: Please don't disable eslint-rules like this, they are there for a reason. 

Also, I don't see why we should props-drill the on press handler down 3 levels when the drilled function just calls functions in the context. 

Just connect to the context in this component.

apps/core/scenes/search/PlacePicker/PlaceItem.js
```diff
@@ -0,0 +1,53 @@
...
+  };
+
+  render() {
+    const { location } = this.props;
+    if (!location || !location.locationId || !location.name) {
+      // eslint-disable-next-line no-console
+      console.error('Incomplete data for the location!');
```
tbergq: This will cause red-screen and probably crash in production build. `console.warn` is probably better, though I am not sure we need any console.log for this 🤔 

also the if-condition can be improved by `!location?.locationId || !location?.name)`

apps/core/scenes/search/PlacePicker/PlacePickerContent.js
```diff
@@ -2,117 +2,133 @@
...
-import type { PlacePickerContent_locations as PlacePickerContentType } from './__generated__/PlacePickerContent_locations.graphql.js';
+import { MODAL_TYPE } from '../SearchConstants';
+import {
+  withSearchContext,
+  type SearchContextState,
+  type Location,
+} from '../SearchContext';
+import PlacePickerList from './PlacePickerList';
+
+type Locations = $ReadOnlyArray<?{|
```
tbergq: Should definitely be a fragment.

apps/core/scenes/search/PlacePicker/PlacePickerList.js
```diff
@@ -0,0 +1,43 @@
...
+
+import * as React from 'react';
+import { ScrollView } from 'react-native';
+import { StyleSheet } from '@kiwicom/universal-components';
+
+import { type Location } from '../SearchContext';
+import PlaceItem from './PlaceItem';
+
+type Locations = $ReadOnlyArray<?{|
+  +node: ?{|
```
tbergq: Create a fragment

apps/core/scenes/search/PlacePicker/PlacePickerRefetchContainer.js
```diff
@@ -0,0 +1,63 @@
...
+      fragment PlacePickerRefetchContainer_locations on RootQuery
+        @argumentDefinitions(
+          input: { type: "LocationsByTermInput!", defaultValue: { term: "" } }
+        ) {
+        locationsByTerm(input: $input) {
+          edges {
+            node {
+              id
+              name
+              locationId
```
tbergq: defer to fragments where these are used, they are not used in this component

apps/core/scenes/search/PlacePicker/PlacePickerRenderer.js
```diff
@@ -4,27 +4,32 @@ import * as React from 'react';
...
     return (
       <QueryRenderer
         query={graphql`
           query PlacePickerRendererQuery($input: LocationsByTermInput!) {
-            ...PlacePickerContent_locations @arguments(input: $input)
+            ...PlacePickerRefetchContainer_locations @arguments(input: $input)
           }
         `}
-        variables={{ input: { term: this.props.defaultValue ?? '' } }}
+        variables={this.setInputVariable()}
```
tbergq: You are not setting anything in this function, you are however getting the variables. The name of the function is misleading.

apps/core/scenes/search/Placepickers.js
```diff
@@ -62,7 +63,7 @@ class Placepickers extends React.Component<Props> {
           onPress={this.handleToPress}
           label="To"
           icon={<Icon name="airplane-landing" />}
-          value={travelTo}
+          value={travelTo?.name || ''}
```
tbergq: Same here

apps/core/scenes/search/SearchContext.js
```diff
@@ -6,6 +6,7 @@ import * as DateFNS from 'date-fns';
 
 import { MODAL_TYPE } from './SearchConstants';
 
+// TODO move types to separate file
```
tbergq: Why?

apps/core/scenes/search/PlacePicker/PlacePickerContent.js
FilipMessa: But currently, there is no feature for this in the TextInput component. So I will delete it. I created the issue in `universal-components` for that. https://github.com/kiwicom/universal-components/issues/238

apps/core/scenes/search/PlacePicker/PlaceItem.js
FilipMessa: Mine idea behind this was about made a log somehow if this happens.


> also the if-condition can be improved by `!location?.locationId || !location?.name)` 
	
Thanks, I always forget about this new feature.

apps/core/scenes/search/PlacePicker/PlacePickerRenderer.js
FilipMessa: You are right 👍, thanks I didn't realize that, but definitely, this is about getting.

apps/core/scenes/search/SearchContext.js
FilipMessa: Thanks for asking, I added this note, and wait if someone has an opinion about that. My idea behind this is, that exists a lot of type definitions already in this file, so I was thinking split it to two files one for code and one for types. It should be more readable but on another side in the current state is easy to import these types together with context.

apps/core/scenes/search/PlacePicker/PlacePickerRefetchContainer.js
FilipMessa: done

apps/core/scenes/search/PlacePicker/PlaceItem.js
FilipMessa: Improved, but I need this pass down through 3 levels because the `onPress(location)` is not the same for every use cases.

apps/core/scenes/search/PlacePicker/PlaceItem.js
FilipMessa: done

apps/core/scenes/search/PlacePicker/PlacePickerContent.js

apps/core/scenes/search/PlacePicker/PlacePickerList.js
FilipMessa: done

apps/core/scenes/search/PlacePicker/PlacePickerRefetchContainer.js
FilipMessa: done

apps/core/scenes/search/PlacePicker/PlacePickerRenderer.js
FilipMessa: done

apps/core/scenes/search/Placepickers.js
```diff
@@ -49,7 +50,7 @@ class Placepickers extends React.Component<Props> {
           onPress={this.handleFromPress}
           label="From"
           icon={<Icon name="airplane-takeoff" />}
-          value={travelFrom}
+          value={travelFrom?.name || ''}
```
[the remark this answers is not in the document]
FilipMessa: done

apps/core/scenes/search/Placepickers.js
FilipMessa: done

apps/core/scenes/search/SearchContext.js
tbergq: The types defined here are describing the state of the Context, nothing more, nothing less, and IMHO it belongs in the same file as the context itself.

apps/core/scenes/search/PlacePicker/PlaceItem.js
tbergq: > I need this pass down through 3 levels because the `onPress(location)` is not the same for every use cases

This is the current click handler

```
const { type, setTravelFrom, setTravelTo, setModalType } = this.props;

    if (type === MODAL_TYPE.ORIGIN) {
      setTravelFrom(location);
    }
    if (type === MODAL_TYPE.DESTINATION) {
      setTravelTo(location);
    }
    setModalType(MODAL_TYPE.HIDDEN);
```

All it does is check some values in the context, then call some values in the context. There is no need for this logic to be defined here. It has nothing to do with this component. 

When I want to know what happens when I click on something, I will look for the click handler in the component where the button is defined, then I have to go back 3 levels to find out what really happens. 

Please move this function to `PlaceItem`

apps/core/scenes/search/PlacePicker/PlacePickerContent.js
tbergq: Fragment composition looks good 😊

apps/core/scenes/search/PlacePicker/PlaceItem.js
FilipMessa: OK, I will change it, but with this approach how you look on *presentational and container components* pattern?

apps/core/scenes/search/SearchContext.js
FilipMessa: OK let's keep these types here.

apps/core/scenes/search/PlacePicker/PlaceItem.js
tbergq: I think this concept is often misunderstood, and done wrong. Just look at our FE where there are `Containers` that are 1000+ lines of code.

You also might see What Dan Abramov says about it in [this](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) old article.

## When to Introduce Containers?
> When you notice that some components don’t use the props they receive but merely forward them down

## Presentational 
• Don’t specify how the data is loaded or mutated
• Receive data and callbacks exclusively via props

## Benefits of This Approach
• You can use the same presentational component with completely different state sources


I would say our presentational components are `@kiwicom/margarita-components` and `@kiwicom/universal-components`. Of course if `PlacePicker.js` was to be reused by some other component, then I would agree it should have the onPress handler passed as a prop. So if some day in the future some other component will reuse this, we could refactor it to take it as a prop.

FilipMessa: Thanks for the explanation. I am still not 100% sure where is the border, but
 
> I would say our presentational components are `@kiwicom/margarita-components` and `@kiwicom/universal-components`.

make sense for me.

PR #236, opened 2019-02-08
Add PassengerCard component

### LINKED ISSUE #120: Component: Create the PassengerCard (Booking Detail)
Create the Passenger card component, it is part of the Booking Detail.
This is a subtask of #73 

the PassengerCart should be inside the `/margarita/packages/components/src/` folder.

Check which components you can use from the [universal-components](https://github.com/kiwicom/universal-components).
Storybook: https://kiwicom-universal-components.netlify.com/

Designs: https://www.figma.com/file/OUhBJ9rPaTbRljptC0lrRrUF/Multistep?node-id=8%3A6

passengercard

without data:

passenger

KarinaDavtyan:
Closed #120

### DIFF
--- a/packages/components/index.js
+++ b/packages/components/index.js
@@ -37,5 +37,6 @@ export { default as Illustration } from './src/Illustration/Illustration';
 export {
   default as IllustrationWithInformation,
 } from './src/IllustrationWithInformation/IllustrationWithInformation';
+export { default as PassengerCard } from './src/passengerCard/PassengerCard';
 
 export { default as Separator } from './src/separator/Separator';
--- a/packages/components/src/passengerCard/BagInformation.js
+++ b/packages/components/src/passengerCard/BagInformation.js
@@ -1,0 +1,28 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { StyleSheet, Text, Icon } from '@kiwicom/universal-components';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+type Props = {|
+  +count: number,
+  +type: string,
+|};
+
+export default function BagInformation({ type, count }: Props) {
+  return (
+    <View style={styles.textWrapper}>
+      <Text>{count} x </Text>
+      <Icon size="small" name="baggage-checked" />
+      <Text>{type}</Text>
+    </View>
+  );
+}
+
+const styles = StyleSheet.create({
+  textWrapper: {
+    flexDirection: 'row',
+    paddingBottom: parseInt(defaultTokens.spaceXXXSmall, 10),
+  },
+});
--- a/packages/components/src/passengerCard/PassengerCard.js
+++ b/packages/components/src/passengerCard/PassengerCard.js
@@ -1,0 +1,143 @@
+// @flow strict
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { Text, StyleSheet, Icon } from '@kiwicom/universal-components';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import BagInformation from './BagInformation';
+import PassengerCardDetail from './PassengerCardDetail';
+import Separator from '../separator/Separator';
+
+type Bag = {|
+  +count: number,
+  +type: string,
+|};
+
+type Props = {|
+  +name: string,
+  +gender: 'female' | 'male' | 'other',
+  +nationality: string,
+  +dateOfBirth: string,
+  +id: string,
+  +insurance: string,
+  +bags: null | Array<Bag>,
+  +passengerCount: number,
+|};
+
+function getTitle(gender) {
+  switch (gender) {
+    case 'male':
+      return 'Mr.';
+    case 'female':
+      return 'Ms.';
+    default:
+      return '';
+  }
+}
+
+export default function PassengerCard(props: Props) {
+  const newPassenger = `${props.passengerCount}. Passenger`;
+  const title = getTitle(props.gender);
+  const passengerWithTitle = `${title} ${props.name}`;
+  const passenger = props.name ? passengerWithTitle : newPassenger;
+
+  const { nationality, dateOfBirth, id, insurance, bags } = props;
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.containerName}>
+        <Icon name="passenger" />
+        <Text size="large">{passenger}</Text>
+      </View>
+      <View style={styles.containerTop}>
+        <PassengerCardDetail
+          value={nationality}
+          label="Nationality"
+          style="normal"
+        />
+        <PassengerCardDetail
+          value={dateOfBirth}
+          label="Date of birth"
+          style="normal"
+        />
+        <PassengerCardDetail value={id} label="ID" style="id_row_wrapper" />
+      </View>
+      <Separator />
+      <View style={styles.containerBottom}>
+        <PassengerCardDetail
+          value={insurance}
+          label="Travel Insurance"
+          style="normal"
+        />
+        <View style={styles.bagsRowWrapper}>
+          <Text type="secondary" style={styles.textPadding}>
+            Bags
+          </Text>
+          <View>
+            {bags &&
+              bags.map(bag => (
+                <BagInformation
+                  key={bag.type}
+                  count={bag.count}
+                  type={bag.type}
+                />
+              ))}
+          </View>
+        </View>
+      </View>
+    </View>
+  );
+}
+
+PassengerCard.defaultProps = {
+  passengerCount: 1,
+  name: '',
+  gender: 'other',
+  nationality: '',
+  dateOfBirth: '',
+  id: '',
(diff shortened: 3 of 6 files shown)


packages/components/src/PassengerCard/PassengerCard.js
```diff
@@ -0,0 +1,135 @@
...
+        <Icon name="passenger" />
+        <Text size="large">{passenger}</Text>
+      </View>
+      <View style={styles.container_top}>
+        <View style={styles.row_wrapper}>
+          <Text type="secondary" style={styles.text_padding}>
+            Nationality
+          </Text>
+          <Text>{nationality}</Text>
+        </View>
```
FilipMessa: What about make component from this?
```
<View style={styles.row_wrapper}>
  <Text type="secondary" style={styles.text_padding}>
     Nationality
   </Text>
   <Text>{nationality}</Text>
 </View>```

File: packages/components/src/PassengerCard/PassengerCard.js [hunk 2]
```diff
@@ -0,0 +1,135 @@
...
+import BagInformation from './BagInformation';
+import Separator from '../separator/Separator';
+
+type Bag = {|
+  +count: number,
+  +type: string,
+|};
+
+type Props = {|
+  +blank: boolean,
```
JoseDuSV: We can make all (just `passengerCount` should be defined) props optional with `''` value (if necessary) as `defaultProps`. And then `blank` prop will not be necessary. 
Then we will need just one condition below for `name` where `X. Passenger` will be used in case of empty name.

packages/components/src/PassengerCard/PassengerCard.js
```diff
@@ -0,0 +1,135 @@
...
+  +dateOfBirth: string,
+  +id: string,
+  +insurance: string,
+  +bags: Array<Bag>,
+  +passengerCount: number,
+|};
+
+export default function PassengerCard(props: Props) {
+  const newPassengerTitle = `${props.passengerCount}. Passenger`;
+  const preffix =
```
jvaclavik: Prefix should be with one `f` 😉

packages/components/src/PassengerCard/PassengerCard.js
```diff
@@ -0,0 +1,135 @@
...
+    props.gender === 'male' ? 'Mr.' : props.gender === 'female' ? 'Ms.' : '';
```
jvaclavik: I would prefer create some function with `switch`. Multiple ternary operators are not easy to read.

packages/components/src/PassengerCard/PassengerCard.js
```diff
@@ -0,0 +1,135 @@
...
+  const passengerTitle = `${preffix} ${props.name}`;
```
jvaclavik: Actually title in english is Mr. or Ms. So this could be something like `passengerWithTitle` and `prefix` could be just `title`, what do you think?

packages/components/src/PassengerCard/PassengerCard.js
```diff
@@ -0,0 +1,135 @@
...
+    borderTopWidth: 2,
+    borderBottomWidth: 2,
+  },
+  container_name: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    paddingTop: parseInt(defaultTokens.spaceSmall, 10),
+    paddingBottom: parseInt(defaultTokens.spaceMedium, 10),
+  },
+  container_top: {
```
jvaclavik: I'm not sure about snake case notation in styles. I think we should keep consistent style and use camelCase as everywhere

packages/components/src/PassengerCard/PassengerCard.js
RobinCsl: @tbergq We need to get some Translation setup going, WDYT? Otherwise, we will need a huge refactor at some point.

packages/components/src/PassengerCard/PassengerCard.js
tbergq: There should be some eslint-rule that verifies camelCase 🤔

packages/components/src/PassengerCard/PassengerCard.js
tbergq: No, this is an example app for developers, they should be able to understand english

packages/components/src/PassengerCard/PassengerCard.js
KarinaDavtyan: Ok! I made PassengerCardDetail component out of this

packages/components/src/PassengerCard/PassengerCard.js
KarinaDavtyan: Will change the case now, eslint rule would be great, didnt notice I was inconsistent in naming

packages/components/src/PassengerCard/PassengerCard.js
KarinaDavtyan: > Actually title in english is Mr. or Ms. So this could be something like `passengerWithTitle` and `prefix` could be just `title`, what do you think?

Yes it would be better

packages/components/src/PassengerCard/PassengerCard.js
KarinaDavtyan: Maybe lets keep this one as it's one line?

packages/components/src/PassengerCard/PassengerCard.js
KarinaDavtyan: Oops😅 well its gonna be renamed to title now

packages/components/src/PassengerCard/PassengerCard.js
RobinCsl: @tbergq  But still, we could showcase how we do translation?

packages/components/src/PassengerCard/PassengerCard.js
KarinaDavtyan: It would be better than with `blank` prop. Thanks!

packages/components/src/PassengerCard/PassengerCard.js
```diff
@@ -0,0 +1,112 @@
...
+  const { nationality, dateOfBirth, id, insurance, bags } = props;
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.containerName}>
+        <Icon name="passenger" />
+        <Text size="large">{passenger}</Text>
+      </View>
+      <View style={styles.containerTop}>
+        <PassengerCardDetail field={nationality} name="Nationality" />
```
jvaclavik: I would use `label` instead of `name` and `value` instead of `field`. It's little bit confusing

KarinaDavtyan: Much better now, thanks!

packages/components/src/PassengerCard/PassengerCard.js
KarinaDavtyan: Changed it to the switch statement

packages/components/src/PassengerCard/PassengerCard.js
tbergq: > we could showcase how we do translation?

IMHO it is not necessary at this point.

packages/components/src/PassengerCard/BagInformation.js
```diff
@@ -0,0 +1,28 @@
...
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+type Props = {|
+  +count: number,
+  +type: string,
+|};
+
+export default function BagInformation({ type, count }: Props) {
+  return (
+    <View key={type} style={styles.textWrapper}>
```
jvaclavik: I think you don't need `key` here

packages/components/src/PassengerCard/PassengerCard.js
jvaclavik: It's beautiful now ❤️  😄

packages/components/src/PassengerCard/BagInformation.js
KarinaDavtyan: Me too, removed now

RobinCsl:
I think you have some issues because of MacOS being case insensitive but other systems are...

I remember this link was useful when I was faced with the same problem: https://coderwall.com/p/mgi8ja/case-sensitive-git-in-mac-os-x-like-a-pro

tbergq:
> I remember this link was useful when I was faced with the same problem

Even easier: 
• `git rm -r cached .`
• `git add . && git commit`

KarinaDavtyan:
> I think you have some issues because of MacOS being case insensitive but other systems are...
> 
> I remember this link was useful when I was faced with the same problem: https://coderwall.com/p/mgi8ja/case-sensitive-git-in-mac-os-x-like-a-pro



> > I remember this link was useful when I was faced with the same problem
> 
> Even easier:
> • `git rm -r --cached .`
> • `git add . && git commit`

I used `git rm -r --cached .`, I think the problem now, are flow issues i didnt expect and fix

PR #239, opened 2019-02-08
Improve responsiveness of LandingPage

KarinaDavtyan:
+ Update responsiveness
+ iframe dimensions fixed
 
+ There is placeholder text


jvaclavik:
screenshot 2019-02-08 at 14 16 25

What about modify section with icons and also section with description as a block elements on low  resolutions?

apps/landingPage/src/components/Demo.js
```diff
@@ -85,6 +85,9 @@ const Container = styled.div`
 
 const HeadingWrapper = styled.div`
   padding-top: 50px;
+  @media (max-width: 600px) {
```
RobinCsl: We should create some media query breakpoints in a different file. Otherwise, changing that in the future could be painful.

apps/landingPage/src/components/Info.js
```diff
@@ -51,15 +53,22 @@ const Container = styled.div`
...
+  }
 `;
 
 const Header = styled.div`
   display: flex;
   flex-direction: column;
   justify-content: space-between;
   align-items: center;
-  height: 250px;
+  height: 250px,
```
RobinCsl: That might be an issue?

KarinaDavtyan: Yes! thanks for noticing

apps/landingPage/src/components/Demo.js
KarinaDavtyan: Agree! What values would that be? is there anything general for kiwi?

RobinCsl: Here's what they do in Nitro Lib
https://github.com/kiwicom/nitrolib/blob/master/src/styles/mq.js
https://github.com/kiwicom/nitrolib/blob/master/src/consts/device.js

KarinaDavtyan: Thanks! Should I copy their breakpoints to the media query config then?

RobinCsl: Sure, why not. We could tweak them later if we need. It definitely is a good starting point.

tbergq: I also created an issue with [orbit-design-tokens](https://github.com/kiwicom/orbit-design-tokens/issues/80) for this

It would be great if it was there.

apps/landingPage/src/components/Footer.js
```diff
@@ -13,23 +13,23 @@ import styled from 'styled-components';
...
-  { title: 'Terms of Use', url: termsOfUse },
-  { title: 'Privacy Policy', url: privacyPolicy },
-  { title: 'Security', url: security },
-];
+// const links = [
+//   { title: 'Terms & Conditions', url: termsAndConditions },
+//   { title: 'Terms of Use', url: termsOfUse },
+//   { title: 'Privacy Policy', url: privacyPolicy },
+//   { title: 'Security', url: security },
+// ];
```
tbergq: why is this code commented?

apps/landingPage/src/components/Info.js
tbergq: This is why wrapping styled components is a good idea, and passing it objects instead 😄 

It eliminates the possibility of this error.

apps/landingPage/src/components/Footer.js
KarinaDavtyan: These links and some other parts, are present in design but still useless placeholders.
And as it was uncertain whether the landing page is shown on presentation or not, I commented it out.

apps/landingPage/src/components/Demo.js
RobinCsl: Okay, so let's wait on `orbit-design-tokens` to figure this out then.

tbergq: Let’s use the ones from nitro-lib in the meanwhile.

KarinaDavtyan: I am adding mediaQueriesConfig with nitrolib breakpoints for now

KarinaDavtyan:
> screenshot 2019-02-08 at 14 16 25
> 
> What about modify section with icons and also section with description as a block elements on low resolutions?
> 
> And btw there are few more design issues:
> 
> * "Get started" headline is not in the center of the page (but the rest of headlines are OK)
> * Text in header menu could be aligned with text on Kiwi.com logo
> 
> But it's not a big deal and it is probably not part of this PR.

Yes there is sth weird with centering, gonna look into that

apps/landingPage/src/components/Footer.js
tbergq: Generally it is better to delete this. If you need to recover it, you can get it from the git-log.

KarinaDavtyan:
> Please remove all commented code, and it should be good to go

All comments removed now

PR #240, opened 2019-02-08
Graphql: Fix arrival and departure for multicty and return

tbergq:
Arrival and departure were only correct for one way trips
Sanitize segments -> will be used later.
Move booking flow types to separate file


apps/graphql/src/dataloaders/bookingsLoader/BookingsLoader.js
```diff
@@ -0,0 +1,156 @@
...
+      inboundSegments.push(segment);
+    } else {
+      outboundSegments.push(segment);
+    }
+
+    return segment;
+  });
+
+  const inbound = {
+    departure: head(inboundSegments)?.departure,
```
RobinCsl: Why is this `?` required only here?

apps/graphql/src/dataloaders/bookingsLoader/BookingsLoader.js
```diff
@@ -0,0 +1,156 @@
...
+  return {
+    isReturn: flight.return === 1,
+    departure: sanitizeRouteStop(flight.departure),
+    arrival: sanitizeRouteStop(flight.arrival),
+  };
+};
+
+const detectType = (booking: BookingApiResult) => {
+  if (Array.isArray(booking.flights)) {
+    const returnFlight = booking.flights.find(flight => flight.return === 1);
```
RobinCsl:
```

Comment (RobinCsl) [inline apps/graphql/src/dataloaders/bookingsLoader/BookingsLoader.js]:
```
const isReturn = flight => flight.return === 1;
```
to make it DRY?

apps/graphql/src/dataloaders/bookingsLoader/BookingsLoader.js
tbergq: If head is passed empty array, it will return undefined. I should use `?` everywhere. I will improve it.

tbergq: Fixed, I also added flow-types for ramda

apps/graphql/src/dataloaders/bookingsLoader/BookingsLoader.js
tbergq: Not really sure what you meant, but re-named the variable and replace find with some to make it clearer what we want from this line of code.

tbergq:
@RobinCsl 🚀

PR #246, opened 2019-02-11
Feat: Manage Menu Group added to Booking Detail screen

RobinCsl:
Summary: I also prepared screens for the corresponding routes. The route configurations might need to be improved, especially when I suspect that the booking id might be necessary for the Share Booking screen for instance.

### DIFF
--- a/apps/core/scenes/bookingDetail/BookingDetail.js
+++ b/apps/core/scenes/bookingDetail/BookingDetail.js
@@ -4,8 +4,8 @@ import * as React from 'react';
 import { QueryRenderer, graphql } from '@kiwicom/margarita-relay';
 import { Text } from '@kiwicom/universal-components';
 
-import TripDetails from './tripDetails/TripDetails';
 import type { BookingDetailQueryResponse } from './__generated__/BookingDetailQuery.graphql';
+import BookingDetailWrapper from './BookingDetailWrapper';
 
 type Props = {|
   +bookingId: ?string,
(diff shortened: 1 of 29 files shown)


apps/core/scenes/bookingDetail/BookingDetail.js
```diff
@@ -13,7 +16,12 @@ type Props = {|
 
 export default class BookingDetail extends React.Component<Props> {
   renderInner = (props: BookingDetailQueryResponse) => {
-    return <TripDetails data={props.bookingDetail} />;
+    return (
+      <ScrollView contentContainerStyle={styles.container}>
```
tbergq: I think we should move this to a separate component, unless this renderInner will turn into a beast

apps/core/scenes/bookingDetail/menuGroups/Manage.js
```diff
@@ -0,0 +1,82 @@
...
+      }
+      case 'android': {
+        shareIcon = 'share-android';
+        break;
+      }
+      default: {
+        shareIcon = 'share';
+        break;
+      }
+    }
```
tbergq: How about creating a ShareIcon component ? We will use this logic also in boarding pass component

apps/web/pages/mmb/other.js
```diff
@@ -0,0 +1,26 @@
...
+const ManageOtherScreen = () => (
+  <Layout>
+    <View style={styles.wrapper}>
+      <ManageOther />
+    </View>
+  </Layout>
+);
+
+const styles = StyleSheet.create({
+  wrapper: {
```
tbergq: Should we consider moving this to layout component?

RobinCsl: I moved it to a local Layout component; it seems specific to MMB views for now.

RobinCsl:
@tbergq 🚀

PR #247, opened 2019-02-11
Graphql: Add arrival, departure and duration to each booking type

tbergq:


apps/graphql/src/dataloaders/bookingsLoader/BookingFlowTypes.js
```diff
@@ -7,12 +7,13 @@ export type ApiRouteStop = {|
     +city_id: string,
   |},
   +when: {|
-    +utc: Date,
-    +local: Date,
+    +utc: number,
```
jvaclavik: Why `number`? I think we should keep it consistent with structure from search result.

tbergq: Because that is what is what data is actually return
https://github.com/kiwicom/margarita/blob/master/apps/graphql/src/datasets/AllBookings.json#L61

apps/graphql/src/dataloaders/bookingsLoader/BookingsLoader.js
```diff
@@ -109,15 +110,18 @@ const sanitizeMulticity = (booking: BookingApiResult) => {
...
+    trips: Object.keys(trips).reduce((acc, key) => {
+      return [...acc, ...trips[key]];
+    }, []),
     departure: head(segments)?.departure,
     arrival: last(segments)?.arrival,
   };
 };
 
-const sanitizeFlight = (flight: ApiFlight) => {
+const sanitizeFlight = (flight: ApiFlight): Segment => {
```
jvaclavik: Probably not issue for this PR, but this Segment could be also more consistent with Segment from search results.

tbergq: Yes, I noticed, I want to change Segment from SearchResult to this structure, but in a later PR

apps/graphql/src/dataloaders/bookingsLoader/BookingFlowTypes.js
jvaclavik: I see, I'm sorry I didn't notice the name of the `const`.

tbergq:
@jvaclavik @RobinCsl 🚀

apps/graphql/src/types/output/Trip.js
```diff
@@ -0,0 +1,36 @@
...
+      type: GraphQLRouteStop,
+      resolve: ({ arrival }: Booking) => arrival,
+    },
+
+    duration: {
+      type: GraphQLInt,
+      resolve: ({ departure, arrival }: Booking) => {
+        const departureRawTime = departure?.time.utc ?? 0;
+        const arrivalRawTime = arrival?.time.utc ?? 0;
+        const departureTime = departureRawTime * 1000;
```
RobinCsl: Could you make precise the units for these computations to be easier to follow?

tbergq:
> but needs rebase

Not really, not sure why github thinks that, because there is no conflict when rebasing. 

@RobinCsl 🚀

PR #248, opened 2019-02-11
Fix: New Text component wrapper for bold and italic to work on mobile

RobinCsl:
Summary: Loading fonts with Expo does not allow the use of `fontStyle` and `fontWeight`, this PR adds a new Text wrapper component to be used instead of @kiwicom/universal-components' Text which takes into account weight='bold' and italic props for iOS and Android.

Note: Once all components use this new Text component, we should add an eslint rule to prevent Text import from '@kiwicom/universal-components'.

### DIFF
--- a/packages/components/src/SearchParamsSummary/SearchParamsSummary.js
+++ b/packages/components/src/SearchParamsSummary/SearchParamsSummary.js
@@ -6,9 +6,10 @@ import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import {
   Icon,
   StyleSheet,
-  Text,
   AdaptableBadge,
 } from '@kiwicom/universal-components';
+
+import Text from '../text/Text';
 
 type Trip = {|
   +city: string,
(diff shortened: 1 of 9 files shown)


packages/components/src/SearchParamsSummary/SearchParamsSummary.js
```diff
@@ -46,9 +47,13 @@ export default function SearchParamsSummary({
     <View style={styles.container}>
       <View style={styles.headerLeftContainer}>
         <View style={styles.citiesContainer}>
-          <Text style={styles.departureCity}>{departure?.city || ''}</Text>
+          <Text weight="bold" style={styles.city}>
+            {departure?.city || ''}
```
tbergq: Do we need this fallback?

packages/components/src/text/Text.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * This wrapper is necessary because loading fonts with Expo does not allow using fontWeight and fontStyle
+ * cf. https://docs.expo.io/versions/latest/guides/using-custom-fonts/
+ */
+
+export default function Text(
+  props: $PropertyType<typeof UniversalComponentsText, 'props'>,
+) {
+  if (Platform.OS !== 'ios' && Platform.OS !== 'android') {
```
tbergq: Will this ever be true?

tbergq: Oh, I read that wrong, but `Platform.OS == 'web'` maybe?

packages/components/src/text/__tests__/Text.web.test.js
```diff
@@ -0,0 +1,34 @@
...
+import { Platform } from 'react-native';
+import { render } from 'react-native-testing-library';
+import { Text as UniversalComponentsText } from '@kiwicom/universal-components';
+
+import Text from '../Text';
+
+const originalPlatform = Platform.OS;
+
+beforeAll(() => {
+  Platform.OS = 'web';
```
tbergq: Maybe we should just use `jest.mock` as in the other tests?

RobinCsl: There was some issue when running Jest; I think react-native-web messes up with `jest.mock('Platform', () => ...)`. I can have another look.

packages/components/src/text/Text.js
RobinCsl: I think in that case, it needs to be the case specifically for Android and iOS since these are our target environments with Expo. Your suggestion would work too, but I guess my proposal is a bit safer for the future? I don't feel so strongly about it, so I can change it if you prefer.

packages/components/src/SearchParamsSummary/SearchParamsSummary.js
RobinCsl: Not sure, this was already in that component. I can try and remove it.

RobinCsl: Removed all of the fallbacks in that file.

packages/components/src/text/Text.js
tbergq: No, let's keep it.

packages/components/src/text/__tests__/Text.web.test.js
tbergq: No, I am ok with this solution as well

PR #249, opened 2019-02-11
Add StopoverDuration

jvaclavik:
screenshot 2019-02-11 at 15 30 55


apps/core/components/ItineraryCard/SectorBorder.js
```diff
@@ -0,0 +1,32 @@
...
+
+const getDuration = (arrivalTime, departureTime) => {
+  return DateFNS.distanceInWords(departureTime, arrivalTime);
+};
+
+export default function SectorBorder({ from, to }: Props) {
+  return (
+    <ItineraryCardRow>
+      <View style={styles.leftShift}>
+        <Badge type="neutral">Returns in {getDuration(from, to)}</Badge>
```
JoseDuSV: This will work only for `return` type. Maybe we should start with more universal `X nights in destination` (skip `then depart from` for now) which is used even for return type on web and then extend it with this `Return in ...` scenario if necessary.

screenshot 2019-02-11 at 16 19 59

jvaclavik: According to the newest design it should be "Returns in ..." (even for nomad), but you are right, it's not universal. I will fix it.

apps/core/components/ItineraryCard/RenderTripSectorItem.js
```diff
@@ -0,0 +1,40 @@
...
+import StopoverDuration from './StopoverDuration';
+import TripSector from './TripSector';
+import ItineraryCardRow from './ItineraryCardRow';
+import type { RenderTripSectorItem as RenderTripSectorItemType } from './__generated__/RenderTripSectorItem.graphql';
+
+type Props = {|
+  +data: ?RenderTripSectorItemType,
+|};
+
+function RenderTripSectorItem({ data, nextSectorDepartureUtc }: Props) {
```
tbergq: `nextSectorDepartureUtc` is no longer a prop here is it?

apps/graphql/src/types/output/Sector.js
```diff
@@ -15,5 +15,9 @@ export default new GraphQLObjectType({
     duration: { type: GraphQLInt },
     origin: { type: Location },
     segments: { type: new GraphQLList(Segment) },
+    stopoverDuration: {
+      type: GraphQLInt,
+      resolve: ({ stopoverDuration }: any): number | null => stopoverDuration,
```
tbergq: I expect this is of the type `Sector` defined in the `ItinerariesloaderTypes.js`, please import the type from there, and replace any with `Sector`

apps/core/components/ItineraryCard/RenderTripSectorItem.js
jvaclavik: Removed

apps/graphql/src/types/output/Sector.js
jvaclavik: Fixed

apps/core/components/ItineraryCard/SectorBorder.js
jvaclavik: Fixed with using prefix "Stays" as we agreed with @JoseDuSV

Issue #112, opened 2019-01-13
Trip detail UI for web


FilipMessa:
Already we have designs for Mobile done, maby it will need some small updates based on designs from the Figma, but we would like to achieve that this component can look different base on the platform (Web, Native) but there will be just one import.

**The  Trip Component is just this part above the result list item, inside the accordion**

**Description of application flow you can find here #70** 

**current code:** https://github.com/kiwicom/margarita/tree/master/packages/components/src/TimelineFlightDetail

**Mobile designs**: https://www.figma.com/file/XD9kQTOJxaqviaxD0g3AhAXS/Booking-Detail?node-id=0%3A1

Web design should be similar like currently exist on www.kiwi.com

**Desktop designs:**

oneway:
<img width="709" alt="screen shot 2019-01-13 at 17 05 13" src="https://user-images.githubusercontent.com/6354326/51087598-a5a19580-1755-11e9-8217-385eca3246e4.png">

return:
<img width="709" alt="screen shot 2019-01-13 at 17 04 03" src="https://user-images.githubusercontent.com/6354326/51087604-b8b46580-1755-11e9-9a2b-33769e27af49.png">

FilipMessa:
close because it will be solved by #70

PR #250, opened 2019-02-11
Refactor: Added ResultsList web layout

JoseDuSV:
Summary: Updated `ResultsList` and `ItineraryCard` web-related styles and conditions.

`TripSector` web layout will be updated in next PR.

@jvaclavik Do you have any plans with `ItineraryCardRow` component? As you can see I just updated its styles but I'm not sure if this wrapper is really necessary. This padding can probably be directly part of the `TripSector` component.

Related to #112

### DIFF
--- a/apps/core/components/ItineraryCard/ItineraryCard.js
+++ b/apps/core/components/ItineraryCard/ItineraryCard.js
@@ -1,8 +1,12 @@
 // @flow
 
 import * as React from 'react';
-import { View, FlatList } from 'react-native';
-import { LocalizedPrice, StyleSheet } from '@kiwicom/universal-components';
+import { View, FlatList, Platform } from 'react-native';
+import {
+  LocalizedPrice,
+  StyleSheet,
+  Icon,
+} from '@kiwicom/universal-components';
 import { formatPrice } from '@kiwicom/margarita-utils';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import { graphql, createFragmentContainer } from '@kiwicom/margarita-relay';
@@ -44,6 +48,7 @@ class ItineraryCard extends React.Component<Props> {
     if (data == null) {
       return null;
     }
+
     // @TODO use real badges
     const badges = [
       {
@@ -63,18 +68,27 @@ class ItineraryCard extends React.Component<Props> {
     };
     return (
       <View style={styles.card}>
-        <View>
+        {Platform.OS === 'web' && (
+          <LocalizedPrice localizedPrice={formatPrice(priceObject)} />
+        )}
+        <View style={styles.sectors}>
           <FlatList
             data={data.sectors}
             keyExtractor={this.keyExtractor}
             renderItem={this.renderSectorItem}
           />
-          <HorizontalDash />
-          <ItineraryCardRow style={styles.lastRow}>
-            <BadgesContainer badges={badges} />
-            <LocalizedPrice localizedPrice={formatPrice(priceObject)} />
-          </ItineraryCardRow>
         </View>
+        {Platform.OS === 'web' ? (
+          <Icon name="chevron-down" color={defaultTokens.paletteInkLighter} />
+        ) : (
+          <>
+            <HorizontalDash />
+            <ItineraryCardRow style={styles.lastRow}>
+              <BadgesContainer badges={badges} />
+              <LocalizedPrice localizedPrice={formatPrice(priceObject)} />
+            </ItineraryCardRow>
+          </>
+        )}
       </View>
     );
   }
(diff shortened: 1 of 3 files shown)


jvaclavik:
> Summary: Updated `ResultsList` and `ItineraryCard` web-related styles and conditions.
> 
> `TripSector` web layout will be updated in next PR.
> 
> @jvaclavik Do you have any plans with `ItineraryCardRow` component? As you can see I just updated its styles but I'm not sure if this wrapper is really necessary. This padding can probably be directly part of the `TripSector` component.

I don't have any special plans with `ItineraryCardRow`, so feel free to delete it.

apps/core/components/ItineraryCard/ItineraryCard.js
```diff
@@ -107,5 +122,35 @@ const styles = StyleSheet.create({
...
+      alignSelf: 'center',
+      flexDirection: 'row',
+      alignItems: 'center',
+      width: '100%',
+      maxWidth: 720,
+      padding: parseInt(defaultTokens.spaceMedium, 10),
+      marginBottom: parseInt(defaultTokens.spaceMedium, 10),
+      borderBottomWidth: 0,
+      borderTopWidth: 0,
+      borderRadius: 3,
```
jvaclavik: You can use `borderRadiusNormal` from `designTokens`

JoseDuSV: Thanks, I thought that there is only 2 and 6. Updated

JoseDuSV:
> > Summary: Updated `ResultsList` and `ItineraryCard` web-related styles and conditions.
> > `TripSector` web layout will be updated in next PR.
> > @jvaclavik Do you have any plans with `ItineraryCardRow` component? As you can see I just updated its styles but I'm not sure if this wrapper is really necessary. This padding can probably be directly part of the `TripSector` component.
> 
> I don't have any special plans with `ItineraryCardRow`, so feel free to delete it.

Ok, thanks. I'll remove it as part of next PR related to `TripSector` web version.

apps/core/components/ItineraryCard/ItineraryCard.js
```diff
@@ -107,5 +122,35 @@ const styles = StyleSheet.create({
...
+    },
+  },
+  sectors: {
+    web: {
+      flex: 1,
+    },
+  },
+  price: {
+    web: {
+      color: '#2e353b',
```
jvaclavik: This color is not in orbit, but maybe it should be there. What do you think? If you agree we could leave here some TODO comment

apps/core/components/ItineraryCard/ItineraryCardRow.js
```diff
@@ -17,5 +17,9 @@ const styles = StyleSheet.create({
   container: {
     paddingHorizontal: 15,
     paddingVertical: 5,
+    web: {
+      paddingHorizontal: 0,
```
RobinCsl: `padding:0` instead?

apps/core/components/ItineraryCard/ItineraryCard.js
```diff
@@ -63,18 +69,27 @@ class ItineraryCard extends React.Component<Props> {
     };
     return (
       <View style={styles.card}>
-        <View>
+        {Platform.OS === 'web' && (
+          <Text style={styles.price}>{formatPrice(priceObject)}</Text>
```
RobinCsl: Should we instead change the `LocalizedPrice` component in @kiwicom/universal-components?

RobinCsl:
Where is the design for this new web layout?

JoseDuSV:
> Where is the design for this new web layout?

Sorry, I should mention it in PR description (now updated) that it is related to #112 .
The design should be as close as possible to current kiwi.com results list.

apps/core/components/ItineraryCard/ItineraryCardRow.js
JoseDuSV: I've tried it but it will not work. The values needs to be overridden directly like this.
I encounter same situation for example even in `ItineraryCard` with `borderBottomWidth` where setting global `borderWidth` to reset border wasn't enough.

apps/core/components/ItineraryCard/ItineraryCard.js
JoseDuSV: I've planned to replace it once the `LocalizedPrice` web style is updated. But you are right, the current design is not that different so I replaced it with `LocalizedPrice` component and will update its web styles in `universal-components`.

apps/core/components/ItineraryCard/ItineraryCard.js
JoseDuSV: Ok, I just replaced this component with `LocalizedPrice` component so I also removed this style. But I'll be updating `LocalizedPrice` with this style so I'll include this comment there.

PR #251, opened 2019-02-11
Fix: ShareIcon passes props to underlying Icon component

RobinCsl:
Summary: This was breaking MenuItem component as cloning ShareIcon did not pass props to underlying icon component.

### DIFF
--- a/packages/components/src/shareIcon/ShareIcon.js
+++ b/packages/components/src/shareIcon/ShareIcon.js
@@ -2,18 +2,26 @@
 
 import * as React from 'react';
 import { Platform } from 'react-native';
-import { Icon } from '@kiwicom/universal-components';
+import { Icon, type StylePropType } from '@kiwicom/universal-components';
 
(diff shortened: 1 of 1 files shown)


packages/components/src/shareIcon/ShareIcon.js
```diff
@@ -4,16 +4,18 @@ import * as React from 'react';
 import { Platform } from 'react-native';
 import { Icon } from '@kiwicom/universal-components';
 
-export default function ShareIcon() {
+type Props = React.ElementProps<typeof Icon>;
```
tbergq: I am not sure about this type, since this gives no error: 

`<ShareIcon pippiLongStock="lol" />`

RobinCsl: Okay, I'll try and find a better one!

RobinCsl:
@tbergq 🚀

PR #260, opened 2019-02-12
MMB: Show trip dates and time for one way booking

tbergq:
Add an interface to be able to re-use FromTo component
for list view and booking detail header.
Add duration component

### DIFF
--- a/apps/graphql/src/types/output/FromToInterface.js
+++ b/apps/graphql/src/types/output/FromToInterface.js
@@ -1,0 +1,23 @@
+// @flow
+
+import { GraphQLInterfaceType } from 'graphql';
+
+import GraphQLRouteStop from './RouteStop';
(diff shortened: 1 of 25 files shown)


apps/graphql/src/types/output/FromToInterface.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import { GraphQLInterfaceType } from 'graphql';
+
+import GraphQLRouteStop from './RouteStop';
+import GraphQLBookingType from './BookingType';
+
+export default new GraphQLInterfaceType({
+  name: 'FromToInterface',
```
RobinCsl: Could you add a description to how `FromToInterface` differs from a Trip, etc.?

tbergq: 🚀

PR #262, opened 2019-02-12
Chore: Use Text component wrapper in @kiwicom/margarita-components

RobinCsl:
Summary: This is a follow up PR to https://github.com/kiwicom/margarita/pull/248; we should use this Text wrapper component to ensure that `fontWeight` bold and `fontStyle` italic get applied correctly on iOS and Android through Expo.

Note that we should use `weight='bold'` and `italic` props instead of passing them through `style` prop (unless we need the text to be bold only on `web` for example).


tbergq:
Could you add the other text-component(s) to eslint-restricted-import to enforce this?

RobinCsl:
@JoseDuSV Absolutely.

@tbergq I planned to do this when I moved all instances of `Text` from `@kiwicom/universal-components` to `@kiwicom/margarita-components`.

PR #264, opened 2019-02-12
MMB: Add return trip info

### LINKED ISSUE #161: MMB: Booking TripInfo Header
screenshot 2019-01-28 at 15 14 35

tbergq:
Add component for rendering correct booking trip type
Add wrapper around date-fns-format to adjust for timezone

closes #161

### DIFF
--- a/apps/core/components/fromTo/FromToWrapper.js
+++ b/apps/core/components/fromTo/FromToWrapper.js
@@ -1,0 +1,23 @@
+// @flow
+
+import * as React from 'react';
+import { StyleSheet } from '@kiwicom/universal-components';
+import { View } from 'react-native';
+
+import FromTo from './FromTo';
+
+type Props = {|
+  +children: React.Element<typeof FromTo>,
+|};
+
+const FromToWrapper = (props: Props) => {
+  return <View style={styles.fromToWrapper}>{props.children}</View>;
+};
+
+const styles = StyleSheet.create({
+  fromToWrapper: {
+    marginBottom: 16,
+  },
+});
+
+export default FromToWrapper;
--- a/apps/core/scenes/bookingDetail/tripDetails/TripInfoReturn.js
+++ b/apps/core/scenes/bookingDetail/tripDetails/TripInfoReturn.js
@@ -1,0 +1,65 @@
+// @flow
+
+import * as React from 'react';
+import { createFragmentContainer, graphql } from '@kiwicom/margarita-relay';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import { StyleSheet, Text } from '@kiwicom/universal-components';
+
+import FromTo from '../../../components/fromTo/FromTo';
+import FromToWrapper from '../../../components/fromTo/FromToWrapper';
+import TripDates from './TripDates';
+import type { TripInfoReturn as BookingType } from './__generated__/TripInfoReturn.graphql';
+
+type Props = {|
+  +data: ?BookingType,
+|};
+
+function TripInfoReturn(props: Props) {
+  return (
+    <>
+      <FromToWrapper>
+        <FromTo
+          textType="attention"
+          fontSize="large"
+          iconColor={defaultTokens.colorIconSecondary}
+          data={props.data}
+          withFlags={true}
+        />
+      </FromToWrapper>
+      <Text style={styles.title} type="secondary">
+        DEPARTURE
+      </Text>
+      <TripDates data={props.data?.outbound} />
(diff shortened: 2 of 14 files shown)


apps/core/scenes/bookingDetail/tripDetails/TripInfoReturn.js
```diff
@@ -0,0 +1,65 @@
...
+  return: {
+    marginTop: 12,
+  },
+});
+
+export default createFragmentContainer(
+  TripInfoReturn,
+  graphql`
+    fragment TripInfoReturn on BookingInterface {
+      ... on BookingReturn {
```
jvaclavik: What exactly this mean?

tbergq: BookingInterface can resolve to different Object types runtime, it means that this fragment is only when it resolves to the type BookingReturn

tbergq:
@jvaclavik @RobinCsl 🙏

apps/core/scenes/bookingDetail/tripDetails/TripInfoReturn.js
```diff
@@ -0,0 +1,65 @@
...
+        <FromTo
+          textType="attention"
+          fontSize="large"
+          iconColor={defaultTokens.colorIconSecondary}
+          data={props.data}
+          withFlags={true}
+        />
+      </FromToWrapper>
+      <Text style={styles.title} type="secondary">
+        DEPARTURE
```
jvaclavik: Maybe we should add some uppercase prop to `Text` component, but it's not very important.

packages/components/src/bookingTypeRenderer/__tests__/BookingTypeRenderer.test.js
```diff
@@ -0,0 +1,42 @@
...
+import * as React from 'react';
+import { render } from 'react-native-testing-library';
+import { View } from 'react-native';
+
+import BookingTypeRenderer from '../BookingTypeRenderer';
+
+const getComponent = (type: string) => (
+  <BookingTypeRenderer
+    type={type}
+    oneWayComponent={<View testID="one-way" />}
```
jvaclavik: Why not to use camelCase notation here?

I noticed that we use many variations of those flight types.
Right now you can find in the project at least following variations: `oneway, oneWay, OneWay`

apps/core/scenes/bookingDetail/tripDetails/TripInfoReturn.js
tbergq: Maybe, but we can also handle that with styles 🤷‍♂️

packages/components/src/bookingTypeRenderer/__tests__/BookingTypeRenderer.test.js
tbergq: You are right, I will change it to camelCase

tbergq: 🚀

apps/core/components/fromTo/FromToWrapper.js
```diff
@@ -0,0 +1,23 @@
+// @flow
```
RobinCsl: Do we really need a component for this? Maybe the style could be added to FromTo component instead?

apps/core/scenes/bookingDetail/tripDetails/TripInfoReturn.js
jvaclavik: Yes, it should be also possible, but I'm not sure if it's not buggy (last comment here: https://stackoverflow.com/questions/35813156/how-to-set-text-text-to-upper-case-in-react-native)

apps/core/components/fromTo/FromToWrapper.js
tbergq: The thing is that OneWay and multicity can Reuse the TripInfo as is. But return is slightly different, so it needs to reuse the internals of the TripInfo component. 

So instead of copy/pasting this style and having it duplicated, I extracted it.

apps/core/scenes/bookingDetail/tripDetails/TripInfoReturn.js
jvaclavik: Btw I noticed that we have prop for `uppercase` in `universal-components` (implemented by styles)

apps/core/components/fromTo/FromToWrapper.js
RobinCsl: Okay, why not.

PR #273, opened 2019-02-13
MMB: Booking badges correct background color

tbergq:
Currently we showed green background no matter status

### DIFF
--- a/apps/core/scenes/bookings/BookingBadges.js
+++ b/apps/core/scenes/bookings/BookingBadges.js
@@ -9,6 +9,7 @@ import {
 } from '@kiwicom/universal-components';
 import { createFragmentContainer, graphql } from '@kiwicom/margarita-relay';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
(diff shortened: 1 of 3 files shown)


apps/core/scenes/bookings/BookingBadges.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 
+/* eslint-disable react-native/no-unused-styles */ // Style is not unused, eslint just don't pick it up when used this way
```
RobinCsl: One way to do that otherwise is to be a bit more verbose and add a switch statement to make sure all styles are being used. Not sure what's the best approach here.

RobinCsl: At least with the switch, there are no surprises.

tbergq: If the switch shut's up the eslint rule, that is definitely better. I will give it a shot

tbergq:
@RobinCsl 🚀

PR #274, opened 2019-02-13
Fix: Fixed failing flow on CircleCI

JoseDuSV:
Summary: Added `max_workers` limit to flow configuration to prevent CircleCI consuming too many resources.

Found this solution after investigating CircleCI logs.
Where we have similar errors as mentioned in this issue https://github.com/flowtype/flow-bin/issues/138 with suggested solution.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -62,6 +62,7 @@ node_modules/react-native/flow-github/
 flow-typed/
 
 [options]
+server.max_workers=4
 emoji=true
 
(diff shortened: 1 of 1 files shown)


tbergq:
pipeline ok 2 times in a row 🤞 for this being the solution 🎉

JoseDuSV:
It now passed 3 times in a row. So hopefully it's not just some cleaned cache like last time and it will solve our issue.

PR #275, opened 2019-02-13
Chore: Upgrading @kiwicom/universal-components v0.0.13

RobinCsl:
Summary: Now that Flow types are correctly being picked up, there were quite a few Flow errors to fix.
Some are related to using react-native-web flow-typed in this repo and they conflict with `@kiwicom/universal-components` in `node_modules`. These will have to be fixed there and they are ignored in this repo in .flowconfig file.
There was also an issue with how Jest resolves `@kiwicom/universal-components`: it only looks at the `main` field in `package.json`, and not at `react-native` field if preset is `jest-expo` or `react-native` in the config file`.jest.json`. That means that Jest resolved the web version of the components instead of the RN version.

### DIFF
--- a/packages/components/src/TimelineFlightDetail/TimelineFlightDetail.js
+++ b/packages/components/src/TimelineFlightDetail/TimelineFlightDetail.js
@@ -13,9 +13,9 @@ import CardHeader from './components/CardHeader';
 type Props = {|
   +additionalInfo: $PropertyType<AdditionalInfoProps, 'additionalInfo'>,
   +carrier: {|
-    +code: string,
-    +name: string,
-    +type?: 'airline' | 'bus' | 'train',
+    code: string,
+    name: string,
+    type?: 'airline' | 'bus' | 'train',
   |},
   +duration: string,
 |};
@@ -48,7 +48,7 @@ export default function TimelineFlightDetail({
       </View>
       <View style={styles.cardContainer}>
         <Accordion
-          defaultExpanded={true}
+          expandedDefault={true}
           header={Header({ carrier, duration })}
         >
           <AdditionalInfo additionalInfo={additionalInfo} />
--- a/packages/components/src/text/Text.js
+++ b/packages/components/src/text/Text.js
@@ -1,10 +1,9 @@
 // @flow
 
 import * as React from 'react';
-import { Platform } from 'react-native';
 import {
   Text as UniversalComponentsText,
-  StyleSheet,
+  type StylePropType,
 } from '@kiwicom/universal-components';
 
 /**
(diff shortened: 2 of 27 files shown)


apps/core/components/ItineraryCard/LocationName.js
```diff
@@ -14,7 +14,7 @@ type Props = {|
 function LocationName({ data, style }: Props) {
   return (
     <Text style={style} numberOfLines={1}>
-      {data?.name}
+      {data?.name ?? ''}
```
tbergq: Sounds like we should accept `?React.Node` here, so we don't have to do this everywhere 🙈

packages/components/src/text/Text.js
```diff
@@ -1,50 +1,37 @@
...
-export default function Text(
-  props: $PropertyType<typeof UniversalComponentsText, 'props'>,
-) {
-  if (Platform.OS !== 'ios' && Platform.OS !== 'android') {
-    return <UniversalComponentsText {...props} />;
-  }
-  const { italic, weight } = props;
+type Props = {|
+  align?: 'left' | 'right' | 'center' | 'justify',
+  +children: React.Node,
```
tbergq: Probably `?React.Node` would be better.

packages/components/src/tripInput/__tests__/TripInput.test.js
```diff
@@ -12,7 +12,10 @@ it('renders', () => {
     shallow(
       <TripInput
         onPress={jest.fn()}
-        icon={<IconMock />}
+        icon={
+          // $FlowFixMe Expected Error
```
tbergq: Could you use `$FlowExpectedError: ` with a comment of why you expect the error?

JoseDuSV:
https://github.com/kiwicom/margarita/blob/c7e8b340230268867330c47f03fc1f1b759debfc/packages/components/src/SearchParamsSummary/SearchParamsSummary.js#L63

Flow is still showing me 3 errors inside `SearchParamsSummary` for each `AdaptableBadge` text prop. But it will probably be solved by Tronds `?React.Node` suggestion.

RobinCsl:
Yes, I planned to modify a few flow types in universal components and release a new patch version for these to be fixed.
cf https://github.com/kiwicom/universal-components/pull/244

packages/components/src/TimelineFlightDetail/TimelineFlightDetail.js
```diff
@@ -13,9 +13,9 @@ import CardHeader from './components/CardHeader';
 type Props = {|
   +additionalInfo: $PropertyType<AdditionalInfoProps, 'additionalInfo'>,
   +carrier: {|
-    +code: string,
-    +name: string,
-    +type?: 'airline' | 'bus' | 'train',
+    code: string,
```
jvaclavik: Why without `+`?

RobinCsl: Because the current types are as follows in universal-components:
https://github.com/kiwicom/universal-components/blob/a252171631194d56ae42d1785f620cd6116a4c61/src/CarrierLogo/CarrierLogoTypes.js#L3

RobinCsl:
There are still some issues with tests. For example:
```
FAIL  packages/components/src/tripTypeSwitch/__tests__/TripTypeSwitch.test.js
  ● Test suite failed to run

    Cannot find module './Button' from 'index.js'

    However, Jest was able to find:
        './Button.web.js'
        './Button.web.js.flow'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['js', 'json', 'jsx', 'node', 'ts', 'tsx'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string

    However, Jest was able to find:
        '../TripTypeSwitch.js'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['js', 'json', 'jsx', 'node', 'ts', 'tsx'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string

      at Resolver.resolveModule (node_modules/jest-runner/node_modules/jest-resolve/build/index.js:203:17)
      at Object.<anonymous> (node_modules/@kiwicom/universal-components/lib/web/Button/index.js:13:38)
```

I did try to add `web.js` to the list of extensions in our Jest config, and it did not help. To be continued.

packages/components/src/tripInput/__tests__/TripInput.test.js
```diff
@@ -12,7 +12,10 @@ it('renders', () => {
     shallow(
       <TripInput
         onPress={jest.fn()}
-        icon={<IconMock />}
+        icon={
+          // $FlowExpectedError: TripInput expects Icon component
+          <IconMock />
```
JoseDuSV: Cannot we use `Icon` component as `IconMock` instead of `null` to get rid of all `FlowExpectedError`?

RobinCsl: I guess the idea behind using this mock is to avoid refactoring this snapshot test if the `Icon` component's props change.
@tbergq Could you provide more context?

JoseDuSV: @RobinCsl Thanks, that makes sense.

PR #277, opened 2019-02-13
Fix autocorrect in PlacePicker

jvaclavik:
- enable autoFocus
- turn off autocorrect: It's annoying to write locations on iOS with autocorrect:

screenshot 2019-02-13 at 11 34 28

### DIFF
--- a/apps/core/scenes/search/PlacePicker/PlacePickerContent.js
+++ b/apps/core/scenes/search/PlacePicker/PlacePickerContent.js
@@ -82,6 +82,8 @@ class PlacePickerContent extends React.Component<Props, State> {
             value={this.state.text}
             prefix={<Icon name="search" />}
             label={<Text weight="bold">{this.getLabel()}:</Text>}
+            autoCorrect={false}
+            autoFocus
           />
         </View>
         <PlacePickerList locations={this.props.locations?.locationsByTerm} />
--- a/apps/core/scenes/search/PlacePicker/PlacePickerList.js
+++ b/apps/core/scenes/search/PlacePicker/PlacePickerList.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react';
-import { View, FlatList } from 'react-native';
+import { View, FlatList, Keyboard } from 'react-native';
 import { StyleSheet } from '@kiwicom/universal-components';
 import { graphql, createFragmentContainer } from '@kiwicom/margarita-relay';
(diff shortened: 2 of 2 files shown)


apps/core/scenes/search/PlacePicker/PlacePickerContent.js
```diff
@@ -83,10 +83,11 @@ class PlacePickerContent extends React.Component<Props, State> {
         <View>
           <Text weight="bold">{this.getLabel()}:</Text>
           <TextInput
-            placeholder="write a place..."
+            placeholder="Write a place..."
             onChangeText={this.handleChangeText}
             value={this.state.text}
             prefix={<Icon name="search" />}
+            autoCorrect={false}
```
JoseDuSV: It doesn't seem to work. Sadly we are using `universal-components` version of `TextInput`.
Where this props is not passed down. So we will need to update it first.

FilipMessa: I will do that, I currently working on autoFocus for TextInput so this should be easy.

RobinCsl: cf https://github.com/kiwicom/universal-components/pull/245

tbergq:
Looks good, just a little improvement which I think could be done within this PR. It is standard behaviour that you dismiss the keyboard on scroll. Now you cannot select the bottom options, because they are covered by the keyboard. 

just add `onScroll={Keyboard.dismiss}` to the `FlatList`

jvaclavik:
> Looks good, just a little improvement which I think could be done within this PR. It is standard behaviour that you dismiss the keyboard on scroll. Now you cannot select the bottom options, because they are covered by the keyboard.
> 
> just add `onScroll={Keyboard.dismiss}` to the `FlatList`


I was thinking about adding `keyboardShouldPersistTaps="always"` to FlatList. You could press the item without necessity of dismissing keyboard. Maybe we could combine it with your proposed change, what do you think? I tested both in the simulator and it works fine.

tbergq:
> I was thinking about adding `keyboardShouldPersistTaps="always"`

Isn't the current behaviour that the modal closes when I tap a location?

jvaclavik:
Right now it works like in the Kiwi.com app. When you scroll it closes the keyboard and when you don't scroll and click on the item, location is picked (you don't have to click once to close keyboard and then click to pick the location).

PR #283, opened 2019-02-13
Refactor: Added ItineraryCardWrapper component

JoseDuSV:
Summary: Added responsive layout to web version of `ItineraryCard`.
Extracted platform specific parts into separate `ItineraryCardWrapper` component. Because original component started to be hard to read with condition combinations like: web platform && mobile layout.

screenshot 2019-02-13 at 15 40 39

### DIFF
--- a/apps/core/components/ItineraryCard/ItineraryCard.js
+++ b/apps/core/components/ItineraryCard/ItineraryCard.js
@@ -1,20 +1,12 @@
 // @flow
 
 import * as React from 'react';
-import { View, FlatList, Platform } from 'react-native';
-import {
-  LocalizedPrice,
-  StyleSheet,
-  Icon,
-} from '@kiwicom/universal-components';
+import { FlatList } from 'react-native';
 import { formatPrice } from '@kiwicom/margarita-utils';
-import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import { graphql, createFragmentContainer } from '@kiwicom/margarita-relay';
 
-import ItineraryCardRow from './ItineraryCardRow';
-import BadgesContainer from './BadgesContainer';
-import HorizontalDash from './HorizontalDash';
 import RenderTripSectorItem from './RenderTripSectorItem';
+import ItineraryCardWrapper from './ItineraryCardWrapper';
 import type { ItineraryCard as ItineraryCardType } from './__generated__/ItineraryCard.graphql';
 import type { RenderTripSectorItem as RenderTripSectorItemType } from './__generated__/RenderTripSectorItem.graphql';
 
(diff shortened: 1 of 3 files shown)


apps/core/components/ItineraryCard/ItineraryCardWrapper.js
```diff
@@ -0,0 +1,91 @@
...
+const styles = StyleSheet.create({
+  card: {
+    backgroundColor: defaultTokens.backgroundCard,
+    alignSelf: 'center',
+    width: '100%',
+    maxWidth: 720,
+    padding: parseInt(defaultTokens.spaceMedium, 10),
+    marginBottom: parseInt(defaultTokens.spaceMedium, 10),
+    borderRadius: parseInt(defaultTokens.borderRadiusNormal, 10),
+    boxShadow: '0 2px 4px 0 rgba(23,27,30,.1)',
```
jvaclavik: Does this work on RN?

apps/core/components/ItineraryCard/ItineraryCardWrapper.js
```diff
@@ -0,0 +1,91 @@
...
+  },
+  footer: {
+    marginTop: parseInt(defaultTokens.spaceMedium, 10),
+    paddingTop: parseInt(defaultTokens.spaceMedium, 10),
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'space-between',
+    borderStyle: 'dotted',
+    borderColor: '#d2d9e0', // @TODO should be added to design-tokens
+    borderTopWidth: 1,
```
jvaclavik: It could be also from my PR 😄 but we could use here `defaultTokens.borderWidthCard` what do you think?

apps/core/components/ItineraryCard/ItineraryCardWrapper.js
jvaclavik: I get it now, it's web 😄 I'm still confused that `.native` means react native and without that it's web.

apps/core/components/ItineraryCard/ItineraryCardWrapper.js
JoseDuSV: Ok, updated. This one is mine. But I also updated `borderBottomWidth` and `borderTopWidth` in native version from older PR with this token 😉

apps/core/components/ItineraryCard/ItineraryCardWrapper.js
JoseDuSV: Good point. I renamed web version file to `.web.js` so it should be more clear even on first glance now.

jvaclavik: It's much better right now, thanks!

PR #296, opened 2019-02-14
Eslint: Enable import/no-extraneous-dependencies

tbergq:
Fix missing dependencies

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -21,25 +21,32 @@
   "dependencies": {
     "@kiwicom/margarita-core": "^0",
     "@kiwicom/margarita-relay": "^0",
+    "@kiwicom/margarita-utils": "^0",
+    "@kiwicom/margarita-components": "^0",
     "file-loader": "^3.0.1",
     "next": "^7.0.2",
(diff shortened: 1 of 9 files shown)


apps/web/package.json
```diff
@@ -21,26 +21,33 @@
...
+    "@kiwicom/margarita-components": "^0",
     "file-loader": "^3.0.1",
     "next": "^7.0.2",
     "next-images": "^1.0.4",
     "next-plugin-transpile-modules": "^0.1.3",
     "react-native-web": "^0.10.0",
     "schedule": "^0.4.0",
-    "url-search-params-polyfill": "^5.0.0"
+    "url-search-params-polyfill": "^5.0.0",
+    "webpack-bundle-analyzer": "^3.0.4"
```
JoseDuSV: Does this one needs to be in `dependencies` and not in `devDependencies`?

tbergq: It will generate an eslint-error if it is not, so yes

JoseDuSV: Oh, I see. Thanks
So approved 🚀

PR #300, opened 2019-02-14
Use '@kiwicom/relay', remove flow-typed Relay types

mrtnzlml:
Please verify that everything works correctly before merging. Also, I am just showing you here what to do. You should now cleanup the project a little bit. The idea is that you are not going to use Relay directly but you'll always use `@kiwicom/relay` package.


tbergq:
Verified that it works, I will check for any usage of react-relay and clean it out.

PR #309, opened 2019-02-15
Refactor: TripSector web layout

JoseDuSV:
Summary: Updated `TripSector` layout for web.

screenshot 2019-02-15 at 11 08 41

Note: Because of Relay fragments I wasn't able to use our usual approach with `TripSector.native/web.js`. And which component should be rendered is handled inside `RenderTripSectorItem`. Reason for this is that even that component content is similar for both platforms, the layout and nesting is completely different.
The other option is to merge both component into one `TripSector` and decide which layout should be used inside. But I find this approach slightly cleaner because we have both layouts separated to smaller components/files so it's easier to read.

### DIFF
--- a/apps/core/components/ItineraryCard/StopoverDuration.js
+++ b/apps/core/components/ItineraryCard/StopoverDuration.js
@@ -1,11 +1,9 @@
 // @flow
 
 import * as React from 'react';
-import { StyleSheet, Badge } from '@kiwicom/universal-components';
-import { View } from 'react-native';
 import * as DateFNS from 'date-fns';
 
-import ItineraryCardRow from './ItineraryCardRow';
+import StopoverDurationWrapper from './StopoverDurationWrapper';
 
 type Props = {|
   +stopoverDuration: ?number,
@@ -22,19 +20,14 @@ export default function StopoverDuration({
   stopoverDuration,
   locationName,
 }: Props) {
-  return stopoverDuration ? (
-    <ItineraryCardRow>
-      <View style={styles.leftShift}>
-        <Badge type="neutral">
-          Stays {getDuration(stopoverDuration)}
-          {locationName && ` in ${locationName}`}
-        </Badge>
-      </View>
-    </ItineraryCardRow>
-  ) : null;
+  if (stopoverDuration == null) {
+    return null;
+  }
+
+  return (
+    <StopoverDurationWrapper>
+      Stays {getDuration(stopoverDuration)}
+      {locationName && ` in ${locationName}`}
+    </StopoverDurationWrapper>
+  );
 }
-const styles = StyleSheet.create({
-  leftShift: {
-    paddingStart: 103,
-  },
-});
--- a/apps/core/components/ItineraryCard/FlightTimes.js
+++ b/apps/core/components/ItineraryCard/FlightTimes.js
@@ -1,8 +1,9 @@
 // @flow
 
 import * as React from 'react';
+import { View, Platform } from 'react-native';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
-import { StyleSheet } from '@kiwicom/universal-components';
+import { StyleSheet, Text } from '@kiwicom/universal-components';
 import { graphql, createFragmentContainer } from '@kiwicom/margarita-relay';
 
 import type { FlightTimes as FlightTimesType } from './__generated__/FlightTimes.graphql';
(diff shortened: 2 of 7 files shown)


FilipMessa:
I prefer to stay with split components for TripSector, but I am not sure if we need two separated fragment for that, on the other hand, why not?  

@tbergq You have a strong opinion about how work with relay fragments, so what do you think about this approach?

apps/core/components/ItineraryCard/StopoverDuration.js
```diff
@@ -23,18 +21,9 @@ export default function StopoverDuration({
...
-          Stays {getDuration(stopoverDuration)}
-          {locationName && ` in ${locationName}`}
-        </Badge>
-      </View>
-    </ItineraryCardRow>
+    <StopoverDurationContent>
+      Stays {getDuration(stopoverDuration)}
+      {locationName && ` in ${locationName}`}
+    </StopoverDurationContent>
   ) : null;
```
tbergq: I think 

```
if (stopoverDuration == null) {
	return null;
}

return (
...
)
```

is more readable

tbergq:
It does not look optimal to keep those fragments duplicated over 2 components. As far as I can see, there is just some differences in styling and a couple of extra components added. 

Maybe it would be better to use our stylesheet to tweak the styles for web and native. And only show those extra components if Platform.OS === 'web'

JoseDuSV:
@tbergq @FilipMessa I'm also not that happy with this solution. I've tried to keep everything in one file, but the problem is that the structure (components order) is quite different even if it's using same elements. And with all the conditions, plus responsivity for web rules quickly became quite unreadable.

But the question is how much strict we need to be with design, to have it same as on kiwi.com. I think that with just slight change in components order (basically keep the same order as on mobile, just render them with different styles) we can manage everything in one file.

Something like (just a quick sketch, font sizes and colors will be unified):
screenshot 2019-02-18 at 16 13 22

What do you think?

tbergq:
I think this exposes a weakness in the approach we have chosen. 

Maybe it would be better to have more different code in `apps/mobile` and `apps/web` and focus more on smaller reusable parts in `apps/core` 🤔

JoseDuSV:
This case definitely points out a weak point. The question is how big is the possibility that we will encounter same situation in the future.
We can definitely rethink our approach or at least adjust it a bit. I'm just not that sure if it's worth it just because of this component, where it is still quite solvable as I've suggested.

I think that for now we can use new structure as I sketched to have only one file, so we can move forward. And meanwhile we can start some discussion to be prepared for similar cases in the future.

FilipMessa:
This is an edge-case and now there is no clear way how to solve this.
But it is clear, solve this is not part of this PR. 
The Margarita shouldn't be the copy of the kiwi.com, we just take inspiration for design, because there are no designs specific to the margarita, so I like JosefDuda's suggested solution. 

But this edge-case, we should consider and think about that, because of this can happen again and also because one of our goals is to show off that this is a good way how to build an application. We can change the designs for the Margarita, but there will be applications which can't.

apps/core/components/ItineraryCard/TripSectorWeb.js
```diff
@@ -0,0 +1,132 @@
...
+          style={styles.date}
+        />
+      </View>
+      <View style={styles.marginEnd}>
+        <Text style={styles.duration} numberOfLines={1}>
+          {getDuration(data?.duration)}
+        </Text>
+        <View style={styles.destinations}>
+          <LocationName data={data?.origin} style={styles.destination} />
+          <Icon name="route-no-stops" color="#bac7d5" size="large" />
```
jvaclavik: this colour could be included like `defaultTokens. colorIconTertiary`

JoseDuSV:
Refactored. Components order is now same as in mobile version and thanks to that we can have everything in just one file without many unreadable conditions.

screenshot 2019-02-20 at 14 12 26

apps/core/components/ItineraryCard/TripSectorWeb.js
JoseDuSV: Thanks, updated.

apps/core/components/ItineraryCard/StopoverDuration.js
JoseDuSV: Definitely, updated.

JoseDuSV:
@FilipMessa Branch updated from master and ready to be merged 🚀

PR #316, opened 2019-02-17
Added GraphQL Voyager

michalsanger:


apps/voyager/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "@kiwicom/margarita-voyager",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {},
+  "devDependencies": {
+    "serve": "^10.1.2"
+  },
+  "scripts": {
+    "start": "yarn serve -s public -p 3002",
+    "deploy": "now public -n margarita-graphql-voyager"
```
tbergq: is this copy/pasted from somewhere? I don't think we have now deploy setup in this repo 🤔

RobinCsl: We don't have it set up, but Michal deployed it.
As long as we don't change the GraphQL endpoint https://github.com/kiwicom/margarita/pull/316/files#diff-d6baed1b9b5e115ca918c719129038cdR37, we should not need to deploy it again.
Do you think auto-deploy could be necessary at some point?

tbergq: I guess not, but will running `yarn deploy` work? if not, we should remove it.

RobinCsl: I don't think it will, so 👍 for removing it.

michalsanger: So far it's on my Now account. And the deploy is supposed to be run manually and pretty rare I'd say.

PR #322, opened 2019-02-18
Graphql: Add RouteStop type to Sector and Segment

tbergq:
This should make reuse on FE easier when time and location is on same object.
Next iterations: Use new structure on FE.
Replace Trip type on MMB graphql with Sector type.

### DIFF
--- a/apps/graphql/src/dataloaders/ItinerariesloaderTypes.js
+++ b/apps/graphql/src/dataloaders/ItinerariesloaderTypes.js
@@ -1,6 +1,7 @@
 // @flow
 
 import { type Location } from './LocationsloaderTypes';
+import type { RouteStop } from './bookingsLoader/BookingFlowTypes';
 
 export type ItinerariesSearchParameters = {|
   +travelFrom: string,
@@ -22,8 +23,8 @@ export type Price = {|
 |};
 
 export type Time = {|
-  +local: ?string,
-  +utc: ?string,
+  +local: ?(string | number),
+  +utc: ?(string | number),
 |};
 
 export type Transporter = {|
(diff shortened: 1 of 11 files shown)


michalsanger:
@FilipMessa told me `RouteStop` is deprecated. Does it fit into our dictionary?

FilipMessa:
@michalsanger I was talking about the `RouteType`, which was part of our old schema, it dosen't exist yet

tbergq:
I did not find anything in the kiwi glossary that directly should replace the naming RouteStop. 

It is just an object for having time and location of a segment or sector.

FilipMessa:
In my opinion, this change is OK with the kiwi dictionary, and structure. And It is improvements for our schema design.

michalsanger:
Few issues:
- the name should not be `RouteStop` if there is no `Route`. Maybe `SegmentStop`?
- `cityId` and `cityName` points there should be `city`, or maybe `location` to support other types of stops?
- why `airport` when there will be the `location` and wee support also trains a and buses?
- if there is `time`, should there be also duration?

tbergq:
I don't think `SegmentStop` is any better, since this exists on Segment and Sector

`cityId` might not be correct, because this is actually the IATA code of the Sector/Segment, `cityName` also comes from the tequila response as `cityTo` and we have no location code to fetch the location. 

`airport` might not be the best, since this also might be bus stop or train stop

Suggestion: replace `cityId` with IATA
And airport with some general name for buss/airport/train stop

FilipMessa:
- I suppose that **duration** is not what should be there, the time here represent arrival time or departure time, based on a parent. The duration should be part of sector and segment schema which currently is.

michalsanger:
Btw where is it used on FE? Could post screenshot here please?

tbergq:
simulator screen shot - iphone x - 2019-02-19 at 10 33 43
simulator screen shot - iphone x - 2019-02-19 at 10 33 10
simulator screen shot - iphone x - 2019-02-19 at 10 32 11

PR #324, opened 2019-02-18
Query passengers and contact details

KarinaDavtyan:
Summary: Querying passengers and contact details from `AllBookings.json`

Note: `visaRequired` field is **hardcoded** and set to `false` for now

### DIFF
--- a/apps/graphql/src/dataloaders/bookingsLoader/BookingsLoader.js
+++ b/apps/graphql/src/dataloaders/bookingsLoader/BookingsLoader.js
@@ -16,23 +16,25 @@ import type {
 const sanitizeBookings = (
   bookings: $ReadOnlyArray<BookingApiResult>,
 ): $ReadOnlyArray<Booking> => {
-  return bookings.map(booking => {
-    const type = detectType(booking);
-    const typeSpecificData = getTypeSpecificData(booking, type);
-    const passengers = sanitizePassengers(booking);
-    return {
-      bid: booking.bid,
-      status: booking.status,
-      passengerCount: booking.passengers.length,
-      contact: {
-        email: booking.contact.email,
-        phone: booking.contact.phone,
-      },
-      passengers,
-      type,
-      ...typeSpecificData,
-    };
-  });
+  return bookings.map(sanitizeBooking);
+};
+
+export const sanitizeBooking = (booking: BookingApiResult) => {
+  const type = detectType(booking);
+  const typeSpecificData = getTypeSpecificData(booking, type);
+  const passengers = sanitizePassengers(booking);
+  return {
+    bid: booking.bid,
+    status: booking.status,
+    passengerCount: booking.passengers.length,
+    contact: {
+      email: booking.contact.email,
+      phone: booking.contact.phone,
+    },
+    passengers,
+    type,
+    ...typeSpecificData,
+  };
 };
 
 const sanitizePassengers = (booking: BookingApiResult): Array<Passenger> =>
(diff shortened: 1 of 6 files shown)


apps/graphql/src/dataloaders/bookingsLoader/BookingsLoader.js
```diff
@@ -18,17 +19,38 @@ const sanitizeBookings = (
...
+      },
+      passengers,
       type,
       ...typeSpecificData,
     };
   });
 };
 
+const sanitizePassengers = (booking: BookingApiResult): Array<Passenger> =>
+  booking.passengers.map((passenger, index) => {
```
tbergq: Index is not used, can you please remove it?

We should update our eslint to catch these ones in the future.

apps/graphql/src/types/output/Passenger.js
```diff
@@ -0,0 +1,45 @@
...
+      type: GraphQLString,
+    },
+    category: {
+      type: GraphQLString,
+    },
+    firstname: {
+      type: GraphQLString,
+    },
+    id: {
+      type: GraphQLInt,
```
tbergq: Can you make this a global id instead?

apps/graphql/src/types/output/Passenger.js
```diff
@@ -0,0 +1,45 @@
...
+    },
+    insurance_type: {
```
tbergq: please use camelCase

apps/graphql/src/types/output/Passenger.js
tbergq: https://github.com/kiwicom/margarita/blob/master/apps/graphql/src/types/output/BookingInterface.js#L16

apps/graphql/src/types/output/Passenger.js
KarinaDavtyan: Can I add camelcase eslint rule, as snake case keep happening again and again
If yes, can I put `ItinerariesMock.js` in `.eslintignore` or I could just disable this rule for this file

tbergq: > Can I add camelcase eslint rule

Please do not do it for now, as we get most the rules from `@kiwicom/eslint-config` I will have a look at it and see if we can improve it for our needs.

KarinaDavtyan: Ok sure, wont touch it! I ll just fix it to `insuranceType` then

apps/graphql/src/dataloaders/bookingsLoader/BookingsLoader.js
KarinaDavtyan: yes, missed that. Is removed now

apps/graphql/src/types/output/Passenger.js
KarinaDavtyan: Ok, changed to global id now

apps/graphql/src/types/output/Passenger.js
```diff
@@ -0,0 +1,44 @@
...
+  GraphQLInt,
+  GraphQLBoolean,
+} from 'graphql';
+import globalID from '@kiwicom/graphql-global-id';
+
+export default new GraphQLObjectType({
+  name: 'Passenger',
+  description: 'Passenger details',
+  fields: {
+    id: globalID(({ bid }) => bid),
```
tbergq: `id: globalID(({ id }) => id)`?

KarinaDavtyan: 😓 changing now

PR #330, opened 2019-02-18
Feat: Adding documentation for Margarita

RobinCsl:
Summary: These docs were kickstarted using `npm init docs` and use NextJs. It's a pretty basic setup and we wouldneed to create more components and custom logic to have a nicer feel to them. I also believe the styling could be improved (https://github.com/jxnblk/mdx-docs/blob/master/docs/pages/theming.md).
Let's start documenting!

Note: I am not sure how to keep in sync our README.md at the root of the repo; we could potentially add a route in `_app.js` that somehow points to a file higher up in the folder structure and remove the `index.md` file in `apps/documentation`.

### DIFF
--- a/apps/documentation/package.json
+++ b/apps/documentation/package.json
@@ -1,0 +1,27 @@
+{
+  "name": "@kiwicom/margarita-docs",
+  "private": true,
+  "version": "0.0.0",
+  "scripts": {
+    "start": "next -p 1337",
+    "build": "next build && next export"
+  },
+  "dependencies": {
+    "@mdx-js/mdx": "^0.15.0",
+    "mdx-docs": "^2.0.0-alpha.0",
+    "remark-autolink-headings": "^5.1.0",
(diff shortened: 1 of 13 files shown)


apps/documentation/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "mdx-docs": "^2.0.0-alpha.0",
+    "next": "^7.0.1",
+    "next-images": "^1.0.4",
+    "react": "^16.4.1",
+    "react-dom": "^16.4.1",
+    "remark-autolink-headings": "^5.1.0",
+    "remark-emoji": "^2.0.2",
+    "remark-images": "^0.16.1",
+    "remark-slug": "^5.1.1",
+    "remark-unwrap-images": "^0.1.0"
```
tbergq: all dependencies here that are defined in other package.json should be moved to `peerDependencies`

Should be `next`, `react`, `react-dom` atleast

PR #331, opened 2019-02-18
Eslint: set no-unused-vars rule to throw error for unused args

tbergq:


packages/components/src/tripTypeButton/__tests__/TripTypeButton.test.js
```diff
@@ -16,7 +16,7 @@ const { getByType } = render(
 
 it('should execute onPress method', () => {
   fireEvent(getByType(TouchableWithoutFeedback), 'press');
-  expect(onPress).toHaveBeenCalled();
+  expect(onPress).toHaveBeenCalledWith(undefined);
```
FilipMessa: *Just for my information*
Why do you change `toHaveBeenCalled()` into `toHaveBeenCalledWith(undefined)`?

tbergq: because we have warnings from eslint -> `jest/prefer-called-with`

FilipMessa: thanks

PR #337, opened 2019-02-19
Refactor: Modal implementation

### LINKED ISSUE #329: Refactor: Modal implementation
Replace current `Modal` implementation with solution based on new flexible modal version from `universal-components`.
This update will allow modal to be placed anywhere in components structure, not just in root like current implementation.

JoseDuSV:
Summary: Replaced `Modal` implementation with solution based on `universal-components` version.

Removed current limitation when modal needs to be placed in structure root.

In next iteration we can dissolve `SearchModal` component and place modal implementation directly to each logical block (TripType, Passengers, PlacePickers).

Closes #329


packages/components/src/modal/Modal.js
```diff
@@ -1,34 +1,54 @@
...
+  +children: React.Node,
+  +isVisible: boolean,
+  +onClose?: () => void,
+|};
 
-const Modal = (props: ModalProps) => {
+export default function Modal({ children, isVisible, onClose }: Props) {
+  const { width, height } = Dimensions.get('window');
+  // Center modal content on web and tablets
+  const centerContent = Platform.OS === 'web' || height / width < 1.6;
```
tbergq: Should we maybe put this logic into some module, so we can say `Device.isTablet()`?

packages/components/src/modal/Modal.js
```diff
@@ -1,34 +1,54 @@
...
-      {props.visible && <View style={styles.overlay} />}
-      <ModalWrap {...props} />
-    </>
+    <UCModal
+      isVisible={isVisible}
+      onRequestClose={onClose}
+      onBackdropPress={onClose}
+      style={[styles.modal, centerContent && styles.modalCentered]}
+    >
+      <View style={[styles.content, centerContent && styles.contentCentered]}>
```
RobinCsl: This might be problematic if we would like the content to be scrollable. I think it would be worth it to experiment with a ScrollView instead.

tbergq: True, and you could combine it with `alwaysBounceVertical=false` so that it does not scroll unless it should

packages/components/src/modal/Modal.js
RobinCsl: I agree.

packages/components/src/modal/Modal.js
JoseDuSV: @RobinCsl I've played with your idea for a bit. But at the end I don't think we currently have any use case where scrollable modal will be necessary.

Maybe you've got any specific use-case in mind, just let me know.

The only scrollable content that I've found is place picker and itinerary detail. Where scrolling is solved inside component itself. 

Because place picker have static input on top. With scrollable list underneath.
screenshot 2019-02-19 at 14 58 23

And there is just scrollable list with static button on top in itinerary detail. But whole modal content doesn't need to be scrollable.
screenshot_20190219-145731

This extended modal simply just provides content wrapper with shared style and logic (centered on web/tablet, at the bottom on mobile). 
But while considering all possible usages I've added `wrapperStyle` prop which enables extending wrapper style. So for example wrapper height can be set to fixed value or better to percents, like currently used `70%` for place picker.

packages/components/src/modal/Modal.js
JoseDuSV: Good idea. Extracted to `Device` inside `margarita-utils`.

I'm also considering something similar for quite frequently repeating `Platform.OS === 'web'`.
What about our own `Platform` implementation with something like `Platform.isWeb()` (isMobile, isAndroid, isIos)?

tbergq: Sure, why not

JoseDuSV: 👍 ok, I'll create separate PR for it.

PR #338, opened 2019-02-19
Graphql: Remove unused fields from Segment type

tbergq:


apps/graphql/src/dataloaders/ItinerariesloaderTypes.js
```diff
@@ -37,12 +37,8 @@ export type Vehicle = {|
 |};
 
 export type Segment = {|
-  +arrivalTime: ?Time,
```
jvaclavik: Why do you delete those fields? We probably do not need it now but we will need it in trip detail, or no?

tbergq: It is under RouteStop.time

FilipMessa: You can find this data in the `RouteStop`, but now there are some issues with naming which should be improved. 
https://github.com/kiwicom/margarita/pull/322
- feel free to write down your opinion about that change thanks

PR #339, opened 2019-02-19
Graphql: Remove cityId from RouteStop type

tbergq:
Rename airport to stop, since it could be any kind of stop

see discussion in #322

### DIFF
--- a/apps/graphql/src/dataloaders/itinerariesHelpers.js
+++ b/apps/graphql/src/dataloaders/itinerariesHelpers.js
@@ -133,7 +133,7 @@ const sanitizeSector = (
 
 const apiRouteItemToArrival = (routeItem: ?ApiRouteItem) => ({
   cityName: routeItem?.cityTo,
-  cityId: routeItem?.flyTo,
+  code: routeItem?.flyTo,
   time: {
     utc: routeItem?.utc_arrival,
     local: routeItem?.local_arrival,
@@ -142,7 +142,7 @@ const apiRouteItemToArrival = (routeItem: ?ApiRouteItem) => ({
 
 const apiRouteItemToDeparture = (routeItem: ?ApiRouteItem) => ({
   cityName: routeItem?.cityFrom,
-  cityId: routeItem?.flyFrom,
+  code: routeItem?.flyFrom,
   time: {
     utc: routeItem?.utc_departure,
     local: routeItem?.local_departure,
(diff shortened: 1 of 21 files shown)


FilipMessa:
@michalsanger can you please look on this PR also? thanks

apps/graphql/src/dataloaders/itinerariesHelpers.js
```diff
@@ -151,17 +151,17 @@ const apiRouteItemToDeparture = (routeItem: ?ApiRouteItem) => ({
...
   time,
 }: {|
   +cityName: ?string,
-  +cityId: ?string,
+  +code: ?string,
   +time: {| +utc: ?string, +local: ?string |},
 |}): RouteStop => ({
   cityName,
-  cityId,
+  cityId: null,
```
michalsanger: Why is `cityId` not removed from the `RouteStop`?

apps/graphql/src/types/output/RouteStop.js
```diff
@@ -9,18 +9,19 @@ import type { GraphqlContextType } from '../../services/GraphQLContext';
 
 export default new GraphQLObjectType({
   name: 'RouteStop',
-  description: 'Departure or arrival for a segment',
+  description: 'Departure or arrival for a segment or sector',
   fields: {
     cityName: {
```
michalsanger: I suggest to remove also `cityName`. The `Location` entity should have new property `city: LocationArea` from where name of the city can be resolved as `stop?.city?.name`

apps/graphql/src/dataloaders/itinerariesHelpers.js
tbergq: Because tequila does not return the cityId, it returns the iata code of the stop.

tbergq: We have cityId from the mocked data from MMB, though we are currently not using it, so I guess I can remove it, and add it later if we need it

apps/graphql/src/types/output/RouteStop.js
tbergq: Yes, that should probably be doable. I see the location query returns a field called city, we should be bale to get it from there.

apps/graphql/src/dataloaders/itinerariesHelpers.js
tbergq: Sorry, scratch that, I use it to get `destinationImageUrl` in `BookingInterface`. So cityId has to stay in the resolvers.

tbergq:
@michalsanger 🚀

PR #341, opened 2019-02-19
Replace Trip type with Sector type

tbergq:
Sector is the correct name for this type according to kiwicom glossary


tbergq:
> rename the React components to be tripsector

Will do it next PR

Issue #276, opened 2019-02-13
Shouldn't be possible to search without locations


jvaclavik:
This input:
![screenshot 2019-02-13 at 12 05 04](https://user-images.githubusercontent.com/858321/52707280-9d708c00-2f87-11e9-99e6-c7190508738f.png)

will produce this:
![screenshot 2019-02-13 at 12 05 36](https://user-images.githubusercontent.com/858321/52707307-b0835c00-2f87-11e9-907d-7e5b50e49842.png)

JoseDuSV:
We talked about it with @FilipMessa and for now we will at least add some default values like `Praha` (or Brno?) -> `Anywhere`. So the case when inputs are empty shouldn't happen.

RobinCsl:
Some form validation could be useful, but since we are going to use the placepicker eventually, I guess it's not so necessary?

tbergq:
In current app, you get an alert if you try to search without inputing anything.

tbergq:
![img_4946](https://user-images.githubusercontent.com/7705041/52708528-aca50900-2f8a-11e9-9962-a322506a4976.jpg)

jvaclavik:
@tbergq And does it work for iOS simulator and web? Because I haven't seen this alert.

RobinCsl:
@tbergq We could do something like that, yes.
@jvaclavik That's because that's the Kiwi.com app, not Margarita 😉

PR #353, opened 2019-02-20
Feat: Display warning if no locations

RobinCsl:
Summary: The current behaviour is to display the message only if both travelFrom and travelTo are null.
If you want to reproduce it you can replace in `SearchContext.js`
```
const defaultState = {
  tripType: 'return',
  travelFrom: null, // defaultPlaces.origin,
  travelTo: null, // defaultPlaces.departure,
...
```

**EDIT**: This PR adds a new Alert component (together with a context to use it) in @kiwicom/margarita-components. It displays a Modal UI element on web and uses react-native's Alert on mobile.

Closes https://github.com/kiwicom/margarita/issues/276

### DIFF
--- a/apps/core/scenes/search/Search.js
+++ b/apps/core/scenes/search/Search.js
@@ -3,7 +3,12 @@
 import * as React from 'react';
 import { View, Platform } from 'react-native';
 import { StyleSheet, Button, Icon } from '@kiwicom/universal-components';
-import { Illustration } from '@kiwicom/margarita-components';
+import {
+  Illustration,
+  withAlertContext,
+  type AlertContextState,
+  type AlertContent,
+} from '@kiwicom/margarita-components';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import {
   withNavigation,
@@ -41,6 +46,7 @@ type Props = {
   +infants: number,
   +bags: number,
   +layout: number,
+  +setAlertContent: (alertContent: AlertContent | null) => void,
 };
 
 class Search extends React.Component<Props> {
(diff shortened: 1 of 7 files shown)


tbergq:
Why not just use an alert for this?

RobinCsl:
> Why not just use an alert for this?

You mean on mobile? Because on web I believe we would want to use a modal and not the browser alerts.

tbergq:
Maybe we could wrap it, so on mobile, we use native alert, but on web, we show a modal?

RobinCsl:
Sounds good to me, and it will definitely look much nicer on mobile.

RobinCsl:
@tbergq 🚀 

(Just weird Flow error that `Alert` cannot be imported from `react-native`, but I suspect this is because `react-native-web` does not export it...)

tbergq:
I was hoping for something a little bit more reusable. Next time we want an alert, we have to repeat the modal and everything. 

Just thinking, how about adding a modal in component in `Layout.js`, then add an Alert context, when there is pushed a message to this context, we show the alert modal for web

RobinCsl:
Sure, that's a good idea. Back to WIP 😊

apps/core/scenes/search/Search.js
```diff
@@ -58,20 +65,24 @@ class Search extends React.Component<Props> {
...
-      dateFrom: format(dateFrom, DATE_FORMAT),
-      dateTo: format(dateTo, DATE_FORMAT),
-      ...(tripType === 'return'
-        ? {
-            returnDateFrom: format(returnDateFrom, DATE_FORMAT),
-            returnDateTo: format(returnDateTo, DATE_FORMAT),
-          }
-        : {}),
-    });
+    if (travelFrom == null && travelTo == null) {
```
jvaclavik: What if there is no `travelFrom `?

apps/core/scenes/search/Search.js
```diff
@@ -101,6 +112,7 @@ class Search extends React.Component<Props> {
           </View>
         </View>
         <SearchModal />
+        <Alert message="Please fill this form completely before you proceed" />
```
JoseDuSV: I'm still not sure about this implementation. What if there will be multiple `Alert` components on one screen (or in nested components)?
Context doesn't seems to add much in this case.
I liked @tbergq original idea to have just one `Alert` in some higher component. And be able to just call something like `showAlert('message')` on alert context to display the message.

apps/core/scenes/search/Search.js
```diff
@@ -41,6 +46,8 @@ type Props = {
   +infants: number,
   +bags: number,
   +layout: number,
+  +isAlertVisible: boolean,
```
JoseDuSV: This prop doesn't seem to be used anywhere.

apps/core/scenes/search/Search.js
RobinCsl: Good point. On Kiwi FE, the button is actually disabled in that case. So I don't know if alerting was the best UX solution.

apps/core/scenes/search/Search.js
RobinCsl: Sure, I will change it.

JoseDuSV: Thanks. For web we can probably add it directly into `_app.js` or `Layout`. I'm just not that sure about mobile, probably `App.js` can also work.

RobinCsl:
Hopefully ready now 😊 @tbergq @jvaclavik @JoseDuSV

PR #355, opened 2019-02-20
MMB: Add expand button with icon under sector info

tbergq:
Add animation for icon on click

### DIFF
--- a/apps/core/scenes/bookingDetail/sectorDetails/SectorDetails.js
+++ b/apps/core/scenes/bookingDetail/sectorDetails/SectorDetails.js
@@ -4,11 +4,13 @@ import * as React from 'react';
 import { Card, StyleSheet } from '@kiwicom/universal-components';
 import { Separator, BookingTypeRenderer } from '@kiwicom/margarita-components';
 import { createFragmentContainer, graphql } from '@kiwicom/margarita-relay';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 
 import Header from './Header';
 import SectorInfoOneWay from './SectorInfoOneWay';
 import SectorInfoMulticity from './SectorInfoMulticity';
 import SectorInfoReturn from './SectorInfoReturn';
+import SegmentContainer from './segments/SegmentContainer';
 import type { SectorDetails as BookingType } from './__generated__/SectorDetails.graphql';
 
 type Props = {|
@@ -27,6 +29,8 @@ function SectorDetails(props: Props) {
         returnComponent={<SectorInfoReturn data={props.data} />}
         multicityComponent={<SectorInfoMulticity data={props.data} />}
       />
+      <Separator style={styles.bottomSeparator} />
+      <SegmentContainer />
     </Card>
   );
 }
@@ -34,10 +38,17 @@ function SectorDetails(props: Props) {
 const styles = StyleSheet.create({
   card: {
     paddingHorizontal: 18,
+    borderBottomWidth: StyleSheet.hairlineWidth,
+    borderBottomColor: defaultTokens.paletteInkLighter,
+    paddingBottom: 0,
   },
   separator: {
     marginTop: 7,
     marginBottom: 17.5,
+  },
+  bottomSeparator: {
+    marginHorizontal: -18,
+    marginTop: 16,
   },
 });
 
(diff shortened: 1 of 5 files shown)


apps/core/scenes/bookingDetail/sectorDetails/SectorDetails.js
```diff
@@ -27,18 +29,26 @@ function SectorDetails(props: Props) {
...
     paddingHorizontal: 18,
+    borderBottomWidth: StyleSheet.hairlineWidth,
+    borderBottomColor: defaultTokens.paletteInkLighter,
   },
   separator: {
     marginTop: 7,
     marginBottom: 17.5,
   },
+  bottomSeparator: {
+    marginHorizontal: -18,
```
jvaclavik: Why do you need this? Could you please add some screenshot how it looks like? It could be more clear for me.

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentContainer.js
```diff
@@ -0,0 +1,78 @@
...
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import { Animated } from 'react-native';
+
+type Props = {||};
+
+type State = {|
+  +expandSegments: boolean,
+|};
+
+export default class SegmentContainer extends React.Component<Props, State> {
```
jvaclavik: Do we have to create `Props` type? It should be possible to add there just `null` instead of `Props`

apps/core/scenes/bookingDetail/sectorDetails/SectorInfoMulticity.js
```diff
@@ -17,7 +17,7 @@ function SectorInfoMulticity(props: Props) {
   return trips.map((trip, index) => (
     <React.Fragment key={index}>
       <SectorInfo data={trip} />
-      <Separator style={styles.separator} />
+      {index !== trips.length - 1 && <Separator style={styles.separator} />}
```
jvaclavik: I'm not sure if it's useful here, but if you would change it to `FlatList` you could use `ItemSeparatorComponent` here: https://facebook.github.io/react-native/docs/flatlist#itemseparatorcomponent

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentContainer.js
```diff
@@ -0,0 +1,78 @@
...
+      expandSegments: false,
+    };
+
+    this.spinValue = new Animated.Value(0);
+  }
+
+  spinValue: Animated.Value;
+
+  toggleExpanded = () => {
+    Animated.timing(this.spinValue, {
```
JoseDuSV: When I tap on arrow quickly multiple times it can end up in wrong position.
Maybe we can disable button/toggle/state updates during animation.

screenshot 2019-02-20 at 15 25 35

Otherwise I love it.

apps/core/scenes/bookingDetail/sectorDetails/SectorDetails.js
tbergq: Because the card component has a paddingHorizontal of 18, and the seperator is from screen start -> screen end

tbergq: [Figma](https://www.figma.com/file/BNs81bMp3VVuxqKIICbnNvqd/MMB?node-id=1%3A193)

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentContainer.js
tbergq: Nope, Props needs to be an object, 
screenshot 2019-02-20 at 15 33 36

apps/core/scenes/bookingDetail/sectorDetails/SectorInfoMulticity.js
tbergq: Having a FlatList under a ScrollView has performance issues

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentContainer.js
tbergq: Yeah, I will have a look at that 👍

jvaclavik:
(left iOS simulator, right figma design)
screenshot 2019-02-20 at 15 36 41



Arrow is not is not vertically centered on iOS

tbergq:
> Arrow is not is not vertically centered on iOS

Yes, I forgot that, thanks

tbergq:
@jvaclavik @JoseDuSV 🚀

apps/core/scenes/bookingDetail/sectorDetails/SectorInfoMulticity.js
jvaclavik: I see, so if the component is used somewhere with `ScrollView` parent, it make sense.

PR #363, opened 2019-02-21
MMB: Add map view to segment details

tbergq:


tbergq:
hmm, this might not work on web 🤔

tbergq:
Fixed so it will not crash on web.

PR #364, opened 2019-02-21
Graphql: Add gps coordinates to locations

tbergq:


apps/graphql/src/apps/location/dataloaders/Locations.js
```diff
@@ -24,6 +24,10 @@ function sanitizeLocations(locations: $ReadOnlyArray<ApiLocation>) {
     name: location.name,
     slug: location.slug,
     timezone: location.timezone,
+    coordinates: {
+      lat: location.location.lat,
```
jvaclavik: `location.location` is weird, maybe would be better to rename the first one (to `place` maybe), but it's not important.

tbergq: Not sure it makes sense, we are working with locations, not sure why tequila have chosen to call the gps-coordinate for location 🤷‍♂️

PR #367, opened 2019-02-21
Fix: TimelineFlightDetail accordion behaviour

JoseDuSV:
Summary: Detail is not expanded in default mode now.
Changed name of `CardHeader` to correct one.

I was also considering changing name to something more universal like `TimelineTransportDetail`. What do you think?


RobinCsl:
> `TimelineTransportDetail`

Yes, sounds good to me!

JoseDuSV:
@RobinCsl Thanks. Renamed and I also updated one value in styles to have line on left aligned with `TimelineInformation` icon according to current designs.

PR #368, opened 2019-02-21
Feat: Add type into the GraphQL Location schema

FilipMessa:
Added type field into the Location schema.

### DIFF
--- a/apps/graphql/src/apps/location/dataloaders/Locations.js
+++ b/apps/graphql/src/apps/location/dataloaders/Locations.js
@@ -17,12 +17,27 @@ export type LocationInput =
       +code: string,
     |};
 
+function sanitizeCity(location: ApiLocation) {
+  if (location.city != null) {
+    return {
+      id: location.city.id,
+      locationId: location.city.id,
+      code: location.city.code,
+      flagURL: null,
+      slug: location.city.slug,
+      name: location.city.name,
+    };
+  }
+  return null;
+}
+
 function sanitizeLocations(locations: $ReadOnlyArray<ApiLocation>) {
   return locations.map(location => ({
     id: location.id,
     locationId: location.id,
     name: location.name,
     slug: location.slug,
+    type: location.type,
     timezone: location.timezone,
     coordinates: {
       lat: location.location.lat,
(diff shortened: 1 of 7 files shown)


apps/graphql/src/apps/location/types/outputs/Location.js
```diff
@@ -32,6 +32,9 @@ export default new GraphQLObjectType({
     coordinates: {
       type: GraphQLCoordinate,
     },
+    type: {
+      type: GraphQLString,
```
tbergq: Maybe it should be an enum?

FilipMessa: I was thinking about that, but I can't find any documentation about all possible values from TequilaAPI. But I asked for that so we will see.

apps/graphql/src/apps/location/types/enums/LocationType.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+
+import { GraphQLEnumType } from 'graphql';
+
+export default new GraphQLEnumType({
+  name: 'LocationType',
+  values: {
+    station: {
```
tbergq: Values should be UPPER_CASED

apps/graphql/src/apps/location/dataloaders/Locations.js
```diff
@@ -17,12 +17,34 @@ export type LocationInput =
       +code: string,
     |};
 
+function sanitizeLocationType({ type }: ApiLocation) {
+  if (type) {
+    return type.replace(/_(\w)/g, m => m[1].toUpperCase());
```
tbergq: could you add a comment as to why this is necessary?

apps/graphql/src/apps/location/types/enums/LocationType.js
FilipMessa: So this it is the convention for Enums ?

tbergq: seems like it yeah, all enums in `incubator/universe` follow this convention, and all enums we have in margarita so far.

apps/graphql/src/apps/location/dataloaders/Locations.js
FilipMessa: done

tbergq: Ok, now that I know why we does this, it seems unnecessary to me. 

```
BUS_STATION: {
      value: 'bus_station',
},
```

will resolve this to the enum anyway.

FilipMessa: 🚀

Issue #366, opened 2019-02-21
Create empty ResultDetail screen


jvaclavik:
https://www.figma.com/file/OUhBJ9rPaTbRljptC0lrRrUF/Multistep?node-id=8%3A6

JoseDuSV:
I think we already have the issues for this one. Top part #72 will be mostly reused from MMB.
And #73 for passenger related part, this one probably just needs to be slightly updated according to current designs.

jvaclavik:
> I think we already have the issues for this one. Top part #72 will be mostly reused from MMB.
> And #73 for passenger related part, this one probably just needs to be slightly updated according to current designs.

Yes, but this issue should be mainly for creating empty screen and connect it with the ResultsScreen (I need it for another PR).

PR #369, opened 2019-02-21
Create empty ResultDetailScreen

jvaclavik:
Creating empty screen for ResultDetail (#366)

### DIFF
--- a/apps/core/components/ItineraryCard/itineraryDetail/ItineraryDetail.js
+++ b/apps/core/components/ItineraryCard/itineraryDetail/ItineraryDetail.js
@@ -10,11 +10,13 @@ import ItineraryDetailWrapper from './ItineraryDetailWrapper';
 type Props = {|
   +localizedPrice: string,
   +onClose: () => void,
+  +onBookPress: string => void,
 |};
 
 export default class ItineraryDetail extends React.Component<Props> {
   handleBookPress = () => {
-    // @TODO redirect to booking screen
+    this.props.onBookPress('lol'); // @TODO use real ID
+    this.props.onClose();
   };
 
   render() {
--- a/apps/core/scenes/resultDetail/ResultDetail.js
+++ b/apps/core/scenes/resultDetail/ResultDetail.js
@@ -1,0 +1,28 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { StyleSheet, Text } from '@kiwicom/universal-components';
+import { withNavigation } from '@kiwicom/margarita-navigation';
+
+type Props = {|
+  +detailId: ?string,
+|};
+
+class ResultDetail extends React.Component<Props> {
+  render() {
+    return (
+      <View style={styles.container}>
+        <Text>{this.props.detailId}</Text>
+      </View>
+    );
+  }
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+  },
+});
+
+export default withNavigation(ResultDetail);
--- a/apps/mobile/src/screens/ResultDetailScreen.js
+++ b/apps/mobile/src/screens/ResultDetailScreen.js
@@ -1,0 +1,20 @@
+// @flow
+
+import * as React from 'react';
+import { ResultDetail } from '@kiwicom/margarita-core';
+import { type NavigationScreenProp } from 'react-navigation';
+
+type NavigationParams = {|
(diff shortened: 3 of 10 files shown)


apps/core/components/ItineraryCard/itineraryDetail/ItineraryDetail.js
```diff
@@ -10,11 +10,13 @@ import ItineraryDetailWrapper from './ItineraryDetailWrapper';
 type Props = {|
   +localizedPrice: string,
   +onClose: () => void,
+  +onBookPress: string => void,
 |};
 
 export default class ItineraryDetail extends React.Component<Props> {
   handleBookPress = () => {
-    // @TODO redirect to booking screen
+    this.props.onBookPress('lol'); // @TODO use real ID
```
tbergq: 😄

apps/mobile/src/screens/ResultDetailScreen.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import * as React from 'react';
+import { ResultDetail } from '@kiwicom/margarita-core';
+
+type Props = {|
+  +navigation: {|
```
tbergq: Can you use [this](https://github.com/kiwicom/margarita/blob/master/apps/mobile/src/screens/BookingDetailScreen.js#L15) type instead?

apps/core/components/ItineraryCard/itineraryDetail/ItineraryDetail.js
jvaclavik: Easter egg for you 😄

apps/core/scenes/resultDetail/ResultDetail.js
```diff
@@ -0,0 +1,28 @@
...
+import * as React from 'react';
+import { View } from 'react-native';
+import { StyleSheet, Text } from '@kiwicom/universal-components';
+import { withNavigation } from '@kiwicom/margarita-navigation';
+
+type Props = {|
+  +id: ?string,
+|};
+
+class ResultDetail extends React.Component<Props> {
```
JoseDuSV: Shouldn't this screen be called `Booking` or `Booking"Something"` (indicating first step of multistep booking process)? There is just result overview on top of this screen and rest is the passengers info.

jvaclavik:
> Just switch the type as commented, looks good apart from that

Done. I also renamed `id` to `detailId` to keep the same convention as in `BookingDetail`

apps/core/scenes/resultDetail/ResultDetail.js
jvaclavik: That's something what I was thinking about and I'm not sure about it. 
I kept this name because it would be too similar with screens which we already have in the app (BookingDetailScreen, BookingsListScreen).

It's not perfect, because as you said it would be fine to know from the name that you are in the first step of bigger process.

tbergq: Maybe we should add some folders here as well,

MMB 			- 		Search
BookingDetail   -		BookingResults

WDYT?

jvaclavik: It would be still confusing when you have it in your editor or you find for files. I would prefer keep it flat. 

What about rename current bookings:
```
BookingDetailScreen -> MMBDetail
BookingsList -> MMBList
```

And this new screen:
```
ResultDetail -> BookingCreate  (or something like that)
```

JoseDuSV: @jvaclavik I totally understand your concern about mixing naming with MMB.

I think that adding folders can work. It will be nice to have next booking steps (payment, summary...) under some namespace. But it can be definitely part of different PR.

jvaclavik: Even now it's hard to understand it if you are looking for some file:

screenshot 2019-02-21 at 16 35 00

JoseDuSV: It doesn't look that confusing to me (...yet 😄). But I totally get your point.
I'm ok with both options (folders, MMBSomething). I think we can merge this PR as it is to move forward. And create new issue and start discussion there or on slack about better structure/naming for this and next booking multistep screens.

tbergq:
LGTM, @JoseDuSV, please also review again.

PR #378, opened 2019-02-22
Graphql: Add gps location of all segments to BookingInterface

tbergq:
We will need these coordinates to set the MMB map to correct region


apps/graphql/src/apps/booking/types/outputs/BookingInterface.js
```diff
@@ -66,8 +69,42 @@ export const commonFields = {
...
+          })),
+        ];
+      }, []);
+    },
+  },
 };
 export default new GraphQLInterfaceType({
   name: 'BookingInterface',
-  fields: commonFields,
+  fields: {
```
RobinCsl: Is that change necessary?

tbergq: Absolutely not, I will undo it.

tbergq: 🚀

apps/graphql/src/apps/booking/types/outputs/BookingInterface.js
```diff
@@ -66,6 +69,38 @@ export const commonFields = {
     type: GraphQLBookingType,
     description: 'OneWay, Multicity or Return',
   },
+  segmentLocations: {
+    type: GraphQLList(GraphQLCoordinate),
+    description:
+      'This will contain the gps-coordinates for all stops on the booking',
```
RobinCsl: In the same way we have `deprecated` flags, would it be possible to have some kind of `future feature` flag? In my opinion, a description should use present tense, but I understand how it is necessary here if there is no such `feature` flag.

tbergq: Yeah, you are right, I will change it to `This contains ...`

tbergq: 🚀

PR #380, opened 2019-02-22
Feat: Starting documentation on GraphQL server

RobinCsl:


RobinCsl:
I also removed --frozen-lockfile from CircleCI config. I don't know if that will solve the issues regarding `installing-dependencies` failing pretty much all the time...

apps/documentation/pages/apps/graphql/index.md
```diff
@@ -0,0 +1,107 @@
...
+│    ├── mutations              # GraphQL mutations concerning feature
+│    ├── resolvers              # more complex GraphQL resolvers
+│    └── types                  # GraphQL types concerning feature
+│        ├── enums
+│        ├── inputs             # definitions of INPUT types exposed to the client, e.g. arguments in a query
+│        └── outputs            # definitions of OUTPUT types exposed to the client, e.g. GraphQL type for GPS coordinates
+
+```
+
+The `dataloaders` folder is where the logic to connect to external APIs is stored. It has this name because it relies under the hood on https://github.com/facebook/dataloader which allows batching (i.e. trigger one API request even if your code needs to call the same API endpoint with different queries) and caching (i.e. using previously obtained data and not calling the API endpoint if the same call has already been made before).
```
tbergq: Maybe we should mention that it is pr-request cache?

tbergq:
Looks good 😄

apps/documentation/pages/apps/graphql/index.md
RobinCsl: Sure, I will add that with link to corresponding line of code.

RobinCsl: 🚀

PR #381, opened 2019-02-22
MMB: Add markers to sector map indicating all stop for the booking

tbergq:


apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
```diff
@@ -3,25 +3,96 @@
 import * as React from 'react';
 import { MapView } from '@kiwicom/margarita-map';
 import { StyleSheet } from '@kiwicom/universal-components';
+import { graphql, createFragmentContainer } from '@kiwicom/margarita-relay';
+import { getCenter } from 'geolib';
+import memoize from 'memoize-one';
+import stringify from 'json-stable-stringify';
```
RobinCsl: I think I read about this somewhere on Slack, right? This is to make sure the stringified version is always ordered in the same way, right?

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
```diff
@@ -3,25 +3,96 @@
 import * as React from 'react';
 import { MapView } from '@kiwicom/margarita-map';
 import { StyleSheet } from '@kiwicom/universal-components';
+import { graphql, createFragmentContainer } from '@kiwicom/margarita-relay';
+import { getCenter } from 'geolib';
```
RobinCsl: Is geolib also used by FE?

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
```diff
@@ -3,25 +3,96 @@
...
+    this.map = ref;
+  };
+
+  getMappedCoordinates = () => {
+    return mapCoordinates(this.props.data?.segmentLocations);
+  };
+
+  fit = () => {
+    if (this.map != null) {
+      // $FlowExpectedError: fitToCoordinates does exist on map, but flow is confused by .native|.web exports
```
RobinCsl: I have a suspicion that `react-native-web` flow-typed definitions are the reason behind many of these false positives.

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
tbergq: yes [docs here](https://github.com/substack/json-stable-stringify#json-stable-stringify)

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
tbergq: 🤷‍♂️ it is used in kiwicom/mobile

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
tbergq: I suspect [this one](https://github.com/kiwicom/margarita/blob/master/packages/map/src/MapView.web.js)

PR #390, opened 2019-02-25
Fix: Wrong sector hours duration

JoseDuSV:
Summary: Updated formatting to correctly handle duration longer than one day.

Before
screenshot 2019-02-25 at 10 32 07

After
screenshot 2019-02-25 at 10 31 43

Note: In next iteration we should move these helpers into `@kiwicom/margarita-utils` (it will be used by multiple components) and write proper tests for it.


tbergq:
Would it be possible to reuse `Duration` component from margarita-components for this?

JoseDuSV:
@tbergq Sure it can. Replaced. Thanks a lot

PR #392, opened 2019-02-25
Web: Move MMBWrapper out of pages folder

tbergq:
It will create a route if it is located there.
Use media queries for mmb layout


apps/web/components/Wrapper.js
```diff
@@ -0,0 +1,34 @@
...
+
+import Layout from './Layout';
+
+type Props = {|
+  +children: React.Node,
+|};
+
+const ChildrenWrapper = styled.div({
+  marginTop: defaultTokens.spaceXSmall,
+  marginLeft: '25%',
```
RobinCsl: Could we have mobile-first approach so that styles are faster to be parsed on mobile? I mean by that that CSS should apply first for mobile, then media query for tablet and finally media query for desktop.

tbergq: Sure

tbergq: 🚀

PR #393, opened 2019-02-25
Fix: Scrollable web MMBWrapper

JoseDuSV:
Summary: While adding itinerary detail to `booking-detail` I've noticed that wrapper on web is not scrollable. Updated styles to make it scroll.


tbergq:
I prefer we use #394, WDYT?

JoseDuSV:
> I prefer we use #394, WDYT?

@tbergq Sure, it will be more consistent with other pages. So I'm closing this one.

PR #394, opened 2019-02-25
MMB: Use View instead of div in mmb-Wrapper

tbergq:
Using div instead of View broke the functionality of nested scroll view.
It no longer scrolled

### DIFF
--- a/apps/web/components/Wrapper.js
+++ b/apps/web/components/Wrapper.js
@@ -4,6 +4,7 @@ import * as React from 'react';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import styled from 'styled-components';
 import { LAYOUT } from '@kiwicom/margarita-utils';
+import { ScrollView } from 'react-native';
 
 import Layout from './Layout';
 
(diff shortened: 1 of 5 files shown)


apps/web/components/Wrapper.js
```diff
@@ -4,14 +4,16 @@ import * as React from 'react';
...
 import Layout from './Layout';
 
 type Props = {|
   +children: React.Node,
 |};
 
-const ChildrenWrapper = styled.div({
+const ChildrenWrapper = styled(View)({
+  flex: 1,
   marginTop: defaultTokens.spaceXSmall,
```
JoseDuSV: Can you please just switch margin values to padding. Because content is currently scrolling just within area with margin. And content is cut on top because of top margin.

screenshot 2019-02-25 at 14 46 40

tbergq: This approach will not move the scroll-bar out to the right even if we use padding, because the scroll-bar is attached to the div resulting from the ScrollView

tbergq: We might need to move the `ScrollView` out of `core` and wrap `web` and `mobile` with the scroll view instead. I will check it

JoseDuSV: Right, it worked with div but in this case the structure is slightly different.

We probably need to add extra View for example like this.
screenshot 2019-02-25 at 15 13 30

tbergq: The problem with that approach is that you are talking the scroll responsability away from the Scroll view and moving int to the `Container` component.

JoseDuSV: 👍 Right, I didn't noticed that there is `ScrollView` inside.

tbergq:
@JoseDuSV 🚀

PR #396, opened 2019-02-25
Feat: SectorDetail initial structure

JoseDuSV:
Summary: Added Sector details rendering to `ItineraryDetail` and `BookingDetail`.

Note: This PR contains only initial structure and proper `SectorDetail` content will be added separately to keep sane PR size.

### DIFF
--- a/apps/core/components/ItineraryCard/itineraryDetail/SectorsList.js
+++ b/apps/core/components/ItineraryCard/itineraryDetail/SectorsList.js
@@ -1,0 +1,29 @@
+// @flow
+
+import * as React from 'react';
+import { graphql, createFragmentContainer } from '@kiwicom/margarita-relay';
+
+import type { SectorsList as SectorType } from './__generated__/SectorsList.graphql';
+import SectorDetail from '../../sectorDetail/SectorDetail';
+
+type Props = {|
+  +data: ?SectorType,
+|};
+
+function SectorsList(props: Props) {
+  const sectors = props.data?.sectors ?? [];
+  return sectors.map((sector, sectorIndex) => {
+    return <SectorDetail key={sectorIndex} data={sector} />;
+  });
+}
+
+export default createFragmentContainer(
+  SectorsList,
+  graphql`
+    fragment SectorsList on Itinerary {
+      sectors {
+        ...SectorDetail
+      }
+    }
+  `,
+);
--- a/apps/core/components/ItineraryCard/ItineraryCard.js
+++ b/apps/core/components/ItineraryCard/ItineraryCard.js
@@ -102,6 +102,7 @@ class ItineraryCard extends React.Component<Props, State> {
         </Hoverable>
         {detailOpened && (
           <ItineraryDetail
+            data={data}
             localizedPrice={localizedPrice}
             onClose={this.handleDetailClose}
             onBookPress={onBookPress}
@@ -123,6 +124,7 @@ export default createFragmentContainer(
         currency
         amount
       }
+      ...ItineraryDetail
     }
   `,
 );
--- a/apps/core/components/ItineraryCard/__generated__/ItineraryCard.graphql.js
+++ b/apps/core/components/ItineraryCard/__generated__/ItineraryCard.graphql.js
@@ -8,6 +8,7 @@
 
 /*::
 import type { ReaderFragment } from 'relay-runtime';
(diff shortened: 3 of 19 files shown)


apps/core/components/ItineraryCard/itineraryDetail/SectorsList.js
```diff
@@ -0,0 +1,29 @@
...
+import SectorDetail from '../../sectorDetail/SectorDetail';
+
+type Props = {|
+  +data: ?SectorType,
+|};
+
+function SectorsList(props: Props) {
+  const sectors = props.data?.sectors ?? [];
+  return sectors.map((sector, sectorIndex) => {
+    return <SectorDetail key={sectorIndex} data={sector} />;
```
tbergq: The tequila api-response provides an ID for each sector. I think it is time we add that id to the Sector object

apps/core/components/ItineraryCard/itineraryDetail/SectorsList.js
```diff
@@ -0,0 +1,29 @@
...
+export default createFragmentContainer(
+  SectorsList,
+  graphql`
+    fragment SectorsList on Itinerary {
+      sectors {
+        ...SectorDetail
+      }
+    }
+  `,
+);
```
tbergq: also, if you change this fragment to 
```
export default createFragmentContainer(
  SectorsList,
  graphql`
    fragment SectorsList on Sector @relay(plural: true) {
      ...SectorDetail
    }
  `,
);
```

You should be able to reuse the sector list from all of the BookingType components as well

apps/core/components/ItineraryCard/itineraryDetail/SectorsList.js
jvaclavik: Are you sure? I think that there are ids for segment, but not sector.

tbergq: Oh, this might be what we discussed before. Yeah, I guess you are right.

apps/core/components/ItineraryCard/itineraryDetail/SectorsList.js
JoseDuSV: Maybe I'm missing something, but how I will then use it for example for `SectorsListReturn`?
Right now `SectorsList` is kind of equal to `SectorsListMulticity` just with different fragment definition.

tbergq:
```
unction SectorsListReturn(props: Props) {
  return (
    <>
      <SectorsList data={[props.data?.outbound, props.data?.inbound]} />
    </>
  );
}

export default createFragmentContainer(
  SectorsListReturn,
  graphql`
    fragment SectorsListReturn on BookingInterface {
      ... on BookingReturn {
        inbound {
          ...SectorsList
        }
        outbound {
          ...SectorsList
        }
      }
    }
  `,
);
```

JoseDuSV: Thanks for clarification. Updated and it definitely looks cleaner now.

apps/core/components/sectorsList/SectorsList.js
```diff
@@ -0,0 +1,31 @@
...
+import * as React from 'react';
+import { graphql, createFragmentContainer } from '@kiwicom/margarita-relay';
+
+import SectorDetail from './SectorDetail';
+
+type Props = {
+  +data: ?$ReadOnlyArray<?{|
+    +$fragmentRefs: any,
+  |}>,
+};
```
tbergq: This is a little bit weird, but I see that the generated flow types generates errors. We should check this against `@kiwicom/relay` package. Could you add some TODO comment so that we don't forget this.

apps/core/components/ItineraryCard/itineraryDetail/SectorsList.js
tbergq: Thanks, looks good now

apps/core/components/sectorsList/SectorsList.js
JoseDuSV: Right. TODO comment added.

JoseDuSV:
@tbergq Reverted back to separate lists version 🚀

PR #397, opened 2019-02-25
Add contact details form 

jvaclavik:
Android

screenshot 2019-02-25 at 16 34 45

iOS

screenshot 2019-02-25 at 16 35 13

Web

screenshot 2019-02-25 at 16 35 39

### DIFF
--- a/apps/core/scenes/resultDetail/ResultDetail.js
+++ b/apps/core/scenes/resultDetail/ResultDetail.js
@@ -2,13 +2,18 @@
 
 import * as React from 'react';
 import { View, ScrollView } from 'react-native';
-import { StyleSheet } from '@kiwicom/universal-components';
+import {
+  StyleSheet,
+  Card,
+  TextInput,
+  Picker,
+} from '@kiwicom/universal-components';
+import { HeaderWithIcon, PassengerCards } from '@kiwicom/margarita-components';
 import {
   withNavigation,
   Routes,
   type Navigation,
 } from '@kiwicom/margarita-navigation';
-import { PassengerCards } from '@kiwicom/margarita-components';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 
 type Props = {|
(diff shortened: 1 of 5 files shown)


apps/core/scenes/resultDetail/ResultDetail.js
```diff
@@ -2,17 +2,93 @@
...
+              onChangeText={this.handleChangeEmail}
+              label="E-mail"
+              type="email"
+              autoCorrect={false}
+            />
+          </View>
+          <View style={[styles.line, styles.row]}>
+            <View style={styles.countryCode}>
+              <Picker
+                selectedValue="+420"
```
JoseDuSV: We should probably use `selectedValue={this.state.phoneCountryCode}` for now. 
And later we can change default `phoneCountryCode` value according to location (but it is long-term nice to have).

jvaclavik: There is a problem that `selectedValue` could be `null`, which is not allowed in the `Picker`. Actually I think it's a bug in `Picker`, but there should be possibility of not choosing any value. Right now you have `placeholder` in the `Picker` which should be invisible all the time, because there is no way how to not set up the value.

JoseDuSV: Right. We should allow `Picker` selected value to be null. What about like this https://github.com/kiwicom/universal-components/pull/256 ?

jvaclavik: Perfect, thanks! 

It's merged, so I will change it and it should work in new `universal-components` release

packages/components/src/headerWithIcon/HeaderWithIcon.js
```diff
@@ -0,0 +1,46 @@
...
+export default function HeaderWithIcon({ label, iconName }: Props) {
+  return (
+    <View style={styles.container}>
+      <Icon
+        name={iconName}
+        color={defaultTokens.colorIconSecondary}
+        size="large"
+        style={styles.icon}
+      />
+      <Text style={styles.label} weight="bold">
```
JoseDuSV: You can use text from `@kiwicom/margarita-components` to make text bold even on iOS.

jvaclavik: Done

PR #399, opened 2019-02-25
Update LandingPage mobile styling

KarinaDavtyan:
Summary: update styling

Note:
+ iframe size is bigger than in design otherwise would be almost impossible to use
+ header is still not implemented

simulator screen shot - iphone 8 - 2019-02-25 at 10 46 43

### DIFF
--- a/apps/landingPage/src/components/Footer.js
+++ b/apps/landingPage/src/components/Footer.js
@@ -11,6 +11,7 @@ import {
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import styled from 'styled-components';
 
+import { BREAKPOINTS } from '../mediaQueriesConfig';
 import logo from '../images/logo.png';
 import {
   codeKiwiInstagram,
@@ -71,6 +72,11 @@ const FooterContainer = styled.div`
   justify-content: space-between;
   height: 105px;
   align-items: center;
+  @media (max-width: ${BREAKPOINTS.BIG_MOBILE}px) {
+    flex-direction: column;
+    height: 25vh;
+    justify-content: space-evenly;
+  }
 `;
 
 const Logo = styled.img`
@@ -90,5 +96,9 @@ const FooterRightContainer = styled.div`
   width: 14vw;
   justify-content: space-around;
   padding-right: 35px;
-  flex-wrap: wrap;
+  @media (max-width: ${BREAKPOINTS.BIG_MOBILE}px) {
+    width: 65vw;
+    padding: 0;
+    align-items: center;
+  }
 `;
--- a/apps/landingPage/src/components/Headline.js
+++ b/apps/landingPage/src/components/Headline.js
@@ -1,7 +1,7 @@
 // @flow
 
 import React from 'react';
-import { Heading, Text, Button } from '@kiwicom/orbit-components/lib/';
+import { Heading, Text } from '@kiwicom/orbit-components/lib/';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import styled from 'styled-components';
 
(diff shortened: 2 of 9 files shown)


apps/landingPage/src/components/Footer.js
```diff
@@ -71,6 +72,11 @@ const FooterContainer = styled.div`
   justify-content: space-between;
   height: 105px;
   align-items: center;
+  @media (max-width: ${BREAKPOINTS.BIG_MOBILE}px) {
```
tbergq: @RobinCsl should we follow `mobile-first` here as well?

apps/landingPage/src/components/Headline.js
```diff
@@ -13,87 +13,137 @@ import airport from '../images/airport.jpg';
...
+  }
+
+  handleWindowSizeChange = () => {
+    const width = typeof window !== 'undefined' ? window.innerWidth : 0;
+    this.setState({ width });
+  };
+
+  render() {
+    const alignment = this.state.width > BREAKPOINTS.TABLET ? 'left' : 'center';
+    const isTablet = this.state.width <= BREAKPOINTS.TABLET;
```
tbergq: would it be possible to achieve this with css as well?

apps/landingPage/src/components/Headline.js
```diff
@@ -13,87 +13,137 @@ import airport from '../images/airport.jpg';
...
+                  </Button>
+                </WrapperRightButton>
+              </WrapperButtons>
+            </ContainerLeftText>
+            <PlaneImage displayMobile="none" />
+          </ContainerBottom>
+        </Container>
+      </HeadlineContainer>
+    );
+  }
```
tbergq: Also this component is growing quite big, would it be possible to separate it to smaller logical components?

apps/landingPage/src/components/Info.js
```diff
@@ -8,44 +8,101 @@ import styled from 'styled-components';
...
+    window.removeEventListener('resize', this.handleWindowSizeChange);
+  }
+
+  handleWindowSizeChange = () => {
+    const width = typeof window !== 'undefined' ? window.innerWidth : 0;
+    this.setState({ width });
+  };
+
+  render() {
+    const alignment = this.state.width > BREAKPOINTS.TABLET ? 'left' : 'center';
```
tbergq: Also, if possible handle with css, if not, I see that this resize code is copy pasted. It could be extracted to a HOC.

apps/landingPage/src/components/Footer.js
RobinCsl: I believe we should follow `mobile-first` approach everywhere. Thanks for pointing it out!

tbergq: Let's just get this merged, and we can create a new PR to refactor media-queries to mobile first

apps/landingPage/src/components/Headline.js
KarinaDavtyan: Extracting *Get Started* part to the separate component

apps/landingPage/src/components/Headline.js
```diff
@@ -13,87 +13,137 @@ import airport from '../images/airport.jpg';
...
-                  color={defaultTokens.paletteWhite}
-                >
-                  Get started
-                </LinkNoStyle>
-              </Button>
-              <WrapperRightButton>
-                <Button type="white">
+            </HeaderAndDesignWrapper>
+            <ContainerIcons>
+              <img src={kiwiLogo} height={43} alt="kiwi_logo" />
```
jvaclavik: `alt` should be without `_`. It should be human readable (it can be automatically read by some scanner).

apps/landingPage/src/components/Headline.js
```diff
@@ -13,87 +13,137 @@ import airport from '../images/airport.jpg';
...
+                <Text size="large">
+                  PLACEHOLDER!!!!!You've had a look at Tequila, but you've
+                  preffered yourTequila Tequila mixed with something else. Just
+                  grab Margarita.You've had a look at Tequila, but you've always
+                  preffered your Tequila mixed with something else.
+                </Text>
+              </WrapperText>
+              <WrapperButtons>
+                <Button block={isTablet || false} size={buttonSize}>
                   <LinkNoStyle
```
jvaclavik: I would suggest rename it to more readable form `LinkWithoutStyle`

KarinaDavtyan: Alright, will rename it

apps/landingPage/src/components/Headline.js
KarinaDavtyan: Agree, will change, and I dont know why I am putting snake case everywhere 😓

apps/landingPage/src/components/Info.js
KarinaDavtyan: Should be better now, I've added `withSizeDetect` HOC

apps/landingPage/src/components/withSizeDetect.js
```diff
@@ -0,0 +1,29 @@
+// @flow
+
+import React from 'react';
+
+export default function WithSizeDetect(WrappedComponent) {
```
tbergq: maybe just `withWindowSize` is more descriptive.

apps/landingPage/src/components/withSizeDetect.js
```diff
@@ -0,0 +1,29 @@
+// @flow
+
+import React from 'react';
```
tbergq: You also have to `import * as React from 'react'` for flow types to work

apps/landingPage/src/components/withSizeDetect.js
KarinaDavtyan: Renamed

PR #389, opened 2019-02-25
Grahpql: Add stopCoordinates to Segment type

tbergq:
When drawing lines on the map we need to pass an array of
gps coordinates to the draw-line component. This field will make
the drawing of the map a lot easier than if we have to fetch
the coordinates from departure?.stop?.coordinates?.lat and
arrival?.stop?.coordinates?.lat

PR #406, opened 2019-02-26
MMB: Draw dots as map markers with location name

tbergq:
Remove segmentLocations from graphql BookingInterface type
see [comment](https://github.com/kiwicom/margarita/pull/389#issuecomment-466920186) for more info

### DIFF
--- a/apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
+++ b/apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
@@ -6,50 +6,121 @@ import { StyleSheet } from '@kiwicom/universal-components';
 import { graphql, createFragmentContainer } from '@kiwicom/margarita-relay';
 import { getCenter } from 'geolib';
 import memoize from 'memoize-one';
-import stringify from 'json-stable-stringify';
+import { uniqBy, prop } from 'ramda';
 
 import type { SegmentMap as BookingType } from './__generated__/SegmentMap.graphql';
 import MapLines from './MapLines';
+import SegmentMapMarker from './SegmentMapMarker';
 
 type Props = {|
   +data: ?BookingType,
 |};
 
+type RouteStop = {|
+  +stop: ?{|
+    +city: ?{|
+      +name: ?string,
+    |},
+    +locationId: ?string,
+    +coordinates: ?{|
+      +latitude: ?number,
+      +longitude: ?number,
+    |},
+  |},
+|};
+
+type Segments = ?$ReadOnlyArray<?{|
+  +arrival: ?RouteStop,
+  +departure: ?RouteStop,
+|}>;
+
+type MarkerData = {|
+  +latitude: ?number,
+  +longitude: ?number,
+  +name: ?string,
+  +locationId: string,
+|};
+
+type CompareByType = {|
+  +locationId: string,
+|};
+
 const DEFAULT_PADDING = { top: 40, right: 40, bottom: 40, left: 40 };
 
-const mapCoordinates = memoize(
-  (segmentLocations: ?$PropertyType<BookingType, 'segmentLocations'>) => {
-    const coordinates = segmentLocations ?? [];
-    return coordinates.map(coordinate => ({
-      latitude: coordinate?.lat,
-      longitude: coordinate?.lng,
-    }));
-  },
-);
+const mapMarkerData = memoize((props: Props) => {
+  switch (props.data?.type) {
+    case 'BOOKING_ONE_WAY':
+      return mapSegmentsToMarkerData(props.data?.sector?.segments);
+    case 'BOOKING_RETURN':
+      return mapSegmentsToMarkerData(getReturnSegments(props));
+    case 'BOOKING_MULTICITY':
+      return mapSegmentsToMarkerData(getMulticitySegments(props));
+    default:
+      return null;
+  }
+});
 
-class SegmentMap extends React.Component<Props> {
+const getReturnSegments = (props: Props) => {
+  const outbound = props.data?.outbound?.segments ?? [];
+  const inbound = props.data?.inbound?.segments ?? [];
+  return [...outbound, ...inbound];
+};
+
+const getMulticitySegments = (props: Props) => {
+  const sectors = props.data?.sectors ?? [];
+  return sectors.reduce((acc, curr) => {
+    const segments = curr?.segments ?? [];
+    return [...acc, ...segments];
+  }, []);
+};
+
+const mapSegmentsToMarkerData = (segments: Segments) => {
+  if (segments == null) {
+    return [];
+  }
+  const bookingSegments = segments.reduce((acc, curr) => {
+    return [
+      ...acc,
+      {
+        ...curr?.departure?.stop?.coordinates,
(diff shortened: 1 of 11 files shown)


apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
```diff
@@ -62,15 +129,15 @@ class SegmentMap extends React.Component<Props> {
         }}
         zoomEnabled={false}
```
jvaclavik: Btw are you sure that we shouldn't allow users to zoom in? 

It's hard to say if the point is in Vienna or Bratislava:
screenshot 2019-02-26 at 13 10 18

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMapMarker.js
```diff
@@ -0,0 +1,48 @@
...
+    position: 'absolute',
+    top: -9,
+    start: -3,
+    alignItems: 'center',
+  },
+  dot: {
+    backgroundColor: defaultTokens.paletteInkDark,
+    height: 6,
+    width: 6,
+    borderRadius: 3,
```
jvaclavik: We could use here `defaultTokens.borderRadiusNormal`, what do you think?

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
```diff
@@ -6,50 +6,117 @@ import { StyleSheet } from '@kiwicom/universal-components';
...
-    return coordinates.map(coordinate => ({
-      latitude: coordinate?.lat,
-      longitude: coordinate?.lng,
-    }));
-  },
-);
+const mapMarkerData = memoize((props: Props) => {
+  switch (props.data?.type) {
+    case 'BOOKING_ONE_WAY':
+      return mapSegmentsToMarkerData(props.data?.sector?.segments);
```
RobinCsl: Just to be uniform, could you extract this as a function `getOneWaySegments`?

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
JoseDuSV: I'm ok with disabled zoom. Window is quite small for any fiddling with map.
We can later implement expanded movable/scrollable map on tap like in mobile app.

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
```diff
@@ -6,50 +6,117 @@ import { StyleSheet } from '@kiwicom/universal-components';
...
+      },
+      {
+        ...curr?.arrival?.stop?.coordinates,
+        name: curr?.arrival?.stop?.city?.name,
+        locationId: curr?.arrival?.stop?.locationId ?? '',
+      },
+    ];
+  }, []);
+
+  return uniqBy<MarkerData, { +locationId: string }>(
```
RobinCsl: I find this difficult to read. Would it be possible to extract the types?

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
```diff
@@ -62,15 +129,15 @@ class SegmentMap extends React.Component<Props> {
...
       >
-        {coordinates.map(coordinate => (
-          <MapView.Marker
-            key={stringify(coordinate)}
-            coordinate={{
-              latitude: coordinate?.lat,
-              longitude: coordinate?.lng,
-            }}
+        {locations.map(location => (
+          <SegmentMapMarker
```
RobinCsl: Would it make sense to only pass `location` prop? and then deal with destructuring in the component?

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
tbergq: Current usage in (native)app is the map not zoomable or dragable. When you click on it, you open map in full-screen modal.

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
tbergq: I don't think that makes sense, because there are no computations to be done. I don't need a function to return `props.data?.sector?.segments`

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
tbergq: I prefer it this way, since if we are to rename any property in the location object I don't need to change anything in `SegmentMapMarker` component.

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMapMarker.js
tbergq: If someone updates `defaultTokens.borderRadiusNormal` to be another value, this dot would no longer appear circular. They are by coincidence the same value.

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
tbergq: 🚀

tbergq:
@RobinCsl @jvaclavik 🚀

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
```diff
@@ -6,50 +6,121 @@ import { StyleSheet } from '@kiwicom/universal-components';
...
         edgePadding: DEFAULT_PADDING,
       });
     }
   };
 
   render() {
-    const coordinates = this.props.data?.segmentLocations ?? [];
-    const center = getCenter(this.getMappedCoordinates());
+    const locations = this.getMarkerData() ?? [];
+    const center = getCenter(this.getMarkerData());
```
RobinCsl:
```suggestion
    const center = getCenter(locations);
```
To avoid invoking the function again.

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
jvaclavik: > When you click on it, you open map in full-screen modal.

I'm OK with that, it makes sense this way

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMapMarker.js
jvaclavik: Oh, I didn't realise it should be a circle. But still there is token for it: `defaultTokens.borderRadiusCircle`

jvaclavik:
@tbergq One comment and then 👍& 🚀

apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMapMarker.js
tbergq: image

PR #414, opened 2019-02-26
Refactor css to mobile-first approach in LandingPage

KarinaDavtyan:
Summary:
+ refactor media queries
+ fix some style inconsistencies with design

### DIFF
--- a/apps/landingPage/src/components/Demo.js
+++ b/apps/landingPage/src/components/Demo.js
@@ -29,7 +29,7 @@ export default function Demo() {
       <LinksWrapper>
         {links.map(el => {
           return (
-            <Button type="white" key={el.link}>
+            <Button type="white" key={el.title}>
               <LinkWithoutStyle href={el.link}>{el.title}</LinkWithoutStyle>
             </Button>
           );
@@ -48,18 +48,20 @@ const Container = styled.div`
 `;
 
 const HeadingWrapper = styled.div`
-  padding-top: 50px;
-  @media (max-width: ${BREAKPOINTS.BIG_MOBILE}px) {
-    padding-top: 0;
+  @media (min-width: ${BREAKPOINTS.BIG_MOBILE}px) {
+    padding-top: 40px;
   }
 `;
 
 const LinksWrapper = styled.div`
   display: flex;
-  width: 34vw;
+  width: 85vw;
   justify-content: space-around;
   padding: 30px 0 40px 0;
   flex-wrap: wrap;
+  @media (min-width: ${BREAKPOINTS.TABLET}px) {
+    width: 34vw;
+  }
 `;
 
 const LinkWithoutStyle = styled.a`
--- a/apps/landingPage/src/components/Headline.js
+++ b/apps/landingPage/src/components/Headline.js
@@ -1,7 +1,7 @@
 // @flow
 
 import React from 'react';
-import { Heading, Text } from '@kiwicom/orbit-components/lib/';
+import { Heading } from '@kiwicom/orbit-components/lib/';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import styled from 'styled-components';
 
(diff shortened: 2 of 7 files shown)


apps/landingPage/src/components/Headline.js
```diff
@@ -75,137 +75,139 @@ const HeadlineContainer = styled.div`
...
+  }
+`;
+
+const PlaneImageMobile = styled.div`
+  width: 94vw;
+  height: 280px;
   background-image: url(${airport});
   background-size: cover;
   background-repeat: no-repeat;
   background-position: 50% 50%;
```
JoseDuSV: Just a minor thing. But maybe we can set this to something around `50% 90%` to have plane in focus in some edge cases.

`50% 50%`
screenshot 2019-02-27 at 10 43 34

`50% 90%`
screenshot 2019-02-27 at 10 43 47

apps/landingPage/src/components/Demo.js
```diff
@@ -48,18 +48,21 @@ const Container = styled.div`
 `;
 
 const HeadingWrapper = styled.div`
-  padding-top: 50px;
-  @media (max-width: ${BREAKPOINTS.BIG_MOBILE}px) {
-    padding-top: 0;
+  padding-top: 0;
```
jvaclavik: Do we need this?

KarinaDavtyan: nope, no needed. Removing

apps/landingPage/src/components/Headline.js
KarinaDavtyan: Yes, looks much better like that.

JoseDuSV:
Just few small things not directly related to styling. But to clean console a bit. Then I think we can merge it.

https://github.com/kiwicom/margarita/blob/8135786d6340ed8af29ff404df229ca30735ad29/apps/landingPage/src/components/Headline.js#L28-L32
* This causes `<h1> cannot appear as a descendant of <p>` error. So we should probably use just `div` as a wrapper instead of `<Text>`. Same for row 40.

https://github.com/kiwicom/margarita/blob/8135786d6340ed8af29ff404df229ca30735ad29/apps/landingPage/src/components/Demo.js#L32-L34
* Right now we have a same link for github and docs. Which causes `Keys should be unique...` error. We need to change it to something unique (maybe just title can work).

KarinaDavtyan:
> Just few small things not directly related to styling. But to clean console a bit. Then I think we can merge it.
> 
> [margarita/apps/landingPage/src/components/Headline.js](https://github.com/kiwicom/margarita/blob/8135786d6340ed8af29ff404df229ca30735ad29/apps/landingPage/src/components/Headline.js#L28-L32)
> 
> Lines 28 to 32 in [8135786](/kiwicom/margarita/commit/8135786d6340ed8af29ff404df229ca30735ad29)
> 
>  <Text align={alignment}> 
>    <Heading type="display"> 
>      Ever dreamt of building the next Kiwi.com? 
>    </Heading> 
>  </Text> 
> * This causes `<h1> cannot appear as a descendant of <p>` error. So we should probably use just `div` as a wrapper instead of `<Text>`. Same for row 40.
> 
> [margarita/apps/landingPage/src/components/Demo.js](https://github.com/kiwicom/margarita/blob/8135786d6340ed8af29ff404df229ca30735ad29/apps/landingPage/src/components/Demo.js#L32-L34)
> 
> Lines 32 to 34 in [8135786](/kiwicom/margarita/commit/8135786d6340ed8af29ff404df229ca30735ad29)
> 
>  <Button type="white" key={el.link}> 
>    <LinkWithoutStyle href={el.link}>{el.title}</LinkWithoutStyle> 
>  </Button> 
> * Right now we have a same link for github and docs. Which causes `Keys should be unique...` error. We need to change it to something unique (maybe just title can work).

+ Yeah, nesting `<Text>` and `<Heading>` was ugly. Hopefully `<Heading>`  will get `align` prop as well.  Now its wrapped in `div`
+ Alright, changed for title

PR #420, opened 2019-02-27
Use __typename instead of booking type enum

tbergq:


.eslintignore
```diff
@@ -8,3 +8,4 @@ apps/web/.next
 apps/landingPage/.cache
 apps/landingPage/public
 apps/storybook/config/storyLoader.js
+packages/components/storybook/story-loader.js
```
RobinCsl: This file should be deleted now, I believe.

Issue #114, opened 2019-01-14
Investigate: Where get data for Booking screen and design GraphQL schema for that.


FilipMessa:
We will need implementing the GraphQL schema for Booking screen.

Related Issues  #72, #73 

Check the Tequila API documentation and suggest the GraphQL schema.

*Pls don't start implemented it yet, just prepare the schema for CR first*

FilipMessa:
If you will need to call `/booking` API, you must use a different `apikey` than current. There are different products with different policies, so check https://tequila.kiwi.com/portal/my-applications  and look for the `apikey` inside the *searchandbook* application.

JoseDuSV:
Proposal for `/booking/check_flights` implementation from Booking API https://tequila.kiwi.com/portal/resources/booking_api.

In short, we need to check if itinerary is valid/bookable and receive actual itinerary data to show them inside booking screen header #72 .

Query name:
```
checkItinerary
```

Input:
```
input CheckItineraryInput {
  """Token received in search results"""
  bookingToken: String!

  """Number of bags for the booking"""
  bags: Int!

  """Number of children, adults and infants travelling"""
  passengers: PassengersInput!
}
```

Return:
```
type ItineraryValidity {
  """All segments successfully checked"""
  itineraryChecked: Boolean

  """Itinerary is valid (not sold-out or cancelled)"""
  itineraryValid: Boolean

  """Actual itinerary data"""
  itinerary: Itinerary
}
```

What do you think @FilipMessa @tbergq @jvaclavik @KarinaDavtyan @RobinCsl ?

FilipMessa:
LGTM
I would like to ask, just for my information, bags can be zero, right? and passengers must be at least 1, right?

JoseDuSV:
@FilipMessa Right, passengers number can be 1-9. And number of bags can be 0.

jvaclavik:
And what if there are more types of the bag? For example with Ryanair you can have personal item (free), cabin bag (paid) and large bag (paid)

JoseDuSV:
@jvaclavik According to description it seems to be related only to checked baggages `The number of bags for the booking, even if bags_price states that the first (or even second) checked baggage is free, it is necessary to request it`.

jvaclavik:
> @jvaclavik According to description it seems to be related only to checked baggages `The number of bags for the booking, even if bags_price states that the first (or even second) checked baggage is free, it is necessary to request it`.

OK, in that case it looks fine

[...]

PR #421, opened 2019-02-27
Feat: Added bookingToken to Itinerary

JoseDuSV:
Summary: Added bookingToken to Itinerary type. And implemented it for redirect from `/results` to `/result-detail`.

It will be required for Booking API calls ( related to #114 )

### DIFF
--- a/apps/core/components/ItineraryCard/itineraryDetail/ItineraryDetail.js
+++ b/apps/core/components/ItineraryCard/itineraryDetail/ItineraryDetail.js
@@ -14,12 +14,12 @@ type Props = {|
   +data: ?ItineraryType,
   +localizedPrice: string,
   +onClose: () => void,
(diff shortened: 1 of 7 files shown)


apps/core/components/ItineraryCard/itineraryDetail/ItineraryDetail.js
```diff
@@ -14,12 +14,12 @@ type Props = {|
   +data: ?ItineraryType,
   +localizedPrice: string,
   +onClose: () => void,
-  +onBookPress: string => void,
+  +onBookPress: (token: ?string) => void,
```
jvaclavik: Why do you use `token` here? Why not just `?string`?

JoseDuSV: Right, updated.

PR #424, opened 2019-02-27
Chore: Moving `universal-components` into `margarita`

RobinCsl:
NOTE: There are still many issues with tests and flow

- tests: `universal-components` used `react-native` jest preset and `margarita` uses `jest-expo`; I believe this creates some of the test errors.
    - another family of errors is linked to which files are being resolved by `jest`, and if there are `Button.web.js` and `Button.native.js`, it's not clear that Jest decides correctly (even more visible if you instead have `Button.js` and `Button.native.js`)

- flow: we have `flow-typed` types (which I cannot find the source on flow-typed 🤷‍♂️) for `react-native-web`, which allows our setup to "work"; but removing this flow-typed definitions and keeping the .flowconfig `module_name_mapper` option to map `react-native` to `react-native-web`, means we get errors in `apps/mobile` as we don't have `react-native-web` installed there (and rightly so!)
These flow-typed definitions are not complete nor fully accurate, so I would be in favour of removing them, but I don't have a solution for `apps/mobile` to work correctly otherwise. It seems flow configs per workspace are not a thing. (We could always .flowconfig ignore apps/mobile and then have another flow config there..., and execute two commands for flow to be checking the entirety of the repo...)

A few things that also need to be addressed (difficult to separate this in multiple PR's sadly):
- build process needs to be improved: I think one of the reasons why `jest` is not picking up the right files is because we are copying `Button.js` and `Button.native.js` in `lib/native`, and `Button.js` corresponding to the web version should not be copied over...

- need to figure out if we want "main" and "react-native" fields in universal-components' package.json file pointing to the lib folders (in which case, you would need to do a yarn build before being able to use them in your projects -- still a lot easier than waiting for another npm release!), or we have just a "main" field pointing to `src/index.js` and we forget about NPM publishing our universal-components package for now. 
The build process could also automate that behaviour of altering the `package.json` to our needs, but it feels kind of _hacky_ so I don't know yet if I like this approach.

Hopefully I'll manage to deal with these issues tomorrow...


michalsanger:
> we have `flow-typed` types (which I cannot find the source on flow-typed)

aren't those just generated stubs?

michalsanger:
Is there a reason not copy folder structure of our monorepo? I'm missing having everything under `src` folder

RobinCsl:
> aren't those just generated stubs?

Yes, they are. But I don't know where they come from as I cannot find them. Inquiring with @FilipMessa 

> Is there a reason not copy folder structure of our monorepo? I'm missing having everything under `src` folder

The only concern of this PR is to move code from `universal-components` to `margarita`. If we are doing something that's not in accordance with the monorepo's file structure, we can do a reorg in another PR if you don't mind.

RobinCsl:
Closing in favour of #451

PR #425, opened 2019-02-27
Create passenger cards

jvaclavik:
**What is new:**
- Add component for list of the `PassengerCard`s (called `PassengerCards`)
- Add button for adding new passenger (`Button` should be rounded, it will be updated in some next PR)
- Click on `PassengerCard` edit button returns `ID`
- Some refactoring of `PassengerCard`

### DIFF
--- a/apps/core/scenes/resultDetail/ResultDetail.js
+++ b/apps/core/scenes/resultDetail/ResultDetail.js
@@ -1,10 +1,10 @@
 // @flow
 
 import * as React from 'react';
-import { View, ScrollView } from 'react-native';
+import { View } from 'react-native';
 import { StyleSheet } from '@kiwicom/universal-components';
 import { withNavigation } from '@kiwicom/margarita-navigation';
-import { PassengerCard } from '@kiwicom/margarita-components';
+import { PassengerCards } from '@kiwicom/margarita-components';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 
 type Props = {|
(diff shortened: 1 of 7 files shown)


apps/core/scenes/resultDetail/ResultDetail.js
```diff
@@ -1,49 +1,57 @@
...
+        id: 'DF45SV9',
+        insurance: 'Travel Insurance Name',
+        passengerCount: 1,
+        bags: [{ count: 1, type: '40x15x30cm, 3kg' }],
+      },
     ];
+
     return (
       <View style={styles.container}>
-        <ScrollView>
```
JoseDuSV: I think we should keep `<ScrollView>`, because we  want whole page with header and contact details scrollable.

packages/components/src/passengerCards/PassengerCards.js
```diff
@@ -0,0 +1,67 @@
...
+  handleAddPassenger = () => {
+    // @TODO
+  };
+
+  // @TODO tokens: we should use different color than colorTextButtonPrimary (it's not text, but color should be white)
+  // @TODO tokens: there is no size 48px for button/icon
+  // @TODO button should be rounded
+  render() {
+    return (
+      <ScrollView contentContainerStyle={styles.scrollView}>
```
JoseDuSV: And this one should be just `View`.

packages/components/src/passengerCard/PassengerCard.js
```diff
@@ -45,94 +25,112 @@ function getTitle(gender) {
...
-    },
+    marginBottom: parseFloat(defaultTokens.spaceSmall),
   },
   containerName: {
     flexDirection: 'row',
     alignItems: 'center',
-    paddingTop: parseInt(defaultTokens.spaceSmall, 10),
-    paddingBottom: parseInt(defaultTokens.spaceMedium, 10),
+    paddingTop: parseFloat(defaultTokens.spaceSmall),
+    paddingBottom: parseFloat(defaultTokens.spaceMedium),
```
JoseDuSV: I'm just curious, but is this change necessary? Personally I like `parseInt` slightly more and I'll not really expect padding from tokens to be float.
But we can definitely keep it this way.

jvaclavik: It's not necessary, you are right it's useless change, According to me it looks more readable without second argument. We can keep it as it was and it hopefully it will be changed in the future.

apps/core/scenes/resultDetail/ResultDetail.js
jvaclavik: 👍 Done

packages/components/src/passengerCards/PassengerCards.js
jvaclavik: Done

packages/components/src/passengerCard/PassengerCard.js
jvaclavik: I changed it back to original implementation

packages/components/src/passengerCards/PassengerCards.js
```diff
@@ -0,0 +1,67 @@
...
+      </View>
+    );
+  }
+}
+
+const styles = StyleSheet.create({
+  addButtonContainer: {
+    alignItems: 'center',
+  },
+  scrollView: {
```
JoseDuSV: Can we please name it something like `container`?

jvaclavik: done

PR #433, opened 2019-02-28
Feat: Added Segments info to SectorDetail

JoseDuSV:
Summary: Added Segments and Layover info.

screenshot 2019-02-28 at 10 10 47

### DIFF
--- a/apps/core/components/sectorDetail/SectorDetail.js
+++ b/apps/core/components/sectorDetail/SectorDetail.js
@@ -3,7 +3,9 @@
 import * as React from 'react';
 import { graphql, createFragmentContainer } from '@kiwicom/margarita-relay';
 
+import Segment from './segment/Segment';
 import SectorHeader from './SectorHeader';
+import LayoverInfo from './LayoverInfo';
 import SectorStopoverDuration from './SectorStopoverDuration';
 import type { SectorDetail as SectorType } from './__generated__/SectorDetail.graphql';
 
@@ -12,11 +14,30 @@ type Props = {|
 |};
 
 const SectorDetail = (props: Props) => {
+  const segmentsData = props.data?.segments ?? [];
+  const segments = segmentsData.map((data, index) => {
+    if (data) {
+      const addLayoverInfo = index < segmentsData.length - 1;
+      return (
+        <React.Fragment key={data.id}>
+          <Segment data={data} />
+          {addLayoverInfo && (
+            <LayoverInfo
+              startTime={data.arrival?.time?.local}
+              endTime={segmentsData[index + 1]?.departure?.time?.local}
+            />
+          )}
+        </React.Fragment>
+      );
+    }
+    return null;
+  });
+
   return (
     <>
       <SectorStopoverDuration data={props.data} />
       <SectorHeader data={props.data} />
-      {/** @TODO segments + layover info */}
+      {segments}
     </>
   );
 };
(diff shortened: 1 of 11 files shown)


apps/core/components/sectorDetail/segment/Segment.js
```diff
@@ -0,0 +1,88 @@
...
+  getFormattedDate,
+  dateFormat,
+} from '../../ItineraryCard/TripSectorHelpers';
+import type { Segment as SegmentType } from './__generated__/Segment.graphql';
+
+type Props = {|
+  +data: ?SegmentType,
+|};
+
+class Segment extends React.Component<Props> {
```
FilipMessa: Is the reason why it is class?

apps/core/components/sectorDetail/segment/Segment.js
```diff
@@ -0,0 +1,88 @@
...
+        />
+        <SegmentStopInfo
+          data={this.props.data?.departure}
+          typeLabel="Departure"
+        />
+        <TimelineTransportDetail
+          carrier={{
+            code: this.props.data?.transporter?.name ?? '',
+            name: 'Carrier name', // @TODO - mock value, value is not returned from api yet
+            type: 'airline', // @TODO - mock value, carrier_type value is not returned from api yet
```
FilipMessa: May I ask you for creating the issue for that? thanks

apps/core/components/sectorDetail/SectorDetail.js
```diff
@@ -12,11 +14,29 @@ type Props = {|
 |};
 
 const SectorDetail = (props: Props) => {
+  const segmentsData = props.data?.segments ?? [];
+  const segments = segmentsData.map((data, index) => {
+    if (data) {
+      return (
+        <React.Fragment key={data.id}>
+          <Segment data={data} />
+          {index < segmentsData.length - 1 && (
```
jvaclavik: Maybe this could be separated in some const, it could be more readable, what do you think?

apps/core/components/sectorDetail/segment/Segment.js
JoseDuSV: Definitely not. Thanks, updated

apps/core/components/sectorDetail/segment/SegmentConstants.js
```diff
@@ -0,0 +1,59 @@
...
+      },
+      {
+        icon: 'seat',
+        label: 'Seat recline',
+        value: '3"',
+      },
+      {
+        icon: 'entertainment',
+        label: 'Audio & video on demand',
+        value: 'Yes',
```
jvaclavik: Why not use `true`/`false` instead of `'Yes'`?

apps/core/components/sectorDetail/SectorDetail.js
JoseDuSV: 👍 separated

apps/core/components/sectorDetail/segment/SegmentConstants.js
JoseDuSV: Those data were basically prepared according to design https://www.figma.com/file/XD9kQTOJxaqviaxD0g3AhAXS/Booking-Detail .
And it's really just mock for now. We are currently not sure if we will be even able to dig this info from api and in which form. Maybe in this case it can also be something like `just audio` but we don't know yet.

apps/core/components/sectorDetail/segment/Segment.js
JoseDuSV: Definitely, issue created: https://github.com/kiwicom/margarita/issues/435

FilipMessa: thanks

apps/core/components/sectorDetail/segment/SegmentConstants.js
jvaclavik: Hmm, true.., OK

PR #437, opened 2019-02-28
Upgrade @kiwicom/relay

tbergq:


apps/core/scenes/bookingDetail/sectorDetails/segments/SegmentMap.js
```diff
@@ -157,10 +158,9 @@ const styles = StyleSheet.create({
   },
 });
 
-export default createFragmentContainer(
-  SegmentMap,
-  graphql`
-    fragment SegmentMapStop on RouteStop {
+export default createFragmentContainer(SegmentMap, {
+  segmentMapStop: graphql`
+    fragment SegmentMap_segmentMapStop on RouteStop {
```
mrtnzlml: @tbergq This change is good because the previous multi-fragment code is not even supported in Relay anymore. However, you could define it outside of `createFragmentContainer` to avoid creating these unused props. A better solution would be to decompose it properly though or update your server as you suggested before.

tbergq: Yes, I would like to have it computed on the server, but it seemed @michalsanger was not a big fan of this approach.

michalsanger: @tbergq me? One of your biggest fans?! :-) Could you explain what is the change about and what was my problem with it please?

tbergq: https://github.com/kiwicom/margarita/pull/389#issuecomment-466920186

michalsanger: I was not a fan of adding new fields into schema, I was suggesting use fragments for re-use. IMHO this principle is still OK, just cleaner syntax was used.

mrtnzlml: @tbergq Didn't hear this reasoning for almost 24 hours. 😄 Next time when you see this I would reconsider to define it [outside of `
createFragmentContainer`](https://facebook.github.io/relay/docs/en/graphql-in-relay.html#relaymask-boolean) as I suggested before. It's because it seems like you are exposing these internal fragments as props. Actually, Relay is usually quite unhappy when you do this and it shows warnings like this one:

> Warning: createFragmentSpecResolver: Expected prop `segmentMapStop` to be supplied to `Relay(Location)`, but got `undefined`. Pass an explicit `null` if this is intentional.

This is not happening in your case? 🤔 Not even talking about the fact that you copy-paste or write manually Flow types here for the fetched fields. 😞

PR #438, opened 2019-02-28
Refactor Landing Page components into smaller ones

KarinaDavtyan:
`Headline` and `Info` related components are in the folders now


apps/landingPage/src/components/Headline/Subtitle.js
```diff
@@ -0,0 +1,95 @@
...
+|};
+
+export default function Subtitle(align: Props) {
+  return (
+    <Container>
+      <HeaderAndDesignWrapper>
+        <BackgroundGradientAccent>
+          <GreenRectangle />
+        </BackgroundGradientAccent>
+        <Text align={!align} type="title2" text={bodyText} />
```
jvaclavik: Maybe we should keep it similar with Text from margarita and have text as a children

KarinaDavtyan: Changed `prop` to children and also renamed component to the `TextHeadline`

tbergq: this looks weird `align={!align}`

KarinaDavtyan: is `align={align}` less weird? as I just changed it to that

tbergq: Yes, it is better 😄

apps/landingPage/src/components/Info/Features.js
```diff
@@ -0,0 +1,72 @@
...
+import React from 'react';
+import { Heading } from '@kiwicom/orbit-components/lib/';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import styled from 'styled-components';
+
+import { BREAKPOINTS } from '../../mediaQueriesConfig';
+
+const features = ['Search', 'Booking', 'MMB', 'Payments'];
+
+const renderCards = (features: Array<string>) => {
```
tbergq: could you use `$ReadOnlyArray<string>` instead?

KarinaDavtyan: Sure! changed now

apps/landingPage/src/components/Info/Features.js
```diff
@@ -0,0 +1,72 @@
...
+  ));
+};
+
+export default function Features() {
+  return (
+    <Container>
+      <Title>
+        <Heading>Features</Heading>
+      </Title>
+      <CardListWrapper>{renderCards(features)}</CardListWrapper>
```
tbergq: Also, I don't understand why renderCards is a function. It looks and acts like a component. 
What about 
```
const RenderCards = ({features}: Props) => {
return features.map(feature => (
    <CardWrapper key={feature}>
      <Heading type="title2">{feature}</Heading>
    </CardWrapper>
  ));
}

...
 <CardListWrapper><RenderCards features={features} /></CardListWrapper>
```

KarinaDavtyan: Ok! changed to the component

KarinaDavtyan:
> Looks much better now with smaller components 🙂

Yes, it is! 🙂

PR #450, opened 2019-03-01
Move VisaDetail to packages/components

KarinaDavtyan:


apps/core/scenes/bookingDetail/menuGroups/Passengers/VisaDetail.js
```diff
@@ -1,67 +1,31 @@
...
+|};
 
 const isVisaRequired = (
   passengers: $ReadOnlyArray<?VisaRequiredType>,
 ): boolean => passengers.some(passenger => passenger?.visaRequired);
 
-const VisaDetail = (props: Props) => {
+const Visa = (props: Props) => {
   const passengers = props.data?.passengers;
   const visaRequired = passengers && isVisaRequired(passengers);
```
tbergq: I know this is from a previous PR, but I think we should rather handle the case where passengers is null/undefined in `isVisaRequired`

```
const isVisaRequired = (
  passengers: ?$ReadOnlyArray<?VisaRequiredType>,
): boolean => {
if (passengers == null) return false;
return passengers.some(passenger => passenger?.visaRequired);	): boolean => passengers.some(passenger => passenger?.visaRequired);
}
```

then just call it like `isVisaRequired(props.data?.passengers)`

packages/components/src/visaDetail/VisaDetail.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { StyleSheet, Text, Icon } from '@kiwicom/universal-components';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+type Props = {
+  +visaRequired: ?boolean,
+};
```
tbergq: Type should be exact

apps/core/scenes/bookingDetail/menuGroups/Passengers/VisaDetail.js
```diff
@@ -1,67 +1,31 @@
...
   +visaRequired: ?boolean,
-};
+|};
 
 const isVisaRequired = (
   passengers: $ReadOnlyArray<?VisaRequiredType>,
 ): boolean => passengers.some(passenger => passenger?.visaRequired);
 
-const VisaDetail = (props: Props) => {
+const Visa = (props: Props) => {
```
jvaclavik: Also file should be renamed

jvaclavik: There are also some names like 
```
import { VisaDetail } from '@kiwicom/margarita-components';
import type { VisaDetail_data as VisaDetailType } from './__generated__/VisaDetail_data.graphql';

...
  +data: ?VisaDetailType,

```

With VisaDetail, it should be consistent.

apps/core/scenes/bookingDetail/menuGroups/Passengers/VisaDetail.js
KarinaDavtyan: Changes, looks better now

packages/components/src/visaDetail/VisaDetail.js
KarinaDavtyan: Fixed!

apps/core/scenes/bookingDetail/menuGroups/Passengers/VisaDetail.js
KarinaDavtyan: Now its `VisaInfoSummary` as its info for all the passengers in the booking, and the component itself called `VisaInfo`

PR #451, opened 2019-03-01
Chore/moving universal into margarita

RobinCsl:


RobinCsl:
Web version does not work yet; this is due to babel building the files differently from `universal-components` for some reason. To be continued.

RobinCsl:
Once it's merged, need to freeze universal-components repo.

RobinCsl:
We'll also have to check how it's going to work with `release-it`: it generated some changelog based on past commits, which might include commits not related to changes into `universal-components`. Likewise, if we make a release and tag it, what naming convention should we use, etc.?

(I believe we should use `@kiwicom/npm-publisher` now/in the future.)

packages/universal-components/package.json
```diff
@@ -0,0 +1,86 @@
+{
+  "name": "@kiwicom/universal-components",
+  "version": "0.0.14",
+  "description": "Component Library for React Native, Expo and React applications",
+  "homepage": "https://github.com/kiwicom/universal-components",
+  "main": "lib/web/index.js",
```
JoseDuSV: Works and looks perfect! Thanks a lot.

The only thing I'm not sure about is that we are using `universal-components` from `/lib`.
So build needs to be created first. We should probably make more clear that calling `yarn workspace @kiwicom/universal-components build` is necessary.

Or better, will it be possible (and will it make sense to you) to use components directly from `universal-components/src`? I think that now it can be confusing that for `@kiwicom/universal-components` you need to create build first, but it is not necessary for `@kiwicom/margarita-components`.

RobinCsl: You're right, thanks for pointing it out.

I'm still not decided which way is best.

I guess for now I could add a postinstall hook that will build the component library for you (which also means that if you are making changes in universal, you need to rebuild it to use it elsewhere...)

Will need to discuss with @mrtnzlml about monorepo architecture and package usage/release.

JoseDuSV: Thanks. I think that postinstall will work for now.

RobinCsl:
Added postinstall hook as @JoseDuSV requested so that `universal-components/lib`  is built. 
Will merge after tests/deploy succeed.

jvaclavik:
👍

PR #452, opened 2019-03-01
Move inline fragments outside of createFragmentContainer

tbergq:


tbergq:
@mrtnzlml copy-pasted flow types removed 🚀

PR #480, opened 2019-03-04
Fix: Modal web stacking context

### LINKED ISSUE #485: BUG: The Logo overlays modall
screen shot 2019-03-05 at 06 11 20

JoseDuSV:
Summary: Updated `Modal` web implementation to fix stacking context problem.
Removed web related tests, because new solution is using `document.body` which is not currently available in our test environment.

Before:
screenshot 2019-03-04 at 15 51 08

After:
screenshot 2019-03-04 at 15 50 14

Note: New solution is using React Portals. 
Next.js example: https://github.com/zeit/next.js/tree/canary/examples/with-portals

Closes #485


jvaclavik:
Seems to me little bit "hacky", that creating, appending and force updating DOM in `componentDidMount`. But I guess you know why you did it that way :)

JoseDuSV:
@jvaclavik It's basically recommended way how to use React Portals in next.js. As seen in example at the end of description. But I totally understand your point, I was also slightly suspicious at the beginning :)

Issue #449, opened 2019-03-01
Add PriceSummary component


jvaclavik:
<img width="264" alt="screenshot 2019-03-01 at 12 53 46" src="https://user-images.githubusercontent.com/858321/53636617-20007900-3c21-11e9-96c6-f5f180d8d688.png">
<img width="191" alt="screenshot 2019-03-01 at 12 48 51" src="https://user-images.githubusercontent.com/858321/53636619-23940000-3c21-11e9-9429-69173ebc1865.png">

PR #483, opened 2019-03-04
Add PriceSummary

jvaclavik:
Create components for for PriceSummary:
https://github.com/kiwicom/margarita/issues/449

mar-05-2019 15-44-14

**Notes:** 
a) in current version overlay background is lighter, but I don't want to re–create gif video.
b) in iOS device text overlay is in the header, it's not a problem, I didn't use header during the development

## Done
- [x] Styling iOS, Android, web
- [x] Add custom elements (links-like Text) to `value` prop in `PriceRow`
- [x] What about bold Text in Total price?
- [x] ~~Use modal window~~?
- [x] Delete testing condition from Search


apps/core/components/priceSummary/Expander.js
```diff
@@ -0,0 +1,56 @@
...
+
+const ArrowExpander = ({ expanded }) => (
+  <Icon
+    name={expanded ? 'chevron-down' : 'chevron-up'}
+    color={defaultTokens.borderColorTableCell}
+    style={styles.arrowIcon}
+  />
+);
+
+class Expander extends React.Component<Props> {
```
FilipMessa: Is the class necessary?

apps/core/components/priceSummary/PriceSummary.js
```diff
@@ -0,0 +1,159 @@
...
+    );
+  };
+
+  render() {
+    const {
+      renderCollapseContent,
+      renderVisibleContent,
+      renderButtons,
+      onButtonPress,
+      buttonLabel = 'Continue',
```
FilipMessa: why not defaultProps?

apps/core/scenes/resultDetail/ResultDetail.js
```diff
@@ -93,6 +105,34 @@ class ResultDetail extends React.Component<Props, State> {
             onChangeCountryCode={this.handleChangePhoneCountryCode}
           />
         </ScrollView>
+        <PriceSummary
```
FilipMessa: what about the React compound pattern? It will be more readable I think but it is just an idea. 
```
<PriceSummary>
<PriceSummary.Visible>
	subtotal
<PriceSummary.Visible>
<PriceSummary.CollapseContent>
	<TableRow label="1x Passenger" value="123 Kč" />
    <TableRow label="1x Cabin bag" value="123 Kč" />
    <TableRow label="1x Personal item" value="123 Kč" />subtotal
</PriceSummary.CollapseContent>
</PriceSummary>
```
What do you think?
https://blog.logrocket.com/guide-to-react-compound-components-9c4b3eb482e9

packages/components/src/table/TableRow.js
```diff
@@ -0,0 +1,84 @@
...
+  +highlightedLabel?: boolean,
+  +renderLabel?: React.Node,
+  +value?: string,
+  +highlightedValue?: boolean,
+  +renderValue?: React.Node,
+  +onPress?: () => void,
+  +style?: StylePropType,
+|};
+
+class TableRow extends React.Component<Props> {
```
FilipMessa: why class?

packages/components/src/table/TableRow.js
```diff
@@ -0,0 +1,84 @@
...
+  renderRow = () => {
```
FilipMessa: Can you make a component from this method?

apps/core/scenes/resultDetail/ResultDetail.js
jvaclavik: I'm not sure about that, because 
- you can't ensure that there will be only one component of this type (`PriceSummary.Visible`, `<PriceSummary.CollapseContent>`)
- Also component should respect position where component is rendered in children. But we can't do it. What if `PriceSummary.CollapseContent` and `PriceSummary.Visible` will be inverted? I think behaviour should be the same, but I'm not sure if this design is good for us in that case.

FilipMessa: OK it was just an idea, so keep it as it is 😄

apps/core/components/priceSummary/Expander.js
jvaclavik: No, refactored

apps/core/components/priceSummary/PriceSummary.js
jvaclavik: True, refactored

packages/components/src/table/TableRow.js
jvaclavik: Refactored

packages/components/src/table/TableRow.js
jvaclavik: Yes, done

packages/components/storybook/story-loader.js
```diff
@@ -0,0 +1,17 @@
+// @flow
```
JoseDuSV: I think that this whole file should be ignored.

jvaclavik: Removed from commit (actually I'm not sure why it's here)

JoseDuSV: Thanks, it was probably just some leftover.

PR #488, opened 2019-03-05
Fix: remove window dependent styles

### LINKED ISSUE #422: Landing Page is not displayed properly on the moment its opened
Its around a second for desktop and a bit longer for mobile, but at least its not happening on every refresh

screenshot 2019-02-27 at 13 49 18

KarinaDavtyan:
Summary: remove HOC `withWindowSize` and instead of that use media queries

Closes #422

### DIFF
--- a/apps/landingPage/components/Info/ResponsiveText.js
+++ b/apps/landingPage/components/Info/ResponsiveText.js
@@ -1,0 +1,39 @@
+// @flow
+
+import React from 'react';
+import { Text } from '@kiwicom/orbit-components/lib/';
+import styled from 'styled-components';
+
+import { BREAKPOINTS } from '../../mediaQueriesConfig';
+
+const text =
+  "PLACEHOLDER!!!!!You've had a look at Tequila, but you've preffered yourTequila Tequila mixed with something else. Just grabMargarita.You've had a look at Tequila, but you've always prefferedyour Tequila mixed with something else.u've had a look at Tequ";
+
+export default function ResponsiveText() {
+  return (
+    <Wrapper>
+      <Text size="large" align="left">
+        {text}
+      </Text>
+    </Wrapper>
+  );
+}
+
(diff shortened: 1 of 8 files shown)


apps/landingPage/components/Info/ResponsiveText.js
```diff
@@ -0,0 +1,54 @@
...
+    display: none;
+  }
+`;
+
+const Desktop = styled.div`
+  display: none;
+  @media (min-width: ${BREAKPOINTS.BIG_MOBILE}px) {
+    display: flex;
+  }
+`;
```
FilipMessa: These `Mobile` and `Desktop` styled components look the same as in the `ButtonInfo.js` so probably we can extract them and reuse. 🤔

KarinaDavtyan: Should I extract them together with `Tablet` into `landingPage/components/styles/index.js` ?

FilipMessa: 🤔 I didn't realize that. So, in that case, it is OK.

apps/landingPage/components/Info/ResponsiveText.js
```diff
@@ -0,0 +1,54 @@
...
+          {text}
+        </Text>
+      </Mobile>
+      <Desktop>
+        <Text size="large" align="center">
+          {text}
+        </Text>
+      </Desktop>
+    </WrapperText>
+  );
```
FilipMessa: With this approach there will be the same text two times in HTML, shouldn't be better do it with nested styles? 
something like:
```
 @media (min-width: ${BREAKPOINTS.BIG_MOBILE}px) {
	  > p {
	      text-align: center;
	    }
```

KarinaDavtyan: Much better like that!

PR #494, opened 2019-03-06
Fix: Set correct TextInput fontFamily

JoseDuSV:
Summary: Switched from system default fontFamily to `Roboto`.

Before:
screenshot 2019-03-06 at 12 38 14

After (Roboto is slightly more compact):
screenshot 2019-03-06 at 12 37 34


jvaclavik:
Am I right that we can't use `Circular Pro` because it's paid font and Margarita is open source?

JoseDuSV:
@jvaclavik Exactly

Issue #477, opened 2019-03-04
Inquire which flow warnings can be removed + why flowtests would pass without FlowExpectedErrors


RobinCsl:
Currently, `yarn flow` gives the following output:

```
Warning ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ packages/components/src/alert/AlertContext.js:4:1

Unused suppression comment.

      1│ // @flow
      2│
      3│ import * as React from 'react';
      4│ /**
      5│  *  $FlowFixMe:
      6│  *  >Cannot import Alert because there is no Alert export in react-native.
      7│  *  Likely related to the fact that `react-native-web` does not export Alert.
      8│  */
      9│ import { Alert, Platform } from 'react-native';
     10│ import { withContext, noop } from '@kiwicom/margarita-utils';
     11│


Warning ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ packages/universal-components/src/Checkbox/CheckboxShared.js:80:10

Unused suppression comment.

     77│         onMouseEnter={this.handleOnMouseEnter}
     78│         onMouseLeave={this.handleOnMouseLeave}
     79│       >
     80│         {/* $FlowFixMe - missing declaration for onFocus + onBlußr - react-native issue ( https://github.com/facebook/react-native/pull/21462 ) */}
     81│         <TouchableWithoutFeedback
     82│           accessibilityRole="button"
     83│           onPress={onPress}


Warning ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ packages/universal-components/src/DatePicker/DatePicker.android.js:61:5

Unused suppression comment.

     58│     }
     59│
     60│     const { action, year, month, day } = picker;
     61│     // $FlowFixMe - it has a reference to DatePickerAndroid.ios.js
     62│     if (action !== DatePickerAndroid.dismissedAction) {
     63│       let customDate;
     64│       if (date && !isNaN(date.getTime())) {


[...]

PR #502, opened 2019-03-06
Fix: Updated .flowconfig and fixed flow errors

JoseDuSV:
Summary: Updated `untyped` section in `.flowconfig` (removed unused lines, added few new paths). 

Updated components with flow errors, mostly related to styles.

Closes: #477

### DIFF
--- a/apps/core/components/priceSummary/PriceSummary.js
+++ b/apps/core/components/priceSummary/PriceSummary.js
@@ -17,7 +17,7 @@ type Props = {|
 |};
 type State = {|
   +expanded: boolean,
-  +opacity: any, // @TODO - set more exact type
+  +opacity: Animated.Value,
   +overlayRendered: boolean,
 |};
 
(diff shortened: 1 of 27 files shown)


apps/core/components/priceSummary/PriceSummary.js
```diff
@@ -22,15 +17,15 @@ type Props = {|
 |};
 type State = {|
   +expanded: boolean,
-  +opacity: AnimatedValue,
+  +opacity: any, // @TODO - set more exact type
```
JoseDuSV: @jvaclavik Any idea about this one? Animation related types are not yet exported from react-native (in version we are using) and if I just quickly tried import it directly like below it generates few new errors.
```
import type AnimatedValue from 'react-native/Libraries/Animated/src/nodes/AnimatedValue';
```

jvaclavik: Hmm, it's very hard to find something relevant about flow types for `AnimatedValue`. Your solution looks fine, what kind of errors does it produce?

JoseDuSV: Ok, so I'll keep todo comments for now. And we will update it later, after expo let us upgrade to some higher version which have it exported. 
Mentioned import causes multiple errors related to missing props on AnimatedValue.

RobinCsl: I will commit my solution to this PR.

PR #503, opened 2019-03-07
Feat: Added PickerButton component

JoseDuSV:
Summary: Separated button component from `Picker` to make it reusable.

Note: This component will be reused for example in new `DateInput`.

### DIFF
--- a/packages/universal-components/src/Picker/Picker.android.js
+++ b/packages/universal-components/src/Picker/Picker.android.js
@@ -4,13 +4,12 @@ import * as React from 'react';
 import { View } from 'react-native';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 
-import { Text } from '../Text';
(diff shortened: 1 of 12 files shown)


jvaclavik:
> Summary: Separated button component from `Picker` to make it reusable.
> 
> Note: This component will be reused for example in new `DateInput`.
> I'm not 100% sure about name of this component and I'm definitely open to any other suggestions.
> But basically it's button that looks and partially behaves (for example displays placeholder if value is not set) like input.

Hmm name is not perfect, I had to open storybook to see how this component actually looks like. I tried to think about the name and maybe `PickerButton` would be better. Because `DateInput` is still some kind of "picker" and it's not Input, not Touchable, not Select, just `Button`, so it makes sense for me.

JoseDuSV:
@jvaclavik Thanks a lot. `PickerButton` suits definitely better.
Renamed.

PR #505, opened 2019-03-07
 the locationsByTerm query with Errors

FilipMessa:
- created two scheme for Errors, `HttpResponseError` and `HttpTimeoutError`
- refactor locationsByTerm output type into UnionType
- fix FE

Questions: 

Is OK that the `catchDataloaderError` helper is too coupled with `@kiwicom/fetch` ?
Don't be just `HttpError` enought with some type field (TimeoutError | ResponseError) instead of two Error types?


RobinCsl:
I guess apps/graphql/src/apps/common/types/outputs/errors/HttpResponseError.js and apps/graphql/src/apps/common/types/outputs/errors/HttpTimeoutError.js could implement the same interface HttpError since they have exactly the same fields, but maybe it's too pedantic.
That means you could just query with one fragment on `HttpError`.

Since you are forwarding the error messages from `@kiwicom/fetch`, I wonder if that would be giving too much information in the error messages on FE? Taking care of these error messages in the GraphQL proxy might also reduce the coupling to `@kiwicom/fetch`.

apps/core/components/PlacePicker/PlacePickerContent.js
```diff
@@ -158,12 +158,14 @@ export default createRefetchContainer(
           input: { type: "LocationsByTermInput!", defaultValue: { term: "" } }
         ) {
         locationsByTerm(input: $input) {
-          edges {
-            node {
-              id
-              name
-              locationId
-              type
+          ... on LocationConnection {
```
RobinCsl: So currently, you are not using the errors you just provided? Will that be part of another PR?

FilipMessa: yes this is just about the GraphQL part right now

FilipMessa:
>  guess apps/graphql/src/apps/common/types/outputs/errors/HttpResponseError.js and apps/graphql/src/apps/common/types/outputs/errors/HttpTimeoutError.js could implement the same interface HttpError since they have exactly the same fields,

I was thinking about the interface but I want to keep this simple as possible and get feedback at first. 

> Since you are forwarding the error messages from @kiwicom/fetch, I wonder if that would be giving too much information in the error messages on FE? 

I just take forwardin statusText and statusCode into the Errors schemas, or you mean `throw err` in the `catchDataloaderError` if something diffrent will happen?

This probably can send you some "private" information, Idea for this is don't throw these kinds of error in the future, but log them to the sentry.

JoseDuSV:
When I think about it I'm slightly more for having just one `HttpError`. Basically because those two errors cannot happen at the same time, so I think that having just one error and handle it on FE by code can be enough.

FilipMessa:
So on yesterday when I couldn't sleep, I realized that my thought about two kinds of errors was not right, I missed an important point that the TimeouError can happen only on GraphQL server side when we call TequilaAPI, and the client's timeout error can happen when we calling the GraphQL server from the client, which is a different kind of error.

So Base on this I decided to use just one HttpError, and if TimeoutError will be returned I send to client HttpError with code: 503 and message: Service Timeout

PR #508, opened 2019-03-08
Feat: PlacePicker delete slected location by backspace key

FilipMessa:
User can delete selected location by backspace key

### DIFF
--- a/apps/core/scenes/search/PlacePicker/PlacePickerContent.js
+++ b/apps/core/scenes/search/PlacePicker/PlacePickerContent.js
@@ -21,21 +21,27 @@ import {
 import type { PlacePickerContent_locations as PlacePickerContentType } from './__generated__/PlacePickerContent_locations.graphql';
 import {
   mapOptionToLocation,
-  mapLocationTypeToOptionType,
+  convertLocationTypeToOptionType,
   filterOptions,
   type OptionType,
 } from './helpers';
 
 type Props = {|
+  +pickerType: 'travelFrom' | 'travelTo',
   +locations: ?PlacePickerContentType,
   +relay: RelayRefetchProp,
   +onChangeText: string => void,
   +travelFrom: ?Array<Location>,
   +travelTo: ?Array<Location>,
-  +modalType: 'DESTINATION' | 'ORIGIN' | 'HIDDEN',
   +clearLocation: (type: 'travelTo' | 'travelFrom') => void,
-  +setLocation: (type: LocationSearchType, location: Location) => void,
-  +addLocation: (type: LocationSearchType, location: Location) => void,
+  +setLocation: (
+    type: LocationSearchType,
+    location: Location | Location[],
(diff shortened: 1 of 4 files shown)


apps/core/scenes/search/PlacePicker/PlacePickerContent.js
```diff
@@ -61,27 +66,28 @@ function mapLocationsToOption(locations: ?Array<Location>) {
 }
 
 class PlacePickerContent extends React.Component<Props, State> {
-  state = {
-    searchText: '',
-  };
+  constructor(props: Props) {
```
jvaclavik: Is there any difference to set state in constructor?

RobinCsl: There is a difference only when you use props. It seems like props aren't used here, so the previous state of the code should be used.

apps/core/scenes/search/PlacePicker/PlacePickerContent.js
```diff
@@ -158,12 +166,24 @@ class PlacePickerContent extends React.Component<Props, State> {
           onChangeText={this.handleChangeText}
           options={filterOptions(options, selectedOptions)}
           selected={selectedOptions}
+          onKeyPress={this.handleBackspacePress}
```
RobinCsl: `onKeyPress` sounds like this would trigger on any key press.  So when I press `a`, `onKeyPress` does nothing but somehow the text changes.  I think the logic becomes confusing with `onChangeText` too. Could we improve the API?

apps/core/scenes/search/PlacePicker/PlacePickerContent.js
FilipMessa: I will change it, I was experimenting with state and didn't change it back

apps/core/scenes/search/PlacePicker/PlacePickerContent.js
FilipMessa: OptionPicker's onKeyPress callback trigger when you press whatever key, all logic for filtering keys is in the handleBackspacePress.

apps/core/scenes/search/PlacePicker/PlacePickerContent.js

apps/core/scenes/search/PlacePicker/PlacePickerContent.js
JoseDuSV: I'm quite ok with current solution. But maybe we want to somehow extend `onKeyPress` later and for example add `enter` to confirm/close input or arrows to jump between places (if I want to for example delete just first place from three).

So what about adding something like `handleKeyPress`. And handle different key cases inside (for now just backspace)?

FilipMessa: @JoseDuSV sure, why not?  I will change it as you suggest.

PR #510, opened 2019-03-08
Improve headline responsiveness on big tablet

KarinaDavtyan:
+ fix horizontal scroll
+ wrap buttons in the *Orbit's* `Stack`
+ improve responsiveness on *big tablet* size

#### Before
screenshot 2019-03-07 at 17 48 10

#### After
screenshot 2019-03-08 at 16 31 03


apps/landingPage/components/Header.js
```diff
@@ -67,6 +67,10 @@ const GithubLink = styled.div`
   }
 `;
 
+const LinkWithHeight = styled.div`
+  height: 42px;
```
jvaclavik: I was just thinking. What about use 44px? We could get it from `defaultTokens` (heightButtonNormal) and actually 44pt (~px) is some kind of factory standard for minimal touchable size for tablets and mobile devices (not for PC).

Apple also recommends it:
screenshot 2019-03-08 at 17 03 28

https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/adaptivity-and-layout/

apps/landingPage/components/Headline/GetStartedInfo.js
```diff
@@ -64,28 +84,11 @@ const ContainerLeftText = styled.div`
...
-    flex-direction: row;
+    width: 75vw;
   }
-`;
-
-const WrapperRightButton = styled.div`
-  margin-top: 10px;
-  @media (min-width: ${BREAKPOINTS.TABLET}px) {
-    margin: 0 0 0 1vw;
+  @media (min-width: 1025px) {
```
jvaclavik: What about use some constant here? Btw 1025 is quite interesting number, but I understand that you want style only for bigger screens than 1024px 😄

KarinaDavtyan: Hmm I really should stop taking numbers from my head 😅
I ve created `BIG_TABLET` media query breakpoint which I ll change to 1024 now
And probably missed this one for changing it to constant

apps/landingPage/components/Header.js
KarinaDavtyan: Thats a good point but I m not sure in this case.
I put height here only because somehow wrapping `img` into `a` was increasing height of the `a` more than the `img` height so I couldnt align logo properly in the mobile header

But still, as for the 44px number, I am not sure what was the height intended in design so I just took 42px number by eye

FilipMessa: I think that 42px is OK because the width is much more. Maby and some point we will focus on accessibility and improve it.

apps/landingPage/components/Headline/GetStartedInfo.js
```diff
@@ -45,16 +58,23 @@ const ContainerLeftText = styled.div`
...
   align-items: flex-start;
-
   @media (min-width: ${BREAKPOINTS.BIG_MOBILE}px) and (max-width: ${BREAKPOINTS.TABLET}px) {
     align-items: center;
   }
+  @media (min-width: ${BREAKPOINTS.BIG_MOBILE}px) and (max-width: ${BREAKPOINTS.BIG_TABLET}px) {
+    align-items: center;
+  }
   @media (min-width: ${BREAKPOINTS.TABLET}px) {
+    flex-direction: 'row';
```
FilipMessa:
```suggestion
    flex-direction: row;
```

But probably it can be removed

apps/landingPage/components/Header.js
KarinaDavtyan: Alright, keeping 42px for now then

FilipMessa:
No layout looks broken on DESKTOP (green color), and I have horizontal scrollbar down(orange color)

Screen Shot 2019-03-13 at 15 41 44

KarinaDavtyan:
> No layout looks broken on DESKTOP (green color), and I have horizontal scrollbar down(orange color)
> 
> Screen Shot 2019-03-13 at 15 41 44

No way 😨Will be looking in it now

PR #511, opened 2019-03-08
Delete useless styles from MenuGroup

jvaclavik:
All `MenuGroup` occurrences had the same style

### DIFF
--- a/packages/universal-components/src/MenuGroup/MenuGroup.js
+++ b/packages/universal-components/src/MenuGroup/MenuGroup.js
@@ -40,6 +40,9 @@ export default function MenuGroup({
 
 const styles = StyleSheet.create({
   card: {
+    android: {
+      marginTop: parseFloat(defaultTokens.spaceSmall),
+    },
     padding: 0,
     ios: {
       borderTopWidth: 0.5,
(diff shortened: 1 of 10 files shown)


packages/universal-components/src/MenuGroup/MenuGroup.js
```diff
@@ -39,6 +41,10 @@ export default function MenuGroup({
 }
 
 const styles = StyleSheet.create({
+  title: {
```
JoseDuSV: `MenuGroupTitle` is used only inside `MenuGroup` so I think we can also move this style directly inside `MenuGroupTitle`.

jvaclavik: Done

packages/universal-components/src/MenuGroup/MenuGroupTitle.js
```diff
@@ -36,13 +36,14 @@ const styles = StyleSheet.create({
...
       marginStart: 16,
       marginBottom: 8,
-      textTransform: 'uppercase',
       lineHeight: 17,
     },
     lineHeight: 19,
     marginStart: 13,
     marginTop: 8,
     marginBottom: 5,
+    fontWeight: 'bold',
```
RobinCsl: I believe `bold` should live on Text as a prop, what do you think?

Issue #72, opened 2018-12-17
Booking screen: Connection component


MRTNKCR:
The connection (or flight - but we do also offer rides now :-) ) component offers a basic overview of the trip.

There would be no photo and no Share button.

<img width="533" alt="snimek obrazovky 2018-12-17 v 11 32 42" src="https://user-images.githubusercontent.com/7661448/50082636-bb24ad80-01f1-11e9-8792-5a441f0fdd70.png">

When open, the component should display the trip's details - this can be recycled from the Trip detail screen.

![image from ios](https://user-images.githubusercontent.com/7661448/50082685-d68fb880-01f1-11e9-93de-5b66a1e3ded0.png)

Design in Figma: https://www.figma.com/file/OUhBJ9rPaTbRljptC0lrRrUF/Booking?node-id=5%3A439

FilipMessa:
done

PR #519, opened 2019-03-12
Feat: Added CheckItinerary query

JoseDuSV:
Summary: Added `ItineraryValidity` to schema and sample query implementation to `ResultDetail`.

Note: Itinerary values mapping will be added in next PR. `ResultDetailItinerary` query implementation will be also moved directly into `ResultDetail` (where its data will be used) in different PR.

**IMPORTANT**: To be able to use Booking API `.env/API_KEY` needs to be changed to new one.
It can be found in Tequila dashboard `My applications` -> `searchandbook`


apps/core/scenes/resultDetail/ResultDetailItinerary.js
```diff
@@ -0,0 +1,106 @@
...
+      </>
+    );
+  }
+}
+
+export default createRefetchContainer(
+  ResultDetailItinerary,
+  {
+    data: graphql`
+      fragment ResultDetailItinerary_data on RootQuery
```
jvaclavik: Maybe we could name it more specifically, not just `data`

apps/graphql/src/apps/itinerary/Itinerary.js
```diff
@@ -111,3 +127,13 @@ export type ApiResponseType = {|
     +booking_token: string,
   |}>,
 |};
+
+export type ValidityApiResponseType = {|
+  +flights_checked: boolean,
```
jvaclavik: I think all data from API should be optional:
```
+flights_checked?: boolean,
```

JoseDuSV: True, updated.

apps/core/scenes/resultDetail/ResultDetailItinerary.js
JoseDuSV: I've been thinking about it. And `data` seems to me most universal. Because it will contain mixed content like validity and itinerary data.

apps/core/scenes/resultDetail/ResultDetail.js
```diff
@@ -94,6 +95,7 @@ class ResultDetail extends React.Component<Props, State> {
     return (
       <View style={styles.container}>
         <ScrollView>
+          <ResultDetailItineraryRenderer bookingToken={this.props.detailId} />
```
FilipMessa: this naming is confusing the one place we call this `detailId` an on another is `bookingToken`

apps/graphql/src/apps/itinerary/types/inputs/ItineraryCheckInput.js
```diff
@@ -0,0 +1,28 @@
...
+  GraphQLString,
+  GraphQLInt,
+} from 'graphql';
+
+import PassengersInput from './PassengersInput';
+
+export default new GraphQLInputObjectType({
+  name: 'ItineraryCheckInput',
+  fields: {
+    bookingToken: {
```
FilipMessa: I can find any bookingToken field in BookingInterface Schema, so I a not sure if this is the right name. Where I can find bookingToken in the graphQL Schema?

apps/graphql/src/apps/itinerary/types/outputs/ItineraryValidity.js
```diff
@@ -0,0 +1,21 @@
...
+    itineraryChecked: {
+      type: GraphQLBoolean,
+      description: 'All segments successfully checked',
+    },
+    itineraryValid: {
+      type: GraphQLBoolean,
+      description:
+        'Itinerary is valid (exists and is not sold-out or cancelled)',
+    },
+    itinerary: { type: GraphQLItinerary, description: 'Actual itinerary data' },
```
FilipMessa: What is your motivation to return also itinerary data here?

apps/graphql/src/apps/itinerary/dataloaders/ItineraryValidity.js
```diff
@@ -0,0 +1,88 @@
...
+    destination: null,
+    sectors: null,
+    price: {
+      currency: 'EUR',
+      amount: response.total,
+    },
+  };
+};
+
+const mapItineraryId = flights => {
```
FilipMessa: Be more generic in naming because we must expect that TequilaAPI in future won't be only about flights.

apps/core/scenes/resultDetail/ResultDetailItinerary.js
```diff
@@ -0,0 +1,106 @@
...
+
+import type { ResultDetailItinerary_data as ResultDetailItineraryType } from './__generated__/ResultDetailItinerary_data.graphql';
+
+type Props = {|
+  +data: ?ResultDetailItineraryType,
+  +relay: RelayRefetchProp,
+|};
+
+const REFETCH_INTERVAL = 5 * 1000;
+const REFETCH_TIMEOUT = 3 * 60 * 1000;
```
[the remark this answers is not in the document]
FilipMessa: And I am not sure but what about use polling? https://github.com/kiwicom/relay-example/blob/master/pages/Polling.js

FilipMessa:
I have this error in console. 
Screenshot 2019-03-13 at 10 44 41

Did you think about that, this itinerary validation as a part of Itinerary schema instead of standalone query?  I just considering if check this can't be part of Itinerary schema.

FilipMessa:
And this PR looks big, not necessary big I will be preferred in future split it on FE and BE part.

JoseDuSV:
> I have this error in console.
> Screen Shot 2019-03-13 at 13 33 27
> 
> Did you think about that, this itinerary validation as a part of Itinerary schema instead of standalone query? I just considering if check this can't be part of Itinerary schema.

You should probably need to update your `API_KEY` to get rid of the error.
Thanks for pointing it out, I've also updated PR description to contain this info.

I've been thinking about making validation info part of the itinerary. It will make sense in current state. But I'm slightly worried that it can cause problems when we need to extend this `check` response.
For example I can imagine that `price_change` can be also useful for us, to display some kind notification to user. But it will not make much sense as part of the itinerary schema.

apps/core/scenes/resultDetail/ResultDetailItinerary.js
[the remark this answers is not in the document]
JoseDuSV: Comment added. 
I think that these values are currently quite strongly connected to this component and I don't think we will use refetching with the same configuration anywhere else, which will justify moving it into `@kiwicom/margarita-config`.

Thanks for the Polling example. I think that refetch is slightly more suitable for this situation.
Basically there is advantage of more control. Currently we don't need to run check in some continuous loop. We just wait until check is valid and then do recheck/refetch only when something important like passengers count will change.

apps/core/scenes/resultDetail/ResultDetail.js
JoseDuSV: Good point. Changed to `bookingToken` (from schema) everywhere.

apps/graphql/src/apps/itinerary/types/inputs/ItineraryCheckInput.js
JoseDuSV: It's not on Booking but on Itinerary. The name makes sense to me, because it serves for itinerary booking. But we can definitely change it if you have something better in mind.

apps/graphql/src/apps/itinerary/types/outputs/ItineraryValidity.js
JoseDuSV: Because we need it to fill the itinerary detail in header https://github.com/kiwicom/margarita/issues/72
It will probably not be necessary for mobile, where you can't reach summary screen without visiting search first (and getting the data).
But on web, you can refresh the page or go directly to `/result-detail` from some shared link.

JoseDuSV:
> And this PR looks big, not necessary big I will be preferred in future split it on FE and BE part.

Thanks for pointing it out. True that I can save few rows by keeping FE for next PR. But I was worried that then it will be too abstract and I wanted to show at least minimal implementation and usage.

apps/graphql/src/apps/itinerary/types/inputs/ItineraryCheckInput.js
FilipMessa: make me more sense now thanks

apps/core/scenes/resultDetail/ResultDetailItinerary.js
jvaclavik: OK, let's keep it. I also couldn't find better name

apps/graphql/src/apps/itinerary/types/outputs/ItineraryValidity.js
FilipMessa: I understand your point but I do not have a good feeling about designing schema like that, I need more think about. But maybe it is OK, let's ask for more opinion.

@michalsanger What do you think about this query?

JoseDuSV: @FilipMessa True, I also don't have best feeling about this. In perfect world I can for example imagine to have separate query just to get itinerary data by id, but for now there is no entry point for something like this in Tequila.

FilipMessa:
> You should probably need to update your API_KEY to get rid of the error.
@JoseDuSV so before the merge, we will need update lambda env also

JoseDuSV:
> > You should probably need to update your API_KEY to get rid of the error.
> 
> @JoseDuSV so before the merge, we will need update lambda env also

@FilipMessa True, I don't know if you have the access, otherwise we will need to nicely ask @RobinCsl

RobinCsl:
Guys, please remove the screenshots with a booking token in clear. This is probably not critical, but be mindful that GitHub is a public space.
Please, give me more information by DM on Slack as to which API key needs to be changed.

apps/graphql/src/apps/itinerary/types/outputs/ItineraryValidity.js
JoseDuSV: @FilipMessa Refactored, `valid` and `checked` is now directly part of the Itinerary.
Our Tequila key is working again so I was finally able to properly test it.

apps/graphql/src/apps/itinerary/queries/__tests__/__fixtures__/checkItinerary/CheckItinerary.graphql
```diff
@@ -0,0 +1,33 @@
...
+    local
+  }
+}
+
+{
+  checkItinerary(input: { bookingToken: "token", bags: 1, passengers: { adults: 1 }}) {
+    id
+    valid
+    checked
+    sectors {
```
michalsanger: Why do you include `sectors` in this query?

apps/graphql/src/apps/itinerary/types/outputs/Itinerary.js
```diff
@@ -37,6 +38,15 @@ export default new GraphQLObjectType({
     startTime: { type: GraphQLDateType },
     type: { type: GraphQLString },
     bookingToken: { type: GraphQLString },
+    checked: {
```
michalsanger: Could you name it `isChecked` and `isValid` because it's a boolean prop?

michalsanger:
Is there a need for this query? I mean if there is [`node` query](https://facebook.github.io/relay/graphql/objectidentification.htm) you can fetch itinerary by the ID and get whatever data you want.

apps/graphql/src/apps/itinerary/queries/__tests__/__fixtures__/checkItinerary/CheckItinerary.graphql
JoseDuSV: Same test query like for `searchItineraries` with sectors is used because we expect basically same data, just with extra isChecked and isValid. Because we are also displaying whole itinerary info on the result/booking detail page.

apps/graphql/src/apps/itinerary/types/outputs/Itinerary.js
JoseDuSV: Sure, remaned.

JoseDuSV:
> Is there a need for this query? I mean if there is [`node` query](https://facebook.github.io/relay/graphql/objectidentification.htm) you can fetch itinerary by the ID and get whatever data you want.

I think we need it. Because we are not just checking if itinerary itself is valid. But it is also checking if itinerary is available for given number (and types) of passengers and bags.
So we need to re-check it every time user changes the number of passengers on booking screen.

apps/graphql/src/apps/itinerary/queries/__tests__/__fixtures__/checkItinerary/CheckItinerary.graphql
michalsanger: I just see it's resolved as `[]` in the snapshot so it seems useless here.

JoseDuSV: Data will be mapped to sectors in separate PR. But you are right, it can be confusing to have it there now.
So I removed sectors and will update it once there will be some reasonable content for snapshots.

apps/graphql/src/apps/itinerary/queries/__tests__/__snapshots__/CheckItineraryQueries.test.js.snap
```diff
@@ -0,0 +1,35 @@
...
+    isChecked
+  }
+}
+
+
+~~~~~~~~~~ OUTPUT ~~~~~~~~~~
+{
+  "errors": [
+    {
+      "message": "Fragment \\"RouteStopFragment\\" is never used.",
```
michalsanger: You should fix the query, IMHO this error is not expected

JoseDuSV: 🤦‍♂️ Sorry, didn't checked new snapshot. Fixed, updated.

apps/graphql/src/apps/itinerary/dataloaders/Itinerary.js
```diff
@@ -0,0 +1,76 @@
...
+    bookingToken: response.booking_token,
+    isValid: !response.flights_invalid,
+    isChecked: response.flights_checked,
+    startTime: null,
+    endTime: null,
+    origin: null,
+    destination: null,
+    sectors: null,
+    price: {
+      currency: 'EUR',
```
FilipMessa: pls add TODO comment also here

FilipMessa:
PLS with merge we need update the API key on lambda

RobinCsl:
@FilipMessa 
> PLS with merge we need update the API key on lambda

Again? I already changed an API key this week I believe so maybe it's already set up to work!

FilipMessa:
@RobinCsl oh, in that case, it should work I didn't know that you change API key already so sorry.

apps/graphql/src/apps/itinerary/dataloaders/Itinerary.js
JoseDuSV: Right, updated. The price part on this endpoint is slightly confusing, because it always returns total price just in EUR. I have few other notes from fiddling with API, so I'll add them to our Tequila related issue after I clean them a bit.

FilipMessa: thanks

Issue #121, opened 2019-01-15
Component:Create the PassengerForm  (Booking detail)


FilipMessa:
Create the PassengerForm component it is part of the booking detail flow #73 
design: https://www.figma.com/file/OUhBJ9rPaTbRljptC0lrRrUF/Multistep?node-id=648%3A65

![passengerform](https://user-images.githubusercontent.com/6354326/51162078-3468fc00-1895-11e9-8c55-9803683278db.jpg)

PR #523, opened 2019-03-12
Feat: Added PassengerForm inputs

JoseDuSV:
Summary: Added simple `DateInput` component. Which is basically just composite of `PickerButton` and `DatePicker`.

Added passenger related inputs to `PassengerForm` screen.

Closes: #121

Screenshot 2019-03-12 at 15 38 46

### DIFF
--- a/apps/core/scenes/passengerForm/PassengerForm.js
+++ b/apps/core/scenes/passengerForm/PassengerForm.js
@@ -2,22 +2,140 @@
 
 import * as React from 'react';
 import { View, ScrollView } from 'react-native';
-import { StyleSheet, Text } from '@kiwicom/universal-components';
+import {
+  StyleSheet,
+  SegmentedButton,
+  Picker,
+  TextInput,
+} from '@kiwicom/universal-components';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import { DateInput } from '@kiwicom/margarita-components';
 
-export default function PassengerForm() {
-  return (
-    <View style={styles.container}>
-      <ScrollView>
-        <Text>TODO: Passenger form</Text>
+type Props = {||};
+
+type State = {|
+  gender: ?string,
+  givenName: ?string,
+  lastName: ?string,
+  nationality: ?string,
+  birthDate: ?Date,
+|};
+
+const genderData = [
+  { label: 'Male', value: 'male' },
+  { label: 'Female', value: 'female' },
+];
+
+const nationalityData = [
+  {
+    label: 'Czechia',
+    value: 'CZ',
+  },
+  {
+    label: 'Croatia',
+    value: 'HR',
+  },
+];
+
+export default class PassengerForm extends React.Component<Props, State> {
+  state = {
+    gender: null,
+    givenName: null,
+    lastName: null,
+    nationality: null,
+    birthDate: null,
+  };
+
+  handleGenderChange = (gender: ?string) => {
+    this.setState({ gender });
+  };
+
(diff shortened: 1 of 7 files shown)


apps/core/scenes/passengerForm/PassengerForm.js
```diff
@@ -2,22 +2,140 @@
...
+  containerContent: {
+    flex: 1,
+    alignItems: 'center',
+  },
+  form: {
+    padding: parseInt(defaultTokens.spaceMedium, 10),
+    backgroundColor: defaultTokens.paletteWhite,
+    width: '100%',
+    web: {
+      maxWidth: 720,
```
jvaclavik: Btw could we create this value as some importable constant? We need to use it on more pages so we wouldn't have to repeat it.

packages/universal-components/src/SegmentedButton/SegmentedButton.js
```diff
@@ -18,11 +18,11 @@ export type Props = {|
   /**
    * Selected segment value
    */
-  +selectedValue: string,
+  +selectedValue: ?string,
   /**
    * Dispatched on segment press
    */
-  +onValueChange: (value: string) => void,
+  +onValueChange: (value: ?string) => void,
```
jvaclavik: You don't need `value` here, you can use just this: 
```
+onValueChange: ?string => void,
```

JoseDuSV: Right, updated.

apps/core/scenes/passengerForm/PassengerForm.js
JoseDuSV: Good point, this one is related to other repeating style values which should be added to tokens.
So for now I just added token related todo comment.

I think in next PR we should at least create something like `design-tokens` inside `packages` folder and store those repeating values there to stop flooding code with todos in repeating cases like this.

FilipMessa:
I think that the right place for the `DateInput` should be `/packages/components/` because from my point of the view the universal-component should be atomic. 
And when I checked this component in the storybook, I suppose that this component composition is more Margarita specific than generic.
I don't know It is for discussion.

cc @RobinCsl What is your opinion?

RobinCsl:
@FilipMessa I do not have a strong opinion, but let's say that it could live in `universal-components` if it was both a controlled and non-controlled (I think it's called optionally controlled) component, but in the present state it should probably be in `margarita-components` since it's handling its own logic.
The line is blurry here.

FilipMessa:
I also don't have a strong opinion about, as you said the line is blurry, my point was just about that this component can be easily created as the composition from an existing component. 

But as I said before, I 👽 am not sure about the best approach so @JoseDuSV You must decide 😉

JoseDuSV:
> I also don't have a strong opinion about, as you said the line is blurry, my point was just about that this component can be easily created as the composition from an existing component.
> 
> But as I said before, I 👽 am not sure about the best approach so @JoseDuSV You must decide 😉

It's quite between both. But you are right that as it is just composite, it makes more sense to keep it just inside margarita components.

Moved 🚀

apps/core/scenes/passengerForm/PassengerForm.js
KarinaDavtyan: Good idea. It would be nice to move breakpoints from LandingPage `mediaQueriesConfig.js` to that `design-tokens` package as well maybe

jvaclavik: @JoseDuSV Could you please create issue for it?

PR #531, opened 2019-03-14
Implement PaymentForm

jvaclavik:

### DIFF
--- a/apps/core/scenes/payment/Payment.js
+++ b/apps/core/scenes/payment/Payment.js
@@ -3,7 +3,13 @@
 import * as React from 'react';
 import { View, ScrollView } from 'react-native';
 import { StyleSheet } from '@kiwicom/universal-components';
-import { TableRow, TableRowDivider } from '@kiwicom/margarita-components';
+import {
+  TableRow,
+  TableRowDivider,
+  CreditCardPaymentForm,
+  ExtendedTouchable,
+  Text,
+} from '@kiwicom/margarita-components';
 import { withNavigation, type Navigation } from '@kiwicom/margarita-navigation';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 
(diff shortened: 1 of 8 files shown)


packages/universal-components/src/FormFeedback/FormFeedbackTypes.js
```diff
@@ -2,7 +2,7 @@
 
 import * as React from 'react';
 
-type Type = 'help' | 'error';
+type Type = 'description' | 'error';
```
JoseDuSV: Is this update necessary for payment form or is there any other reason for this change? For me help or hint is quite ok (I can imagine it will be used for something like `Min length 12 characters`).
Description is also ok, but slightly less clear for me. I see it on similar level as label or placeholder.

packages/universal-components/src/TextInput/TextInputTypes.js
```diff
@@ -21,9 +21,10 @@ export type Props = {|
   +minLength?: number, // this prop is supported only on web
   +onFocus?: () => void | Promise<any>,
   +onBlur?: () => void | Promise<any>,
+  +onHelpPress?: () => void,
   +onChangeText?: string => void | Promise<any>,
-  +error?: React.Node,
-  +help?: React.Node,
+  +error?: React.Node, // @TODO rename to renderError etc.?
```
JoseDuSV: I think that just `error` is ok. It is similar to description/help or even label, prefix and suffix.

apps/core/scenes/payment/Payment.js
```diff
@@ -57,6 +154,18 @@ const styles = StyleSheet.create({
     flex: 1,
     backgroundColor: defaultTokens.backgroundBody,
   },
+  centeredContainer: {
+    web: {
+      alignSelf: 'center',
+      maxWidth: 720, // @TODO repeating value, should be added to design-tokens
+      textAlign: 'left',
```
JoseDuSV: This one causes ```Invalid props.style key `textAlign` supplied to `View`.``` warning.

packages/components/src/creditCardPaymentForm/CreditCardPaymentForm.js
```diff
@@ -0,0 +1,148 @@
...
+}
+
+const styles = StyleSheet.create({
+  card: {
+    web: {
+      alignItems: 'center',
+    },
+    flex: 1,
+  },
+  form: {
```
JoseDuSV: Form on web have some variable width (instead of 720) and thanks to that it slightly jumps on input focus

input

Setting ```width: '100%'``` can probably help, like in `PassengerForm`.

I seems to be a repeating pattern so we can probably make something like `FormWrapper` component from it. Which will be just white View on mobile and Card on web. 
To get styling more closer to current kiwi.com website.

Screenshot 2019-03-15 at 14 37 57

apps/core/scenes/payment/Payment.js
```diff
@@ -13,15 +19,107 @@ type Props = {|
...
+            creditCardNumber={creditCardNumber}
+            isCreditCardSaved={isCreditCardSaved}
+            monthOfExpiry={monthOfExpiry}
+            onChangeCardholdersName={this.handleCardholdersNameChange}
+            onChangeCreditCardNumber={this.handleCreditCardNumberChange}
+            onChangeCreditCardSaveSwitch={this.handleChangeCreditCardSave}
+            onChangeMonthOfExpiry={this.handleMonthOfExpiryChange}
+            onChangeSecurityCode={this.handleSecurityCodeChange}
+            onChangeYearOfExpiry={this.handleYearOfExpiryChange}
+            onRedeemPromoCodePress={this.handleRedeemPromoCodePress}
```
JoseDuSV: This one seems to be unused.

packages/universal-components/src/FormFeedback/FormFeedbackTypes.js
jvaclavik: Using help for text below the `TextInput` could be confusing if you want to use icon with help:
 
Screenshot 2019-03-15 at 14 48 52

Do you have idea how to rename it? I didn't want to use something like `rightIcon`, because it's not clear what is the reason why icon is clickable.

packages/universal-components/src/TextInput/TextInputTypes.js
jvaclavik: OK

packages/universal-components/src/FormFeedback/FormFeedbackTypes.js
JoseDuSV: Oh, I see. Thanks. 
I think we can already use `suffix` for something like this, like in Orbit example: https://kiwicom.github.io/orbit-components/?selectedKind=InputField&selectedStory=With%20ButtonLink%20suffix&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

jvaclavik: Good idea!

jvaclavik: Done

packages/universal-components/src/TextInput/TextInputTypes.js
jvaclavik: Comment removed

apps/core/scenes/payment/Payment.js
jvaclavik: Done

packages/components/src/creditCardPaymentForm/CreditCardPaymentForm.js
jvaclavik: Done

apps/core/scenes/payment/Payment.js
jvaclavik: Deleted

Issue #528, opened 2019-03-13
Can't set Button width on Android


jvaclavik:
**Describe the bug**
It's not possible to set specific width for `Button` on Android.

**Proposed solution**
Wrapper inside of Button should help:
```
<View style={{ width: this.props.width }}>
  // rest of Button
</View>
```

JoseDuSV:
Closed by #533

PR #533, opened 2019-03-15
Fix button width on Android

jvaclavik:
#528

### DIFF
--- a/packages/universal-components/src/Touchable/Touchable.js
+++ b/packages/universal-components/src/Touchable/Touchable.js
@@ -69,7 +69,7 @@ export default class Touchable extends React.Component<Props> {
       testID,
       width,
       disabled,
-      ...accessiblityProps
+      ...accessibilityProps
     } = this.props;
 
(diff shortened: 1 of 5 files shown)


packages/universal-components/src/Touchable/Touchable.js
```diff
@@ -103,28 +103,27 @@ export default class Touchable extends React.Component<Props> {
...
-          style={{ width }}
-          useForeground={useForeground}
-          background={TouchableNativeFeedback.Ripple(
-            rippleColor,
-            borderlessRipple,
-          )}
-        >
-          <View style={style}>{children}</View>
-        </TouchableNativeFeedback>
+        <View style={{ width }}>
```
JoseDuSV: Do we need this extra wrapper? 
I think that updating View inside touchable can be enough.
`<View style={[style, { width }]}>{children}</View>`

jvaclavik: True :) done

PR #534, opened 2019-03-15
Feat: Added margarita-design-tokens package

### LINKED ISSUE #136: Move repeating style values to separate package

JoseDuSV:
Summary: Added `@kiwicom/margarita-design-tokens` to store repeating style values which are currently missing in `orbit-design-tokens`.
Cleaned code from related todos.

Note: I've tried to use as generic names as possible where it makes sense. And kept original names if we just replacing actual value from orbit, like in case of badges.

Closes #136

### DIFF
--- a/apps/core/components/ItineraryCard/FlightTimes.js
+++ b/apps/core/components/ItineraryCard/FlightTimes.js
@@ -3,6 +3,7 @@
 import * as React from 'react';
 import { View, Platform } from 'react-native';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import { margaritaTokens } from '@kiwicom/margarita-design-tokens';
 import { StyleSheet, Text } from '@kiwicom/universal-components';
 import { graphql, createFragmentContainer } from '@kiwicom/margarita-relay';
 
@@ -35,7 +36,7 @@ const styles = StyleSheet.create({
     padding: 5,
     web: {
       fontWeight: '500',
-      color: '#2e353b', // @TODO repeating value - should be added to design-tokens
+      color: margaritaTokens.paletteBlueDark,
       padding: 0,
       lineHeight: 17,
     },
--- a/packages/design-tokens/src/MargaritaTokens.js
+++ b/packages/design-tokens/src/MargaritaTokens.js
@@ -1,0 +1,16 @@
+// @flow
+
+export const margaritaTokens = {
+  widthScreenNormal: 720,
+  heightStatusBar: 20,
+  paletteBlueDark: '#2e353b',
+  borderColorInkLight: '#d2d9e0',
+  borderColorInkNormal: '#57626c',
+  backdropColor: '#171b1e',
+
+  heightBadge: 18,
+  backgroundBadgeWarning: '#F9971E',
+  backgroundBadgeWhite: '#F5F7F9',
+  colorTextBadgeWarning: '#FFFFFF',
+  colorTextBadgeWhite: '#7F91A8',
+};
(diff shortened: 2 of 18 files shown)


apps/core/components/ItineraryCard/FlightTimes.js
```diff
@@ -35,7 +36,7 @@ const styles = StyleSheet.create({
     padding: 5,
     web: {
       fontWeight: '500',
-      color: '#2e353b', // @TODO repeating value - should be added to design-tokens
+      color: margaritaTokens.textColorBlueDark,
```
wEprous: The convention in `orbit-design-tokens` is to declare only `textColor` for some `types`, e.g. `textColorPrimary`. Also, we are using `colorText` not `textColor`, but it's totally up to you.

This doesn't apply for `palette` tokens, so there are two options/ways I suggest:
- declaring `paletteBlueDark` token and using it every needed
- declaring component specific tokens, e.g. `colorTextFlighTime` for this case

The second option is probably better because we will know much easier what differences you have and what's needs to be solved by `orbit-design-tokens`.

packages/design-tokens/src/MargaritaTokens.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+
+export const margaritaTokens = {
+  widthScreenNormal: 720,
+  textColorBlueDark: '#2e353b',
+  borderColorGreyLight: '#d2d9e0',
```
wEprous: In terms of color names, I would stick to orbit's. We don't use `grey`, but `cloud` and `ink` instead. Both of these colors are ink-ish.

apps/core/components/ItineraryCard/FlightTimes.js
```diff
@@ -3,6 +3,7 @@
 import * as React from 'react';
 import { View, Platform } from 'react-native';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import { margaritaTokens } from '@kiwicom/margarita-design-tokens';
```
jvaclavik: btw what about rename `margaritaTokens` export to `defaultTokens` in order to keep the same naming as in Orbit? In ideal case we could just change the import and not every occurence :)

JoseDuSV: Then it will battle orbit tokens import and we will need to use something like
```import { defaultTokens as orbitTokens } from '@kiwicom/orbit-design-tokens'```
which will just reverse the situation.

I've been also thinking about spreading `orbit-design-tokens` directly into `margarita-design-tokens` and using just one import. 
But at the end I like it separate for now. I'm slightly worried that if we merge everything together now it will become wild west 🙂 where we can unintentionally overwrite some values from original tokens which we actually need, thanks to the similar naming.
And it can make using token values slightly more difficult. Because for example before using some color, you'll check `orbit-tokens` and even if you find it there, you'll also need to check new tokens if somebody didn't overwritten it.
But with approach in this PR, you'll still just check orbit tokens and if it's not there you'll add it to margarita tokens (if you don't already find it there).

jvaclavik: You are right! I didn't realise that

apps/core/components/ItineraryCard/FlightTimes.js
JoseDuSV: Really thanks for the feedback. 
For now I just used `paletteBlueDark` because we want to mainly collect generic repeating style values for now.
But we will definitely sort them to something more specific in some next iteration. Probably like suggested `colorTextFlighTime: paletteBlueDark` as it is defined in Orbit.

packages/design-tokens/src/MargaritaTokens.js
JoseDuSV: Thanks, updated.

PR #553, opened 2019-03-18
Update to date-fns@v2 and unify dateFormats

jvaclavik:
- Updated DateFNS to version 2. (I will need some functions from `date-fns@v2` in new `RangeDatePicker`). Btw there are some very interesting breaking changes 😂 
- Add all date formats to `margarita-config`. It was on many places in the code and it was quite messy.

### DIFF
--- a/apps/core/components/ItineraryCard/TripSectorHelpers.js
+++ b/apps/core/components/ItineraryCard/TripSectorHelpers.js
@@ -3,9 +3,9 @@
 import * as DateFNS from 'date-fns';
 
 export const timeSimpleFormat = 'H:mm';
-export const dateFormat = 'ddd D MMM';
+export const dateFormat = 'EEE d MMM';
 
 export const getFormattedDate = (
   time: ?string,
   format: ?string = timeSimpleFormat,
-) => DateFNS.format(DateFNS.parse(time), format);
+) => time && format && DateFNS.format(DateFNS.parseISO(time), format);
--- a/apps/core/components/ItineraryCard/TripSectorHelpers.js
+++ b/apps/core/components/ItineraryCard/TripSectorHelpers.js
@@ -1,11 +1,9 @@
 // @flow
 
 import * as DateFNS from 'date-fns';
-
-export const timeSimpleFormat = 'H:mm';
-export const dateFormat = 'EEE d MMM';
+import { HOURS_MINUTES_FORMAT } from '@kiwicom/margarita-config';
 
 export const getFormattedDate = (
   time: ?string,
-  format: ?string = timeSimpleFormat,
+  format: ?string = HOURS_MINUTES_FORMAT,
 ) => time && format && DateFNS.format(DateFNS.parseISO(time), format);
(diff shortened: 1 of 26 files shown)


apps/core/components/ItineraryCard/TripSectorHelpers.js
```diff
@@ -1,11 +1,9 @@
...
-export const timeSimpleFormat = 'H:mm';
-export const dateFormat = 'ddd D MMM';
+import { HOURS_MINUTES_FORMAT } from '@kiwicom/margarita-config';
 
 export const getFormattedDate = (
   time: ?string,
-  format: ?string = timeSimpleFormat,
-) => DateFNS.format(DateFNS.parse(time), format);
+  format: ?string = HOURS_MINUTES_FORMAT,
+) => time && format && DateFNS.format(DateFNS.parseISO(time), format);
```
FilipMessa: Do you need to check if the format exist? because you set default value?

apps/core/scenes/results/Results.js
```diff
@@ -40,13 +41,17 @@ export default class Results extends React.Component<Props> {
...
+    );
 
-    const from = DateFNS.format(DateFNS.parse(dateFrom), dateFormat);
     const tripType = returnDateFrom ? 'Return' : 'OneWay';
     const to = returnDateFrom
-      ? DateFNS.format(DateFNS.parse(returnDateFrom), dateFormat)
+      ? DateFNS.format(
+          DateFNS.parseISO(returnDateFrom),
+          ORDINAL_DAY_MONTH_FORMAT,
+        )
```
FilipMessa:
```
DateFNS.format(
          DateFNS.parseISO(returnDateFrom),
          ORDINAL_DAY_MONTH_FORMAT,
        )
```
don't repeat yourself

apps/core/components/ItineraryCard/TripSectorHelpers.js
jvaclavik: Because `parseISO` don't want null.

```
[flow] Cannot call `DateFNS.parseISO` with `time` bound to `argument` because null or undefined [1] is incompatible with string [2]. (TripSectorHelpers.js:9:38)
```

apps/core/scenes/results/Results.js
jvaclavik: Fixed

apps/core/components/ItineraryCard/TripSectorHelpers.js
FilipMessa: what about something like this?
```
export const getFormattedDate = (time: ?string, format: ?string) =>
  time &&
  DateFNS.format(DateFNS.parseISO(time), format ?? HOURS_MINUTES_FORMAT);

```

jvaclavik: I'm getting your point now, it makes sense now :)

Done

jvaclavik:
@FilipMessa @RobinCsl @JoseDuSV 🚀 ?

PR #558, opened 2019-03-19
Fix: Get rid of type 'disabled' in favour of the html disabled prop

### LINKED ISSUE #473: Disabled prop on Button does not disable click in web version
Opened by @RobinCsl 

Comment by @JoseDuSV 

> Disabled prop seems to be working correctly. It looks to me more like styling/storybook examples issue.
In storybook samples below we have button type set to disabled but disabled prop is still set to true. And because of it the button is styled as disabled but is still clickable.
https://kiwicom-universal-components.netlify.com/?selectedKind=Button&selectedStory=default
https://kiwicom-universal-components.netlify.com/?selectedKind=Button&selectedStory=disabled
I didn't found any disabled button in current designs but for example according to Orbit ( https://bit.ly/2QfgylF ) we probably do not need disabled type. And setting disabled prop should just set button opacity (which currently does) and button type/color should stay the same.

(Original issue https://github.com/kiwicom/universal-components/issues/162)

aalises:
The `disabled` type did not necessarily set the HTML disabled property on the elements. Hence, the styling was applied (cursor, opacity) but the click events were actually performed. We got rid of the disabled type in favour of the html disabled prop which sets the styles and also prevents the click.

Fixes #473

### DIFF
--- a/packages/universal-components/src/Button/Button.stories.js
+++ b/packages/universal-components/src/Button/Button.stories.js
@@ -34,7 +34,6 @@ storiesOf('Button', module)
         'critical',
         'facebook',
         'google',
-        'disabled',
       ],
       'primary',
     );
@@ -127,7 +126,12 @@ storiesOf('Button', module)
         <Separator />
         <Button onPress={noop} type="google" label={googleLabel} />
         <Separator />
-        <Button onPress={noop} type="disabled" label={disabledLabel} />
+        <Button
+          onPress={noop}
+          type="primary"
+          label={disabledLabel}
+          disabled={true}
+        />
         <Separator />
         <Button
           onPress={noop}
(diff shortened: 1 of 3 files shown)


RobinCsl:
It seems like you only modified the stories file and the Flow types but not the actual code. Not sure that solves the issue entirely.

aalises:
@RobinCsl  I think in this case the actual code does not have to be modified as on the `buttonInner.js`there's already this:

```
const styleSheet = StyleSheet.create({
  ...
  disabled: {
    opacity: 0.5,
  },
});
```

Which is based on the disabled **property** and not the type. Maybe I am missing something but seemed like the disabled type did not do anything as the types in the buttons only seem to change the background color.

RobinCsl:
Okay, even better 😄 
I think to finish cleaning up, you would need to remove the `disabled` styles in `packages/universal-components/src/Button/styles/shared.js`. 
Thanks!

PR #559, opened 2019-03-19
Feat: Allow styling TextInput label

### LINKED ISSUE #557: Allow the TextInput label to be styled
I would need to style the label of the TextInput component. Right now, it's only possible to style the container of the label.

**Describe the solution you'd like**
Add a `labelStyle` prop to TextInput component.

aalises:
Created two props from the style prop of `TextInput` to have `labelContainerStyle` and `labelTextStyle` in order to style the different parts of the `FormLabel` component.

**Before:** `<TextInput style={{}} />`
**Now:**  `<TextInput labelContainerStyle={{}} labelTextStyle={{}} />` 

Created tests for the functionality by snapshot difference matching and updated components reliant on the label.

Closes #557


packages/universal-components/src/FormLabel/FormLabel.js
```diff
@@ -13,7 +13,8 @@ type Props = {|
   +disabled?: boolean,
   +required?: boolean,
   +inlineLabel?: boolean,
-  +style?: StylePropType,
+  +labelContainerStyle?: StylePropType,
```
jvaclavik: What about rename `labelContainerStyle` to `containerStyle`? From my perspective it's more clear this way and it's some kind of duplicity of the component name (`label`).

`labelTextStyle` could be also renamed in the same style to `textStyle`

What do you think?

packages/universal-components/src/FormLabel/FormLabel.js
```diff
@@ -36,18 +37,22 @@ export default function FormLabel({
...
+        labelContainerStyle,
+      ]}
     >
       {required && (
         <Asteriks filled={filled} disabled={disabled}>
           *{' '}
         </Asteriks>
       )}
-      <Text style={styles.labelText}>{children}</Text>
+      <Text style={[labelTextStyle, styles.labelText]}>{children}</Text>
```
RobinCsl:
```suggestion
      <Text style={[styles.labelText, labelTextStyle]}>{children}</Text>
```
This means it will be possible to override the style.

packages/components/src/dateInput/DateInput.js
```diff
@@ -59,7 +59,9 @@ export default class DateInput extends React.Component<Props, State> {
     return (
       <View>
         {label != null && (
-          <FormLabel style={labelContainerStyle}>{label}</FormLabel>
+          <FormLabel labelContainerStyle={labelContainerStyle}>
```
RobinCsl: Maybe you could make the new prop available in these other components too, even though that's not the scope of my request. @jvaclavik WDYT?

packages/universal-components/src/FormLabel/FormLabel.js
RobinCsl: I agree with the suggestions!

aalises: Great! changed it.

aalises:
Related to #563 , there are now 3 props related to the styling, those are:

`inputStyle`: Input Container
`containerStyle`: Label Container
`textStyle`: Label Text

There might be a confusion between the `containerStyle` and the `inputStyle`, so maybe renaming the `containerStyle to labelContainerStyle` again would do it? (as container can be also the input one so it's a little bit ambiguous). WDYT @jvaclavik

packages/components/src/dateInput/DateInput.js
jvaclavik: Sounds good!

jvaclavik:
> Related to #563 , there are now 3 props related to the styling, those are:
> 
> `inputStyle`: Input Container
> `containerStyle`: Label Container
> `textStyle`: Label Text
> 
> There might be a confusion between the `containerStyle` and the `inputStyle`, so maybe renaming the `containerStyle to labelContainerStyle` again would do it? (as container can be also the input one so it's a little bit ambiguous). WDYT @jvaclavik 

I would suggest those naming:

`style`: Input Container
`formLabelContainerStyle`: Label Container
`formLabelTextStyle`: Label Text

(You can also use just `label` instead of `formLabel`, but I believe with `form` it's more clear if component has this name)

And then you can use it this way:
```
<FormLabel
  containerStyle={ContainerStyle}
  textStyle={formLabelTextStyle}
  ...
>
  ...
</FormLabel>
```

What do you think about it?

Sorry for late reply, I miss this comment (btw don't hesitate to write me on Slack if you will have to wait for me next time)

aalises:
@jvaclavik Changed the namings for clarity. The inputStyle now is a style and is implemented on #563

packages/components/src/dateInput/DateInput.js
```diff
@@ -59,7 +61,9 @@ export default class DateInput extends React.Component<Props, State> {
     return (
       <View>
         {label != null && (
-          <FormLabel style={labelContainerStyle}>{label}</FormLabel>
+          <FormLabel containerStyle={containerStyle} labelStyle={labelStyle}>
```
jvaclavik: Why not to use the same convention also here?

```
containerStyle => formLabelContainerStyle
labelStyle => formLabelTextStyle
```

packages/components/src/dateInput/DateInput.js
```diff
@@ -13,7 +13,8 @@ import {
 
 export type Props = {|
   +label?: React.Node,
-  +labelContainerStyle?: StylePropType,
+  +containerStyle?: StylePropType,
```
jvaclavik: Why not to use the same convention also here?

containerStyle => formLabelContainerStyle
labelStyle => formLabelTextStyle

aalises: @jvaclavik Done! I think the PR is ready now 👍

apps/core/scenes/passengerForm/PassengerForm.js
```diff
@@ -113,7 +113,7 @@ export default class PassengerForm extends React.Component<Props, State> {
             placeholder="Select"
             confirmLabel="OK"
             label="Nationality"
-            labelContainerStyle={styles.inputLabel}
+            containerStyle={styles.inputLabel}
```
jvaclavik: One more renaming here please 🙏 😉

aalises: Done! Renamed all the Pickers (ios, android, web) props.

packages/universal-components/src/Picker/PickerTypes.js
```diff
@@ -12,7 +12,11 @@ export type PickerOption = {|
 
 export type Props = {|
   +label?: React.Node,
-  +labelContainerStyle?: StylePropType,
+  /**
+   * Styles for the container and the text of the label
```
jvaclavik: I believe that this comment is useless, because naming should be (and it is in this case) self-descriptive.

aalises: Done!

PR #565, opened 2019-03-19
Feat: Added ThemeProvider

JoseDuSV:
Summary: Created initial setup for theming. Current implementation allows global font family change.

The idea behind having separate `tokens` and `styles` inside theme prop is to prevent unnecessary usage of dynamic styles and rebuilding styles with every render (especially for texts). Currently `fontFamily` styles are only updated with `ThemeProvider` props change (for example if some kind of theme switch will be implemented).

Note: Tests are currently failing because changes in `Text` component broke about fifty snapshots 🤕 . So to keep PR readable I will update all the affected snapshots before merge, after we approve the core idea.

Screenshot 2019-03-19 at 13 03 49

### DIFF
--- a/apps/core/scenes/bookingDetail/menuGroups/Passengers/__tests__/__snapshots__/Bag.test.js.snap
+++ b/apps/core/scenes/bookingDetail/menuGroups/Passengers/__tests__/__snapshots__/Bag.test.js.snap
@@ -22,13 +22,12 @@ Object {
         name="baggage-checked"
         size="small"
       />
-      <Text
-        expo={false}
+      <WithTheme
         size="small"
         type="primary"
       >
         3 Cabin bags
-      </Text>
+      </WithTheme>
     </Component>
   </React.Fragment>,
 }
@@ -50,13 +49,12 @@ Object {
         name="baggage-checked"
         size="small"
       />
-      <Text
-        expo={false}
+      <WithTheme
         size="small"
         type="primary"
       >
         3 Checked baggage
-      </Text>
+      </WithTheme>
     </Component>
   </React.Fragment>,
 }
@@ -78,13 +76,12 @@ Object {
         name="baggage-checked"
         size="small"
       />
-      <Text
-        expo={false}
+      <WithTheme
         size="small"
         type="primary"
       >
         3 Personal items
-      </Text>
+      </WithTheme>
     </Component>
   </React.Fragment>,
 }
--- a/packages/universal-components/src/ThemeProvider/ThemeProvider.js
+++ b/packages/universal-components/src/ThemeProvider/ThemeProvider.js
@@ -1,0 +1,73 @@
+// @flow
+
+import * as React from 'react';
+import memoize from 'memoize-one';
+import isEqual from 'react-fast-compare';
+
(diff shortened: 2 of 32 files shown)


RobinCsl:
Nice!! I will review this tomorrow!

jvaclavik:
😄

packages/universal-components/src/ThemeProvider/ThemeProvider.js
```diff
@@ -0,0 +1,82 @@
...
+export default class ThemeProvider extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+
+    this.state = {
+      ...defaultState,
+    };
+  }
+
+  static getDerivedStateFromProps(props: Props) {
```
RobinCsl: From [React Docs](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops), it seems like `getDerivedStateFromProps` should be "avoided" if possible. Did you consider using memoization instead? It seems like it should fit this scenario.
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization

jvaclavik:
LGTM 👍

packages/universal-components/src/ThemeProvider/ThemeProvider.js
JoseDuSV: @RobinCsl Thanks for suggestion I wasn't worried about it. Because this wrapper should be used at the highest hierarchy level. But it's true that we cannot be sure about that.
So I updated it to use `memoize`, it also saved few lines in implementation because with this solution we don't need state anymore.

The change affected `ThemeProvider` and `ThemeHelpers` files. And I also already updated snapshots.

FilipMessa:
I know this is not part of this PR but
I think that we should also change eslint warining messager from : 
```'Please use @kiwicom/universal-components instead, and @kiwicom/margarita-components for TouchableWithoutFeedback',``` 
into 
```'Please use @kiwicom/universal-components instead, and @kiwicom/margarita-components for TouchableWithoutFeedback and Text'```?

File: apps/core/scenes/bookingDetail/menuGroups/Passengers/__tests__/__snapshots__/Bag.test.js.snap [hunk 2]
```diff
@@ -22,13 +22,12 @@ Object {
         name="baggage-checked"
         size="small"
       />
-      <Text
-        expo={false}
+      <WithTheme
```
RobinCsl: Snapshot renders `withTheme`?! Weird!

JoseDuSV: True. But I think that main problem is basically in our usage of shallow rendering with combination of `react-native-testing-library` implementation.
It seems like `enzyme` have the similar issue https://github.com/airbnb/enzyme/issues/539 but sadly there doesn't seem to be any equivalent to suggested `.first().shallow()` inside `react-native-testing-library`.

As we mentioned before, we should probably rethink and unify out testing strategy https://kentcdodds.com/blog/why-i-never-use-shallow-rendering .

Issue #556, opened 2019-03-19
Create the new search itineraries queries.


FilipMessa:
Create the two new queries for search itineraries.
In the margarita app we currently have `searchItineraries` query for searching itineraries.
The `searchItineraries` query accept the `ItinerariesSearchInput` which schema looks like this
```
travelFrom: [String]
travelTo: [String]
dateFrom: Date!
dateTo: Date
returnDateFrom: Date
passengers: PassengerInputs  
```
This input allows a search for **one-way** and also for the **return** type of itineraries. But it is confusing what field you need to be filled for oneway and what is required for return. And also is hard to extend this query, in future will be added multipcity and nomad. So one query for all type of search will change to mass, it will be hard to know what should be filled for when.

Because that looks like a better approach will create a standalone query for every type of search. It will easy to extend it and manage in the future.

Prepare the two searching queries base on the search type, the `searchOnewayItineraries` for oneway and the `SearchReturnItineraries` for return.

the schema design proposal:
```
input DateRangeInput {
	start: Date
	end: Date
  }

  enum SortSearchInput {
    PRICE
    DURATION
    QUALITY
    DATE 
  }

 enum OrderSearchInput {
   ASC
   DESC
 }

input LocationItineraryInput {
  slug: [String]
  ids:  [String] // It should be GlobalIDs not locationIds
 }

input ItineraryOneWayInput {
  origin:  LocationItineraryInput
  destination:  LocationItineraryInput
  outboundDate: DateRangeInput
}

input SearchOnewayInput {
  order: OrderSearchInput
  sort: SortSearchInput
  itinerary: ItineraryOnewayInput
  passengers: PassengersInput // schema already implemented
}

input ItineraryReturnInput {
  origin:  LocationItineraryInput
  destination: LocationItineraryInput
  outboundDate: DateRangeDate
  inboundDate: DateRangeInput
}

input SearchReturnInput {
  order: OrderSearchInput
  sort:  SortSearchInput
  itinerary: ItineraryReturnInput
  passengers: PassengersInput // schema already implemented
}

```

Tequila API docs: https://tequila.kiwi.com/portal/resources/search_api

PS: Don't try to do it in one PR, at least split it into two, one for `searchOnewayItineraries` and one for the `searchReturnItineraries`

tbergq:
I consider it to be a greater mess when we now need add refetch containers for sorting. This means one component for query and one component for refetch pr type. 

[...]

PR #566, opened 2019-03-20
Feat: Create Search One Way GraphQL query

aalises:
Related to #556 

Initial approach at creating the `searchOneWayItineraries` query while keeping the original `searchItineraries`query. It is only the same one without the return fields, and eventually will rename the original query to `searchReturnItineraries`. Have created some common types for all the queries.

This is my first using Relay and Flow (Apollo and TS guy...), and still there are some things I am getting used to.

**TODO:** `DateRangeInput` type (Will do on separate issue later on to unify with DateRangePicker, cc @jvaclavik )


FilipMessa:
This is not the query what we expected, the input `ItinerariesOneWaySearchInput` schema should look like this:

``` 
input SearchOnewayInput {
  order: OrderSearchInput
  sort: SortSearchInput
  itinerary: ItineraryOnewayInput
  passengers: PassengersInput // schema already implemented
}
```

not like in your implementation: 

```
input ItinerariesOneWaySearchInput {
  travelFrom: [String]!
  travelTo: [String]
  dateFrom: Date!
  dateTo: Date
  passengers: PassengersInput
}
```

all design types are described in the issue https://github.com/kiwicom/margarita/issues/556. Do you need more information?

FilipMessa:
Or your plan is do it in smalls PRs?

aalises:
@FilipMessa Yes! was planning on getting the query working with the old format now and then implementing the new search schema on another PR while keeping the other query on for retrocompatibility.

FilipMessa:
OK so don't care about FE now a lot, because FE will still use old query till both new will be done.

So If I understand right, this is just first PR for one-way search, where your goal just made it work. In the next PR you will start to refactor this to the new design.

```
input SearchOnewayInput {
  order: OrderSearchInput
  sort: SortSearchInput
  itinerary: ItineraryOnewayInput
  passengers: PassengersInput // schema already implemented
}
```

Am I right?

aalises:
Exactly! @FilipMessa

apps/graphql/src/apps/itinerary/types/outputs/Itinerary.js
```diff
@@ -14,30 +14,32 @@ export type Itinerary = {|
...
-      resolve: ({ sectors }: ItinerariesType): Sector[] => {
-        if (sectors == null) {
-          return [];
-        }
-        return sectors.map((sector, index) => ({
-          ...sector,
-          stopoverDuration: getStopoverDuration(sector, sectors[index - 1]),
-        }));
-      },
+const itineraryResponseTypes = {
```
FilipMessa: What was your reason to put this out from the `GraphQLObjectType`?

FilipMessa: Just for my information

aalises: @FilipMessa Hey! because on an earlier version I created two different objects, one for each query, and I reused the fields, but now I just use one so we can put it back in.

FilipMessa: OK I understand, we can have it outside, but I am just not sure about that name. 
From my perspective, it is fields not typed.

PR #568, opened 2019-03-20
Render calendar for RangeDatePicker

jvaclavik:
This is the first step in creating range date picker. There are still lot of things to do in next PRs. Even this PR is quite big, but I hope it should be still readable enough, because there are lot of styles and easy components.  
I noticed that I forgot to show in video that the content of RangeDatePicker is scrollable :)

~~Btw I'm not sure why there is a small delay during opening modal and clicking on specific date (you can notice it on simulators). If you have idea, let me know. I plan to spare time on it in next PR.~~ (fixed)

ezgif com-optimize


apps/core/scenes/search/Datepickers.js
```diff
@@ -124,7 +125,7 @@ class Datepickers extends React.Component<Props, State> {
             />
           )}
         </PickersWrapper>
-        <DatePicker
+        <RangeDatePicker
```
JoseDuSV: As mentioned in description, there is something really hungry happening during Picker opening.
I've checked it in Chrome (so it doesn't seems to be platform specific) and as you can see it generates peak on every opening. Then it takes quite a long time (400ms) to actually show picker modal.
Screenshot 2019-03-25 at 11 36 51

apps/core/components/rangeDatePicker/RangeDatePickerTypes.js
```diff
@@ -0,0 +1,18 @@
...
+  +mode?: 'date' | 'time' | 'datetime',
+  +datePickerMode?: 'calendar' | 'spinner' | 'default', // this prop is supported only on android
+  +date?: Date,
+  +minDate?: Date,
+  +maxDate?: Date,
+  +onConfirm: Date => void,
+  +onDismiss: () => void,
+  +labels: {|
+    +cancel: React.Node, // this prop is supported only on web & iOS
+    +confirm: React.Node, // this prop is supported only on iOS
```
JoseDuSV: Comments for `cancel` and `confirm` are probably no longer valid. Buttons are currently visible on all platforms.

apps/core/components/rangeDatePicker/RangeDatePickerTypes.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+import * as React from 'react';
+
+export type Props = {|
+  +isVisible: boolean,
+  +mode?: 'date' | 'time' | 'datetime',
+  +datePickerMode?: 'calendar' | 'spinner' | 'default', // this prop is supported only on android
```
JoseDuSV: This one is not and probably will not be used.

apps/core/components/rangeDatePicker/RangeDatePickerTypes.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+import * as React from 'react';
+
+export type Props = {|
+  +isVisible: boolean,
+  +mode?: 'date' | 'time' | 'datetime',
```
JoseDuSV: I don't think there is a plan to implement other modes, so we can probably remove this one.

apps/core/components/rangeDatePicker/RenderDay.js
```diff
@@ -0,0 +1,96 @@
...
+type Props = {|
+  +day: Date,
+  +onPress: Date => void,
+  +price?: ?string,
+  +selectedDate: Date,
+|};
+
+const DayPrice = ({ price }) => (
+  <View>
+    <Text numberOfLines={1} style={styles.price}>
```
JoseDuSV: Performance issue seems to be related to this component. Basically delay increases with the number of rendered days. When I've quickly tried to remove extra wrapper and used `Text` component directly from `react-native` picker seems to be slightly quicker.

jvaclavik: I don't see any real difference while using `Text` from react native.

When I render only one month of calendar, it's quicker (but still not perfect). 

It's also probably not related with  generating of calendar, because when I use dummy data, delay is practically identical like with real data. So there could be some problem with too many number of elements, Flatlist or mapping weeks into the FlatList. It's probably not related to overdraw.

apps/core/components/rangeDatePicker/RangeDatePickerTypes.js
jvaclavik: Removed

apps/core/components/rangeDatePicker/RangeDatePickerTypes.js
jvaclavik: Removed

apps/core/components/rangeDatePicker/RangeDatePickerTypes.js
jvaclavik: Removed

apps/core/components/rangeDatePicker/RenderDay.js
JoseDuSV: Right, it seems to be more likely related to number of elements.
Text itself is a small fraction of it. When I set number of months to some high value like 50. Show delay decreases with every element removed from `RenderDay` component. And when I keep just something like `return <Text>12</Text>` inside it is relatively quick even with all those 1500 days.
FlatList can be definitely worth a try. It seems to work perfectly for ResultsList where we are rendering probably even more components.

apps/core/components/rangeDatePicker/RangeDatePicker.js
```diff
@@ -0,0 +1,117 @@
+/* @flow */
```
tbergq: My OCD ticks from this flow annotation 😱

apps/core/components/rangeDatePicker/RangeDatePicker.js
```diff
@@ -0,0 +1,117 @@
...
+
+  constructor(props: Props) {
+    super(props);
+
+    this.state = {
+      date: props.date ?? new Date(),
+    };
+  }
+
+  componentDidUpdate = (prevProps: Props) => {
```
tbergq: This should not be an arrow function

apps/core/components/rangeDatePicker/RenderDay.js
jvaclavik: So I believe it should be fine now.

There are following changes:
- calculations of day matrixes were moved from `RangeDatePickerContent` to `RenderMonth` (which is inside on `FlatList`) and `FlatList` got following prop: `initialNumToRender={2}` (thanks @tbergq for the advice)
- implemented own `shouldComponentUpdate` in `RenderMonth`
- use `memoize` for rendering in `RenderMonth`
- fix overdraw in the calendar

Calendar is also faster on **production mode**

apps/core/components/rangeDatePicker/RangeDatePicker.js
jvaclavik: Fixed

apps/core/components/rangeDatePicker/RangeDatePicker.js
jvaclavik: Fixed

apps/core/components/rangeDatePicker/RangeDatePickerContent.js
```diff
@@ -0,0 +1,65 @@
...
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import { getNextMonths, type MonthDate } from './libs';
+import RenderMonth from './RenderMonth';
+import DayNames from './DayNames';
+import RangeDateConfig from './RangeDateConfig';
+
+type Props = {|
+  onDayPress: Date => void,
+  selectedDate: Date,
```
tbergq: please make props read-only

apps/core/components/rangeDatePicker/RenderDay.js
```diff
@@ -0,0 +1,96 @@
...
+              weight="bold"
+              style={[
+                styles.day,
+                isDaySelected && styles.selectedDateText,
+                isDayInPast && styles.dayInPast,
+              ]}
+            >
+              {format(day, 'd')}
+            </Text>
+            {price && <DayPrice price={price} />}
```
tbergq: price = '' -> crash on android

apps/core/components/rangeDatePicker/RenderMonth.js
```diff
@@ -0,0 +1,75 @@
...
+  +selectedDate: Date,
+|};
+
+export default class RenderMonth extends React.Component<Props> {
+  shouldComponentUpdate(nextProps: Props) {
+    const checkDate = props =>
+      props.selectedDate &&
+      props.selectedDate.getMonth() === props.monthDate.month &&
+      props.selectedDate.getFullYear() === props.monthDate.year;
+    return checkDate(nextProps) || checkDate(this.props);
```
tbergq: what about just using `import isEqual from 'react-fast-compare';`

apps/core/components/rangeDatePicker/libs.js
```diff
@@ -0,0 +1,58 @@
...
+  endOfMonth,
+  isSameMonth,
+  startOfISOWeek,
+  startOfMonth,
+  eachWeekOfInterval,
+} from 'date-fns';
+
+export type MonthDate = {|
+  year: number,
+  month: number,
```
tbergq: read-only

apps/core/components/rangeDatePicker/RenderMonth.js
jvaclavik: How would you use it?  `props.selectedDate` and `props.monthDate` is not equal.

apps/core/components/rangeDatePicker/RangeDatePickerContent.js
jvaclavik: Fixed

apps/core/components/rangeDatePicker/RenderDay.js
jvaclavik: Fixed

apps/core/components/rangeDatePicker/libs.js
jvaclavik: Fixed

apps/core/components/rangeDatePicker/RenderMonth.js
tbergq: You are right, forget that comment 🙈

jvaclavik:
Btw `RangeDatePicker` moved to `universal-components`

packages/universal-components/src/RangeDatePicker/RenderDay.js
```diff
@@ -0,0 +1,100 @@
...
+  function handlePress() {
+    onPress(day);
+  }
+
+  const isDaySelected = isSameDay(selectedDate, day);
+  const isDayInPast = isBefore(day, new Date());
+
+  return (
+    <View style={{ flex: 1 }}>
+      <Touchable onPress={handlePress} disabled={isDayInPast}>
```
JoseDuSV: Disabled condition should be extended, past days are ok, but empty day slots (shown below) are still clickable. Best way to try it is in web version.

Screenshot 2019-03-27 at 10 47 12

jvaclavik: Nice bug 🐛 now it should be fixed

packages/universal-components/src/RangeDatePicker/DayNames.js
```diff
@@ -0,0 +1,44 @@
...
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import { StyleSheet } from '../PlatformStyleSheet';
+import { Text } from '../Text';
+import RangeDateConfig from './RangeDateConfig';
+
+const dayNames = ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'];
```
tbergq: Maybe there should be a way to pass inn the day names, or set some localization 🤔  Since I want to use this in hotels, I need this to be translatable.

packages/universal-components/src/RangeDatePicker/RangeDatePicker.js
```diff
@@ -0,0 +1,115 @@
...
+import { Touchable } from '../Touchable';
+import { StyleSheet } from '../PlatformStyleSheet';
+import type { Props } from './RangeDatePickerTypes';
+import RangeDatePickerContent from './RangeDatePickerContent';
+
+type State = {
+  date: Date,
+};
+
+const parsePropsToState = ({ date }: Props) => {
```
FilipMessa: This naming is a little confusing for me, from that name, I will be expected that will parse all props to state. So what about `parseDatePropsToState` or something more descriptive? But it is just Idea

packages/universal-components/src/RangeDatePicker/RangeDateConfig.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+
+export default {
+  dayItemHeight: 52,
+  dayItemPadding: 8,
+  numberOfRenderedMonths: 24,
```
JoseDuSV: Maybe we can decrease this number just to one year. But it is totally up to you. Even two years are ok if it no longer affects the performance.

packages/universal-components/src/RangeDatePicker/RenderMonth.js
```diff
@@ -0,0 +1,76 @@
...
+
+type Props = {|
+  +monthDate: MonthDate,
+  +onDayPress: Date => void,
+  +selectedDate: Date,
+|};
+
+export default class RenderMonth extends React.Component<Props> {
+  shouldComponentUpdate(nextProps: Props) {
+    const checkDate = props =>
```
JoseDuSV: We can move this function outside of class.

packages/universal-components/src/RangeDatePicker/RangeDateConfig.js
tbergq: When it is lazy-calculated, it should not matter 😊 

I see that on web you can select as far into the future as you want, while on mobile there is just one year into the future.

packages/universal-components/src/RangeDatePicker/RenderDay.js
JoseDuSV: Thanks, it's ok now.

packages/universal-components/src/RangeDatePicker/RangeDatePicker.js
```diff
@@ -0,0 +1,115 @@
...
+          <RangeDatePickerContent
+            selectedDate={this.state.date}
+            onDayPress={this.handleChangeDate}
+          />
+          <View style={styles.buttonsContainer}>
+            <Touchable
+              onPress={this.handleDismiss}
+              style={styles.confirmButton}
+            >
+              <Text style={styles.buttonText}>Cancel</Text>
```
JoseDuSV: After moving to universal `Cancel` and `Ok` labels should be defined by props. It is the same case like with day names, basically to have it translatable.

apps/core/components/rangeDatePicker/RenderDay.js
JoseDuSV: Thanks, it definitely feels quicker now. And you are right that production mode will also help.

packages/universal-components/src/RangeDatePicker/DayNames.js
jvaclavik: I can do it this way (it's not perfect, there is probably no better way how to do it in date-fns):
```
const now = new Date();
const arr = eachDay(startOfWeek(now), endOfWeek(now)
arr.reduce((a, d) => {
  a.push(format(d, 'dd'));
  return a;
});
```
https://github.com/date-fns/date-fns/issues/644

What do you think?

jvaclavik: Or do you want to have it from `props`?

tbergq: The problem with this is that you can't do `import locale from 'date-fns/esm/locale/{localeName}'` in react-native. It all has to be bundled in advance. 

Seing this post also raises an interesting question, some locales have Sunday as the first day of the week, is this calendar prepared to handle this?

packages/universal-components/src/RangeDatePicker/RangeDateConfig.js
JoseDuSV: Ok, it probably depends more on our API and how far into future can really return any results.
So we can keep 24 for now or maybe we can allow this value to be defined by props to make component even more flexible.

tbergq: > defined by props to make component even more flexible

Absolutely the best option I would say

packages/universal-components/src/RangeDatePicker/DayNames.js
jvaclavik: 1) And what about pass localization to component? Something like:

```
import locale from 'date-fns/esm/locale/cs'
import {RangeDatePicker} from '@kiwicom/universal-components'
...
<RangeDatePicker 
  locale={locale}
/>
```

2) Right now it isn't possible to have Sunday as a first day in week, but I believe it shouldn't be hard to update it. Is it OK if I will do it in next PR?

It's mainly about using different config for this function:
https://date-fns.org/v1.9.0/docs/startOfWeek

packages/universal-components/src/RangeDatePicker/RenderMonth.js
jvaclavik: Very good point

jvaclavik: Done

packages/universal-components/src/RangeDatePicker/RangeDatePicker.js
jvaclavik: Sounds good, done

packages/universal-components/src/RangeDatePicker/RangeDatePicker.js

packages/universal-components/src/RangeDatePicker/RangeDateConfig.js
jvaclavik: It's true. Done

packages/universal-components/src/RangeDatePicker/DayNames.js
jvaclavik: @tbergq I just added start week support and generated names of the week. So right now you can choose which day should be first in the week.

But localization should be resolved in another PR.

jvaclavik:
@JoseDuSV @tbergq @FilipMessa 🚀

packages/universal-components/src/RangeDatePicker/DayNames.js
tbergq: 1. I think I would prefer that you pass in the name array. Passing the locale, you force the consumer to depend on `date-fns` and we you `Intl` in `mobile` repo for these things. 

2. Ok to do it in a later PR

PR #576, opened 2019-03-21
Feat: Added Profile screen

JoseDuSV:
Summary: Added Profile screen to structure and navigation which will serve for user login and other possible user related settings.


apps/core/scenes/profile/Profile.js
```diff
@@ -0,0 +1,60 @@
...
+import * as React from 'react';
+import { ScrollView } from 'react-native';
+import { StyleSheet, Button, Card } from '@kiwicom/universal-components';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import { margaritaTokens } from '@kiwicom/margarita-design-tokens';
+import { Illustration, Text } from '@kiwicom/margarita-components';
+
+type Props = {||};
+
+export default class Profile extends React.Component<Props> {
```
FilipMessa: Why, not just functional component?

apps/core/scenes/profile/Profile.js
```diff
@@ -0,0 +1,60 @@
...
+  handleSignIn = () => {
+    // @TODO
+  };
+
+  render() {
+    return (
+      <ScrollView
```
FilipMessa: What is reason for ScrollView not just a View?

apps/core/scenes/profile/Profile.js
```diff
@@ -0,0 +1,60 @@
...
+        style={styles.container}
+        contentContainerStyle={styles.containerContent}
+      >
+        <Illustration name="Lounge" style={styles.illustration} />
```
jvaclavik: Why is the illustration only on the web?

Simulator Screen Shot - iPhone X - 2019-03-22 at 10 49 43

apps/core/scenes/profile/Profile.js
```diff
@@ -0,0 +1,59 @@
...
+  containerContent: {
+    alignItems: 'center',
+  },
+  illustration: {
+    paddingTop: parseInt(defaultTokens.spaceMedium, 10),
+    paddingBottom: parseInt(defaultTokens.spaceLarge, 10),
+  },
+  card: {
+    width: '100%',
+    web: {
```
jvaclavik: What about using new `ContentContainer` component instead of this?

jvaclavik:
Please could you attach link to design? I couldn't find it in figma.

apps/core/scenes/profile/Profile.js
JoseDuSV: I didn't included updated Illustration component in this PR. I just added it and should be fine now.

Also maybe if you are asking why I picked `Lounge` illustration instead of `Login` one. Simply because current login illustration available on server is outdated and quite ugly (with plane cut in half). There is already new way much better one prepared in design resources, so we can later swap it when it will be ready.

apps/core/scenes/profile/Profile.js
JoseDuSV: True, thanks, I forget that we already have it. Replaced

JoseDuSV:
> Please could you attach link to design? I couldn't find it in figma.

There is something in Figma https://www.figma.com/file/WwnGeAryGbnODNIHi1PDMtj0/Login?node-id=0%3A1 but it is not that relevant. It's basically just illustration, little bit of text and different kinds of sign in buttons. Our goal for now is not to cover whole registration flow. But just to have ability to log-in  with social account.

apps/core/scenes/profile/Profile.js
jvaclavik: OK, but I think you don't need this style anymore. `maxWidth` should be handled by `ContentContainer`

apps/core/scenes/profile/Profile.js
JoseDuSV: As Honza suggested it is replaced with `ContentContainer` component now. Which also uses `ScrollView` inside. And it is basically to have screen lets say future proof. For now there is not much content so scrolling can seems to be unnecessary. But if we for example expand description later and add more buttons it can easily go over the bottom edge.

But even with current content, scrolling can be necessary if user open the website in some really small window.

apps/core/scenes/profile/Profile.js
JoseDuSV: Because I'll need to handle sign-in buttons events inside class.

FilipMessa: OK so it is prepared for next PR, I understand. Thanks

apps/core/scenes/profile/Profile.js
JoseDuSV: 👍 Right, unnecessary style removed.

PR #579, opened 2019-03-21
Fix: Checkbox text implementation & checked warning

JoseDuSV:
Summary: 
- Updated `Checkbox` component to remove `checked` related console warning in web version by setting `readOnly` prop on input.
- Replaced Text implementation with correct one.
- Removed unnecessary styles folder.


packages/universal-components/src/Checkbox/Checkbox.web.js
```diff
@@ -38,6 +38,7 @@ export default function Checkbox({
         name={name}
         checked={checked}
         style={styleInput}
+        readOnly
```
jvaclavik: I don't understand. Why it should be `readOnly` on the web? What for is Checkbox if you can't update its value? Or is it for something else?

JoseDuSV: Because this input is actually hidden and all switching stuff is handled by the platform shared rest of the component.
It is there just to have valid html output on web.

jvaclavik: OK, I understand now. Thanks 👍

PR #583, opened 2019-03-24
Refactor: split fetch service into two, the fetch and the tequilaFetch.

FilipMessa:
I will need to fetch data from different API than Tequila for localization, before that I split fetch service to two.

- `fetch` as a global fetch service
- `tequilaFetch` service for Tequila API


apps/graphql/src/services/fetch/helpers/extendHttpHeaders.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+
+import type { Options } from '../globalFetch';
+
+export default function extendHttpHeaders(
```
JoseDuSV: Name of this function was slightly confusing for me at first with parameters combination.
And I was thinking why it is not called for example something like `mergeFetchOptions`.

But I understand that we need it just for headers merge and there will not be beccesity to extend it also for other options merging.
So it is quite fine like this.

apps/graphql/src/services/fetch/helpers/extendHttpHeaders.js
```diff
@@ -0,0 +1,16 @@
...
+  options: Object,
```
JoseDuSV: We can probably use `Options` type from `globalFetch` here.

PR #585, opened 2019-03-25
Style arrows in RangeDatePicker

jvaclavik:
Mar-25-2019 11-28-32

Animation is rendered only on iOS and web. On Android there are some problems with weird position of elements before the animation begins. It's probably connected with this one: https://github.com/facebook/react-native/issues/17510

It should be possible to resolve it after some time but it's probably not a priority right now.

### DIFF
--- a/packages/universal-components/src/RangeDatePicker/DayItemArrow.js
+++ b/packages/universal-components/src/RangeDatePicker/DayItemArrow.js
@@ -1,0 +1,97 @@
+// @flow
+
+import * as React from 'react';
+import { Animated, Easing, View, Platform } from 'react-native';
+
+import { type StylePropType, StyleSheet } from '../PlatformStyleSheet';
+import { designTokens } from '../DesignTokens';
+
+type Props = {|
+  +style?: StylePropType,
+|};
+
+const SHARED_ANIMATION_CONFIG = {
+  duration: designTokens.durationArrowScaleAnimation,
+  useNativeDriver: true,
+  easing: Easing.ease,
+};
+
+const isAnimationEnabled = Platform.OS !== 'android';
+
+export default class DayItemArrow extends React.Component<Props> {
+  componentDidMount() {
+    this.animating = true;
+    if (isAnimationEnabled) {
+      this.handleAnimation();
+    }
+  }
+
+  componentWillUnmount() {
+    this.animating = false;
+  }
+
+  animating: boolean;
+  scale: Animated.Value = new Animated.Value(0);
+
+  handleAnimation() {
+    Animated.sequence([
+      Animated.delay(100),
+      Animated.timing(this.scale, {
+        ...SHARED_ANIMATION_CONFIG,
+        toValue: 1,
+      }),
+      Animated.timing(this.scale, {
+        ...SHARED_ANIMATION_CONFIG,
+        toValue: 0,
+      }),
+    ]).start(() => {
+      if (this.animating) {
+        this.handleAnimation();
+      }
+    });
+  }
+
+  render() {
+    return (
+      <View style={[styles.container, this.props.style]}>
+        {isAnimationEnabled && (
+          <Animated.View
+            style={[
+              styles.easingArrow,
+              {
+                transform: [{ scale: this.scale }],
+              },
+            ]}
+          />
+        )}
+        <View style={styles.innerArrow} />
+      </View>
+    );
+  }
+}
+
+const innerArrowSize = 8;
+const easingArrowSize = 20;
+const innerArrowMargin = (easingArrowSize - innerArrowSize) / 2;
+
+const styles = StyleSheet.create({
+  container: {
+    position: 'absolute',
+    transform: [{ rotate: '45deg' }],
+    zIndex: -1,
+  },
+  easingArrow: {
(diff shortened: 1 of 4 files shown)


RobinCsl:
Preview looks great! 🎉

packages/universal-components/src/RangeDatePicker/DayItemArrow.js
```diff
@@ -0,0 +1,103 @@
...
+type Props = {|
+  +style?: StylePropType,
+|};
+
+type State = {|
+  scale: Animated.Value,
+|};
+
+const scaleAnimationDuration =
+  parseFloat(defaultTokens.durationSlow) * 2 * 1000;
```
JoseDuSV: Does `durationSlow` from `defaultTokens` still makes sense here if we modify it that much?
What about adding new value into universal tokens?

jvaclavik: True 😄 done

packages/universal-components/src/RangeDatePicker/DayItemArrow.js
```diff
@@ -0,0 +1,99 @@
...
+|};
+
+const SHARED_ANIMATION_CONFIG = {
+  duration: designTokens.durationArrowScaleAnimation,
+  useNativeDriver: true,
+  easing: Easing.ease,
+};
+
+export default class DayItemArrow extends React.Component<Props, State> {
+  static animating: boolean;
```
tbergq: Does it need to be static?

packages/universal-components/src/RangeDatePicker/RenderDay.js
```diff
@@ -42,28 +43,32 @@ export default function RenderDay({
   return (
     <View style={styles.container}>
       <Touchable onPress={handlePress} disabled={isDayInPast || isFieldEmpty}>
```
JoseDuSV: I'm not sure if it is caused by this PR. But on Android inactive days are still clickable.
Screenshot 2019-03-28 at 11 49 23

packages/universal-components/src/RangeDatePicker/DayItemArrow.js
jvaclavik: No, should I use it in state? And will be fine if I will set state in `componentDidMount`?

packages/universal-components/src/RangeDatePicker/DayItemArrow.js
```diff
@@ -0,0 +1,99 @@
...
+  state = {
+    scale: new Animated.Value(0),
+  };
+
+  componentDidMount() {
+    DayItemArrow.animating = true;
+    this.handleAnimation();
```
JoseDuSV: If I understand it correctly animation is disabled on Android. But the sequence is still running in background on this platform. Can we disable it?

Maybe it will be handy to have something like `isAnimationEnabled` and set it `false` on Android for now. And then just work with this prop. Basically just to prevent to have `Platform.OS !== 'android'` condition in multiple places.

packages/universal-components/src/RangeDatePicker/RenderDay.js
jvaclavik: Should be fixed now

packages/universal-components/src/RangeDatePicker/DayItemArrow.js
jvaclavik: Good point, fixed

packages/universal-components/src/RangeDatePicker/DayItemArrow.js
jvaclavik: Done

packages/universal-components/src/RangeDatePicker/RenderDay.js
JoseDuSV: Thanks, it looks good now.

packages/universal-components/src/RangeDatePicker/DayItemArrow.js
```diff
@@ -0,0 +1,97 @@
...
+  easing: Easing.ease,
+};
+
+const isAnimationEnabled = Platform.OS !== 'android';
+
+export default class DayItemArrow extends React.Component<Props> {
+  componentDidMount() {
+    this.animating = true;
+    if (isAnimationEnabled) {
+      this.handleAnimation();
```
JoseDuSV: I didn't investigated it further. But it will be probably worth checking.
`handleAnimation` is started twice on initial render. It doesn't seems to be a problem of this component, but for some reason `RenderDay` is mounted, unmounted and mounted again after opening the picker.

But we can definitely merge this one and investigate it (if necessary) in different PR.

jvaclavik: I think it should be because I'm using this component twice in RenderDay

JoseDuSV: It is fired twice for each arrow, so that's not a problem, but this is just a side effect. As I mentioned, source of this is actually that `RenderDay` is re-mounted after picker opening. But maybe there is some reason for it.

PR #586, opened 2019-03-25
Feat: Create new searchReturnItineraries query

aalises:
Created new `searchReturnItineraries` query. Related to #556

### DIFF
--- a/apps/graphql/src/apps/itinerary/Itinerary.js
+++ b/apps/graphql/src/apps/itinerary/Itinerary.js
@@ -45,7 +45,7 @@ export type ItinerariesReturnSearchParameters = {|
     +origin: SearchLocation,
     +destination?: SearchLocation,
     +outboundDate: DateRange,
-    +inboundDate?: DateRange,
+    +inboundDate: DateRange,
   |},
   +passengers?: Passengers,
 |};
--- a/apps/graphql/src/apps/itinerary/Itinerary.js
+++ b/apps/graphql/src/apps/itinerary/Itinerary.js
@@ -3,49 +3,58 @@
 import type { Location } from '../location/Location';
 import type { RouteStop } from '../booking/Booking';
 
+type Order = 'ASC' | 'DESC';
+type Sort = 'price' | 'duration' | 'quality' | 'date' | 'popularity';
+type DateRange = {|
+  start: Date,
+  end?: Date,
+|};
+type Passengers = {|
+  +adults?: number,
+  +children?: number,
+  +infants?: number,
+|};
+
+type SearchLocation = {|
+  +ids: string[],
+|};
+
 export type ItinerariesCommonSearchParameters = {|
   +travelFrom: string[],
   +dateFrom: Date,
   +dateTo?: Date,
   +travelTo?: string[],
-  +passengers?: {|
-    +adults?: number,
-    +children?: number,
-    +infants?: number,
-  |},
+  +passengers?: Passengers,
 |};
 
 export type ItinerariesOneWaySearchParameters = {|
-  +order?: 'ASC' | 'DESC',
-  +sort?: 'price' | 'duration' | 'quality' | 'date' | 'popularity',
+  +order?: Order,
+  +sort?: Sort,
   +itinerary: {|
-    +origin: {|
-      +ids: string[],
-    |},
-    +destination?: {|
-      +ids: ?(string[]),
-    |},
-    +outboundDate: {|
-      start: Date,
-      end?: Date,
-    |},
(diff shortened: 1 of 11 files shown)


apps/graphql/src/apps/itinerary/Itinerary.js
```diff
@@ -36,6 +36,32 @@ export type ItinerariesOneWaySearchParameters = {|
     +infants?: number,
   |},
 |};
+
+export type ItinerariesReturnSearchParameters = {|
```
FilipMessa: There are a lot of common types with the `ItinerariesOneWaySearchParameters`,
what about creating types for them? 

For example 
```
type Sort = 'price' | 'duration' | 'quality' | 'date' | 'popularity',
type Order = 'ASC' | 'DESC',
type DateRange = {
 +start: Date,
 +end?: Date,
}
type Passengers: {...}
```

apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
```diff
@@ -53,17 +53,23 @@ export const parseParameters = (input: ItinerariesSearchParameters) => {
 };
 
 export const parseParametersNew = (
-  input: ItinerariesOneWaySearchParameters, // @TODO: Later on will be expanded to be the return one with extra fields
+  input: ItinerariesReturnSearchParameters,
```
FilipMessa: This can handle oneway and return, right? So probably more generic name will be better, for better understanding.

schema.graphql
```diff
@@ -302,6 +316,20 @@ input ItineraryOneWayInput {
   outboundDate: DateRange!
 }
 
+input ItineraryReturnInput {
+  """Origin of the flight"""
```
FilipMessa: Pls, Can you avoid using the flight in the description, because in the future margarita will support also buses and trains?

apps/graphql/src/apps/itinerary/Itinerary.js
aalises: Done!

schema.graphql
aalises: Done!

apps/graphql/src/apps/itinerary/Itinerary.js
FilipMessa: Thanks

apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
aalises: What do you propose @FilipMessa  ? I think maybe `ItinerariesRoundTripSearchParameters` would be nice. However we discussed with @RobinCsl that moving forward we should make it a union type and then add the specific parameters for each of the search queries in separate functions depending on the input (or Flow complains as for example `inboundDate` does not exist when the input type is OneWay)

apps/graphql/src/apps/itinerary/Itinerary.js
```diff
@@ -29,35 +52,27 @@ export type ItinerariesOneWaySearchParameters = {|
       start: Date,
       end?: Date,
     |},
+    +inboundDate?: {|
```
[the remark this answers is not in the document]
aalises: Yep sorry missed this one. Added as well the `SearchLocation` type

apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
FilipMessa: Can be GraphQL input type as the union type?

apps/graphql/src/apps/itinerary/Itinerary.js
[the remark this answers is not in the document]
FilipMessa: np

apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
michalsanger: GraphQL does not support union for input types.

FilipMessa: The `ItinerariesSearchParameters` will be the best but we already use it for old query so maby just `ItinerariesInputSearchParameters`? 🤔

FilipMessa: @michalsanger Yes but this is about the flow union types. As I realized after discussion with @aalises

michalsanger: Oh sorry, thanks for info :-)

PR #589, opened 2019-03-25
Add geoIP query

FilipMessa:


apps/graphql/src/apps/geoIp/apis/index.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
+import fetch from '../../../services/fetch/globalFetch';
+
+export type GeoIP = {|
+  isoCountryCode: ?string,
```
jvaclavik: I think it could be read only with `+`, what do you think?

FilipMessa: 💯
 I always forgot about that 🙈

FilipMessa: 🚀 done

Issue #591, opened 2019-03-25
Migrate the FE to use the new Search GraphQL Queries


aalises:
**Is your feature request related to a problem? Please describe.**
The new search queries (#556) use a different schema which is incompatible with the old one which the FE requires.
**Describe the solution you'd like**
Migrate the FE queries to use the new schema with the new `searchReturnItineraries` query. Get rid of the old query.

PR #592, opened 2019-03-25
Feat: Migrate the FE to use the new Search GraphQL Queries 

aalises:
Related with #591. Get rid of the old query in favour of the new search queries using the new schema.


apps/core/scenes/results/Results.js
```diff
@@ -84,6 +72,59 @@ export default class Results extends React.Component<Props> {
...
+            ...ResultsList_data
+          }
+        }
+      `;
+      break;
+  }
+  return query;
+};
+
+const parseSearchParametersByType = (
```
FilipMessa: Can you add this function inside the Component? The reason is consistency.

aalises: Sure! Done

apps/graphql/src/apps/itinerary/Itinerary.js
```diff
@@ -19,20 +19,12 @@ type SearchLocation = {|
...
-  +passengers?: Passengers,
-|};
-
 export type ItinerariesOneWaySearchParameters = {|
   +order?: Order,
   +sort?: Sort,
   +itinerary: {|
     +origin: SearchLocation,
-    +destination?: SearchLocation,
+    +destination: SearchLocation,
```
FilipMessa: The +destination: SearchLocation, is now required?
And for `ItinerariesOneWaySearchParameter` is not? I am confused.

aalises: They are both not required, changed it back. I had changed it to circumvent some Flow issue.

PR #596, opened 2019-03-26
Fix: Add OpaqueID for new search queries

aalises:


apps/graphql/src/apps/itinerary/Itinerary.js
```diff
@@ -16,7 +16,7 @@ type Passengers = {|
 |};
 
 type SearchLocation = {|
-  +ids: string[],
+  ids: string[],
```
FilipMessa: Why you removed read-only ?

apps/graphql/src/apps/itinerary/Itinerary.js
```diff
@@ -32,7 +32,7 @@ export type ItinerariesOneWaySearchParameters = {|
   +sort?: Sort,
   +itinerary: {|
     +origin: SearchLocation,
-    +destination?: SearchLocation,
+    +destination: SearchLocation,
```
FilipMessa: Are we use this type `ItinerariesCommonSearchParameters` somewhere?

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
```diff
@@ -27,6 +28,28 @@ export const differenceInMinutes = (
   );
 };
 
+export const mapToOpaqueID = (input: any) => {
```
FilipMessa: What we will do with this `any`?

apps/graphql/src/apps/itinerary/Itinerary.js
aalises: Because we are modifying it by casting it `fromGlobalID` and flow complained. Or maybe I can clone to object to avoid that.

apps/graphql/src/apps/itinerary/Itinerary.js
aalises: Not anymore! will remove it

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
aalises: I don't know... got Flow errors all over the place when putting it as ` ItinerariesOneWaySearchParameters | ItinerariesReturnSearchParameters`

apps/graphql/src/apps/itinerary/Itinerary.js
aalises: Done!

apps/graphql/src/apps/itinerary/Itinerary.js
aalises: Placed it back!

FilipMessa: As I can see you use the fromGlobalID with the `map` and `map` return new Array so it should not be modified.

apps/graphql/src/apps/itinerary/Itinerary.js
```diff
@@ -19,20 +19,12 @@ type SearchLocation = {|
...
-  +passengers?: Passengers,
-|};
-
 export type ItinerariesOneWaySearchParameters = {|
   +order?: Order,
   +sort?: Sort,
   +itinerary: {|
     +origin: SearchLocation,
-    +destination?: SearchLocation,
+    +destination: SearchLocation,
```
FilipMessa: The `+destination: SearchLocation,` is now required?

aalises: It was to circumvent the Flow issue with the union, but is now fixed. Placed it back!

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
FilipMessa: I'll look on it but, can you pls write tests for that?

aalises: Done!

JoseDuSV: Maybe I overlooked something but why `mapToOpaqueID` is not part of the `parseParametersNew` like other parameters parsing? Then we can map origin/destination ids directly there and just reduce this function to something like .
```
ids.map(id =>
  fromGlobalId(id),
)
```

aalises: @JoseDuSV for consistency, as in the `Bookings` query is also done on the resolver instead of the `dataLoader`. However I think we should just move both to the `dataLoaders` for simplicity, as you said

aalises: @JoseDuSV Moved everything to dataLoaders for simplicity.

JoseDuSV: Oh, I see. Booking case looks quite similar. But there is no other required parsing (for now because this whole part is currently mocked), so it quite makes sense to do it directly inside resolver instead of wrapping just `fromGlobalId` to another function inside booking dataloader.

But for search we already have the parser with solved input differences so it seems reasonable to me to make this mapping also part of it. If we don't need `mapToOpaqueID` to be reusable somewhere else like mentioned `fromGlobalId`, but according to specific input this doesn't seems to be the case.

FilipMessa: I was thinking about, and @JoseDuSV notes make sense to me.

aalises: So it has been changed 👍 way simpler this way

apps/graphql/src/apps/itinerary/helpers/__tests__/Itineraries.test.js
```diff
@@ -171,3 +172,10 @@ describe('differenceInMinutes', () => {
     expect(differenceInMinutes(null, toISO)).toBeNull();
   });
 });
+
+describe('mapToOpaqueID', () => {
+  it('returns proper formatted ID', () => {
+    const ids = ['STN', 'MGL'];
+    expect(mapToOpaqueID(ids)).toEqual(['I3', '0b']);
```
mrtnzlml: Function `fromGlobalId` accepts opaque IDs and returns unmasked IDs. Therefore your function `mapToOpaqueID` is actually doing the opposite and this test is not correct. This violation will be fixed here: https://gitlab.skypicker.com/incubator/universe/merge_requests/1390 (in version 0.8)

apps/graphql/src/apps/booking/dataloaders/BookingDetail.js
```diff
@@ -1,13 +1,16 @@
...
 import booking6676524 from '../__datasets__/booking-6676524.json';
 import booking8142828 from '../__datasets__/booking-8142828.json';
 import booking16463447 from '../__datasets__/booking-16463447.json';
 import { sanitizeBooking } from './Bookings';
 
 const BookingDetailLoader = {
   load: (id: string) => {
-    switch (id) {
+    const opaqueID = fromGlobalId(id);
+    switch (opaqueID) {
```
mrtnzlml: The same problem here - should be: `const id = fromGlobalId(opaqueID);`

apps/graphql/src/apps/itinerary/helpers/__tests__/Itineraries.test.js
aalises: Thanks @mrtnzlml !! Still getting a bit confused about the different ID types.. 👍

mrtnzlml: Basically, opaque ID (output from `@kiwicom/graphql-global-id`) represents ID from GraphQL specification (https://graphql.github.io/graphql-spec/draft/#sec-ID). The other IDs are just some strings floating through your application which I don't know much about. :)

FilipMessa:
@aalises I approved this but I was forgotten that we need update FE also.

aalises:
@FilipMessa will do it now! basically we have to convert the locations to the opaqueID on the input of the search queries.

PR #600, opened 2019-03-27
Fix: Margarita components web storybook loads universal-components co…

RobinCsl:
…rrectly

Summary: After switching the 'main' field in `package.json` to the source folder of `universal-components`, the web storybook for `margarita-components` stopped working as the files from `universal-components` were not transpiled anymore. This PR fixes that issue by transpiling files from `universal-components` when building the storybook.


aalises:
Thanks @RobinCsl ! 👍

Issue #562, opened 2019-03-19
Create the SortTabs component for web UI


FilipMessa:
The SortTabComponent will be used on the /resultList page for sorting results.

Check the kiwi.com for design

screenshot: 
<img width="722" alt="Screen Shot 2019-03-19 at 16 24 12" src="https://user-images.githubusercontent.com/6354326/54619190-7d3f5b80-4a64-11e9-90c1-88d9d676625f.png">

PR #601, opened 2019-03-27
Feat: Add SortTabs Component

aalises:
Relates to #562. Create SortTabs component for sorting the search based on quality, price or time.

### DIFF
--- a/packages/components/src/sortTabs/SortTab.js
+++ b/packages/components/src/sortTabs/SortTab.js
@@ -12,7 +12,7 @@ import {
 } from '@kiwicom/universal-components';
 
 import TouchableWithoutFeedback from '../TouchableWithoutFeedback';
-import type { SortIconTypes, SortTypes } from './SortTabs';
+import type { SortIconTypes, SortTypes, PriceDurationInfo } from './SortTabs';
 
 export type Props = {|
   +label: string,
@@ -20,6 +20,7 @@ export type Props = {|
   +isFirst: boolean,
   +isLast: boolean,
   +isActive: boolean,
+  +additionalInfo: ?PriceDurationInfo,
   +icon: SortIconTypes,
   +onPress: SortTypes => void,
 |};
@@ -47,8 +48,18 @@ export default class SortTab extends React.Component<Props, State> {
   };
 
   render() {
-    const { label, isFirst, isLast, isActive, icon } = this.props;
+    const {
+      label,
+      isFirst,
+      isLast,
+      isActive,
+      icon,
+      additionalInfo,
+    } = this.props;
     const { isHovered } = this.state;
+    const price = additionalInfo?.price;
+    const currency = additionalInfo?.currency;
+    const duration = additionalInfo?.duration;
     return (
       <Hoverable
         onMouseEnter={this.handleOnMouseEnter}
@@ -73,16 +84,23 @@ export default class SortTab extends React.Component<Props, State> {
                 {label}
               </Text>
               <View style={styles.priceDurationContainer}>
-                <Text style={styles.label} align="left" numberOfLines={1}>
-                  1,394 Kč
-                </Text>
-                <Text
-                  style={styles.durationLabel}
-                  align="left"
-                  numberOfLines={1}
-                >
-                  9h 3m
-                </Text>
+                {price || currency ? (
+                  <Text style={styles.label} align="left" numberOfLines={1}>
+                    {price} {currency}
+                  </Text>
+                ) : null}
+                {duration ? (
+                  <Text
+                    style={[
+                      styles.durationLabel,
+                      (price || currency) && styles.marginStartLabel,
+                    ]}
+                    align="left"
+                    numberOfLines={1}
+                  >
+                    {duration}
+                  </Text>
+                ) : null}
               </View>
             </View>
             <Icon
@@ -102,8 +120,10 @@ export default class SortTab extends React.Component<Props, State> {
 }
 
 const borderRadius = parseInt(defaultTokens.borderRadiusSmall, 10);
+const containerHeight = 75;
 const styles = StyleSheet.create({
   container: {
+    height: containerHeight,
     flex: 1,
     flexDirection: 'row',
     justifyContent: 'space-between',
@@ -139,11 +159,12 @@ const styles = StyleSheet.create({
   durationLabel: {
     color: defaultTokens.paletteInkLight,
     fontWeight: 'normal',
-    marginStart: parseInt(defaultTokens.paddingButtonSmall, 10),
     fontSize: parseInt(defaultTokens.fontSizeTextSmall, 10),
   },
+  marginStartLabel: {
+    marginStart: parseInt(defaultTokens.paddingButtonSmall, 10),
+  },
   label: {
-    flex: 1,
     color: defaultTokens.paletteInkDark,
     fontWeight: designTokens.fontWeightMedium,
     fontSize: parseInt(defaultTokens.fontSizeTextSmall, 10),
--- a/packages/components/src/sortTabs/SortTab.js
+++ b/packages/components/src/sortTabs/SortTab.js
@@ -1,0 +1,153 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import {
+  Text,
+  StyleSheet,
+  Icon,
+  Hoverable,
+} from '@kiwicom/universal-components';
+
+import TouchableWithoutFeedback from '../TouchableWithoutFeedback';
+import type { SortIconTypes } from './SortTabs';
+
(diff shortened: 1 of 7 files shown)


aalises:
First iteration is done. For now the price and duration are hardcoded, but in the future we should pass them as props and obtain the information from the query `(price,duration,currency)`
Screenshot 2019-03-28 at 10 35 38

jvaclavik:
> First iteration is done. For now the price and duration are hardcoded, but in the future we should pass them as props and obtain the information from the query `(price,duration,currency)`. Or maybe we could include now the props and mock the data for now, WDYT?
> Screenshot 2019-03-28 at 10 35 38

From my perspective mocking and passing data to props is a better way

packages/components/src/sortTabs/SortTab.js
```diff
@@ -0,0 +1,153 @@
...
+  Hoverable,
+} from '@kiwicom/universal-components';
+
+import TouchableWithoutFeedback from '../TouchableWithoutFeedback';
+import type { SortIconTypes, SortTypes } from './SortTabs';
+
+export type Props = {|
+  +label: string,
+  +value: SortTypes,
+  +first: boolean,
```
jvaclavik: What about use `isFirst, isLast, isActive`?

packages/components/src/sortTabs/SortTab.js
```diff
@@ -0,0 +1,153 @@
...
+  +last: boolean,
+  +active: boolean,
+  +icon: SortIconTypes,
+  +onPress: (value: SortTypes) => void,
```
jvaclavik: You don't need to write here `value:`. So 
```
+onPress: SortTypes => void,
```
would be enough

FilipMessa:
The code looks nice, probably this mocked data for price, duration, currency should be optional and passed by props, because I checked API response and there missing data yet for that.

just fond one issue broken layout on IOS:
Screen Shot 2019-03-28 at 11 29 34

packages/components/src/sortTabs/SortTab.js
FilipMessa: good point boolean value should be always prefixed with `is...`

jvaclavik: I'm not sure what to do about `hovered`. Maybe is could be `isHovered`?

packages/components/src/sortTabs/SortTab.js
```diff
@@ -0,0 +1,153 @@
...
+            <View>
+              <Text
+                style={[styles.label, active && styles.labelActive]}
+                align="left"
+                numberOfLines={1}
+              >
+                {label}
+              </Text>
+              <View style={styles.priceDurationContainer}>
+                <Text style={[styles.label]} align="left" numberOfLines={1}>
```
jvaclavik: You don't need array here

packages/components/src/sortTabs/SortTab.js
```diff
@@ -0,0 +1,153 @@
...
+                  1,394 Kč
+                </Text>
+                <Text
+                  style={[styles.durationLabel]}
```
jvaclavik: You don't need array here

packages/components/src/sortTabs/SortTabs.js
```diff
@@ -0,0 +1,57 @@
...
+|};
+
+export const SortTabsData: SortTabsDataType = [
+  { label: 'Best', value: 'QUALITY', icon: 'thumb-up' },
+  { label: 'Cheapest', value: 'PRICE', icon: 'money' },
+  { label: 'Fastest', value: 'DURATION', icon: 'clock' },
+];
+
+const SortTabs = ({ selectedValue, onValueChange }: Props) => {
+  const SortTabs = SortTabsData.map((SortTabData, index) => {
```
jvaclavik: Can you rename `SortTabs`? There are two same names and it's not clear.

packages/components/src/sortTabs/SortTab.js
```diff
@@ -0,0 +1,153 @@
...
+  durationLabel: {
+    color: defaultTokens.paletteInkLight,
+    fontWeight: 'normal',
+    marginStart: parseInt(defaultTokens.paddingButtonSmall, 10),
+    fontSize: parseInt(defaultTokens.fontSizeTextSmall, 10),
+  },
+  label: {
+    flex: 1,
+    color: defaultTokens.paletteInkDark,
+    fontWeight: '500',
```
JoseDuSV: You can use `designTokens.fontWeightMedium` from `@kiwicom/orbit-design-tokens` for this one.

packages/components/src/sortTabs/SortTab.js
aalises: Done!

packages/components/src/sortTabs/SortTab.js
aalises: Done!

packages/components/src/sortTabs/SortTab.js
aalises: Done!

packages/components/src/sortTabs/SortTab.js
aalises: Changed everything to be prefixed with `is...`

packages/components/src/sortTabs/SortTab.js
aalises: Done!

packages/components/src/sortTabs/SortTabs.js
aalises: Done! renamed to `SortTabsList`

aalises:
@FilipMessa @jvaclavik Now the price duration stuff is passed as a prop, everything is optional. The workaround for the iOS stuff and no params has been adding `minHeight`, Probably there is a better way to do it?

packages/components/src/sortTabs/SortTab.js
```diff
@@ -0,0 +1,169 @@
...
+            <View>
+              <Text
+                style={[styles.label, isActive && styles.labelActive]}
+                align="left"
+                numberOfLines={1}
+              >
+                {label}
+              </Text>
+              <View style={styles.priceDurationContainer}>
+                <Text style={styles.label} align="left" numberOfLines={1}>
```
FilipMessa: I don't know maybe I am too conservative but when there is no data for display I would expect not to render empty container and text or whatever if it not necessary. 

What do you think? @jvaclavik @JoseDuSV

aalises: I agree Filip, I also thought of this but added lots of conditions. I think it's better not to render the empty container :D will change it.

packages/components/src/sortTabs/SortTab.js
```diff
@@ -0,0 +1,174 @@
...
+import TouchableWithoutFeedback from '../TouchableWithoutFeedback';
+import type { SortIconTypes, SortTypes, PriceDurationInfo } from './SortTabs';
+
+export type Props = {|
+  +label: string,
+  +value: SortTypes,
+  +isFirst: boolean,
+  +isLast: boolean,
+  +isActive: boolean,
+  +priceDurationInfo: ?PriceDurationInfo,
```
FilipMessa: What about `additionalInfo` instead of `priceDurationInfo`? 
Just Idea

packages/components/src/sortTabs/SortTab.js
```diff
@@ -0,0 +1,174 @@
...
+            <View>
+              <Text
+                style={[styles.label, isActive && styles.labelActive]}
+                align="left"
+                numberOfLines={1}
+              >
+                {label}
+              </Text>
+              <View style={styles.priceDurationContainer}>
+                {priceDurationInfo?.price || priceDurationInfo?.currency ? (
```
FilipMessa: Thinking if that will be more readable and have better performance with deconstruction?
`const {currency, price, duration} = priceDurationInfo`

packages/components/src/sortTabs/SortTab.js
JoseDuSV: For me it depends on situation. Sometimes it's ok to keep component there if it's for example somehow affecting rest of the layout. But in this case using conditions makes more sense to me.

packages/components/src/sortTabs/SortTab.js
aalises: Done!

packages/components/src/sortTabs/SortTab.js
aalises: Yes I thought of that but with them being potentially null, Flow complains 👎

FilipMessa: 🤔 right
so maby with this it will pass
```

    const price = priceDurationInfo?.price;
    const currency = priceDurationInfo?.currency;
    const duration = priceDurationInfo?.duration
```

aalises: Changed!

PR #602, opened 2019-03-27
Feat: Added Firebase login

JoseDuSV:
Summary: Implemented Firebase Google sign-in method for both platforms.

The idea is to have Firebase part optional, to prevent making initial project setup more complex (by adding necessity to create Firebase project, obtaining config and getting app credentials for both platforms).
Without Firebase setup the app will work normally like for anonymous user, just with disabled log-in part/button.

Note: **To make login work, don't forget to update your `.env` variables from 1Password**


.env-sample
```diff
@@ -1 +1,11 @@
...
+
+FIREBASE_API_KEY=
+FIREBASE_AUTH_DOMAIN=
+FIREBASE_DATABASE_URL=
+FIREBASE_PROJECT_ID=
+FIREBASE_STORAGE_BUCKET=
+FIREBASE_MESSAGING_SENDER_ID=
+
+GOOGLE_OAUTH_IOS=
+GOOGLE_OAUTH_ANDROID=
```
tbergq: You are aware that all of these values will be available for anyone who downloads the app and reverse-engineers it

apps/core/components/userContext/Firebase.js
```diff
@@ -0,0 +1,27 @@
...
+  apiKey: FIREBASE_API_KEY,
+  authDomain: FIREBASE_AUTH_DOMAIN,
+  databaseURL: FIREBASE_DATABASE_URL,
+  projectId: FIREBASE_PROJECT_ID,
+  storageBucket: FIREBASE_STORAGE_BUCKET,
+  messagingSenderId: FIREBASE_MESSAGING_SENDER_ID,
+};
+
+if (!firebase.apps.length) {
+  if (FIREBASE_API_KEY.length > 0) {
```
tbergq: if (!firebase.apps.length &&  FIREBASE_API_KEY.length > 0)

apps/core/components/userContext/SignInHelpers.native.js
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * NOTE: Direct Firebase login is not available for RN.
+ * We are using `expo` Goole login to obtain user token.
+ * And then passing it to Firebase `GoogleAuthProvider` to log user in.
+ */
+
+export const signInWithGoogle = () => {
+  return new Promise<void>((resolve, reject) => {
+    Google.logInAsync({
```
tbergq: it seems this already is returning a promise. Why not do 
`export const signInWithGoogle = async () => {`

apps/core/components/userContext/SignInHelpers.web.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+import firebase from 'firebase/app';
+
+export const signInWithGoogle = () => {
+  return new Promise<void>((resolve, reject) => {
```
tbergq: same here

.env-sample
JoseDuSV: Yes, these Firebase values should be public, it is even easier to obtain it on web.
Where something like this is suggested (which can look even scarier):
Screenshot 2019-04-01 at 16 16 24

But you can/should then restrict login to certain domain/s (default is just localhost) in Firebase administration. And also set database rules to restrict what and by who can be written and read.

tbergq: Ok, it just sounds scary to provide apiKey to everyone, but if you are sure that it is not a problem...

RobinCsl: You can close down the access in the Google Console to say that the API key can only be used from certain origins, etc. so I think that's why it's okay, even though you should still take measures to restrict access to the key.

JoseDuSV:
As we discussed yesterday, this one should be ready to merge.

As noted: **Don't forget to update your `.env` variables from 1Password in your local versions.**

apps/core/components/userContext/UserContext.js
```diff
@@ -23,9 +41,58 @@ export default class UserContextProvider extends React.Component<Props, State> {
...
+    });
+  };
+
+  signOut = () => {
+    firebase
+      .auth()
+      .signOut()
+      .catch(() => {
+        // @TODO Handle sign-out error - show notification or message
+      });
```
tbergq: signIn and signOut looks very much like promises. Are they? And if so, can we use async/await instead?

.env-sample
RobinCsl: Just one thing here: there is no newline at the end of the file and it can be problematic. I know that in the monorepo, they are careful with this.
It makes some common tools for counting lines in files, etc. not working as intended: https://thoughtbot.com/blog/no-newline-at-end-of-file

apps/core/components/userContext/Firebase.js
```diff
@@ -0,0 +1,25 @@
...
+const CONFIG = {
+  apiKey: FIREBASE_API_KEY,
+  authDomain: FIREBASE_AUTH_DOMAIN,
+  databaseURL: FIREBASE_DATABASE_URL,
+  projectId: FIREBASE_PROJECT_ID,
+  storageBucket: FIREBASE_STORAGE_BUCKET,
+  messagingSenderId: FIREBASE_MESSAGING_SENDER_ID,
+};
+
+if (!firebase.apps.length && FIREBASE_API_KEY.length > 0) {
```
RobinCsl: Not sure about this condition being necessary.

apps/core/components/userContext/Firebase.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+
+import firebase from 'firebase/app';
```
RobinCsl: In my other project, I only have `import firebase from '@firebase/app';` instead of the two imports, and note as well the @ prefix. Not sure how relevant it is.

apps/core/components/userContext/SignInHelpers.native.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+
+import { Google } from 'expo';
+import { Platform } from 'react-native';
+import firebase from 'firebase/app';
+import { GOOGLE_OAUTH_ANDROID, GOOGLE_OAUTH_IOS } from 'react-native-dotenv';
+
+/**
+ * NOTE: Direct Firebase login is not available for RN.
```
RobinCsl: Is that because of the `popup` thing on web?

RobinCsl: (In which case it should be the same on mobile and web I believe)

apps/core/components/userContext/UserContext.js
```diff
@@ -1,18 +1,36 @@
...
+  userEmail: ?string,
+  userDisplayName: ?string,
+  +actions: {|
+    +signIn: () => void,
+    +signOut: () => void,
+  |},
 |};
 
 const defaultState = {
+  isFirebaseAvailable: firebase.apps.length > 0,
```
RobinCsl: I'm kind of confused by this.

apps/core/components/userContext/UserContext.js
```diff
@@ -23,9 +41,58 @@ export default class UserContextProvider extends React.Component<Props, State> {
...
+        });
+      });
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.fireBaseListener) {
+      /**
+       * NOTE: Auth listener is removed by calling the unsubscribe
+       * function returned by `onAuthStateChanged`
```
RobinCsl: Maybe add link to Firebase docs for easier reference.

docs/guide-firebase-user.md
```diff
@@ -0,0 +1,38 @@
...
+sidebar_label: Guide: Firebase implementation
+---
+
+Firebase is used for sample implementation of _**user login**_ and _**data storing**_.
+
+---
+
+**Note:** This part is completely _**optional**_ and can be replaced with any other solution of your choice.
+
+_**Firebase**_ and _**Google login**_ related `.env` variables can be left _**empty**_. In this case the app will operate normally, only the user login part will be _**disabled**_.
```
RobinCsl: So this is why you're checking for the length of firebase.apps?

apps/core/components/userContext/Firebase.js
JoseDuSV: But just `import firebase from 'firebase/app';` is not enough. If I want to use `auth`, I need to import it as well.

The other option is to use `import * as firebase from 'firebase';` and import everything, but it is not recommended and you'll get console warning:
`It looks like you're using the development build of the Firebase JS SDK.
When deploying Firebase apps to production, it is advisable to only import
the individual SDK components you intend to use.`

And what exactly you mean by `@ prefix`? I don't see any `@` laying around this file

RobinCsl: >  If I want to use auth, I need to import it as well.
I think you need to import it in the files where you use `firebase.auth(). ...`, but in here you're only initializing the app, so I'm not sure it's necessary.

The `@` prefix was actually a mistake on my part which I will need to fix in my project, the `@firebase/app` package is not meant to be used on its own. 

>This package is not intended for direct usage, and should only be used via the officially supported firebase package.

cf https://www.npmjs.com/package/@firebase/app

.env-sample
JoseDuSV: Thanks, good to know. Extra line added.

apps/core/components/userContext/UserContext.js
JoseDuSV: Sure, link added.

apps/core/components/userContext/UserContext.js
JoseDuSV: To make it clear I've added short description and link to our docs for more info.

docs/guide-firebase-user.md
JoseDuSV: Exactly. As mentioned in other comment, I've linked this description in `isFirebaseAvailable` notes to make it clearer.

apps/core/components/userContext/SignInHelpers.native.js
JoseDuSV: Yes, we cannot use firebase `signInWithPopup` function directly in RN. So we are using expo Google login for this part.

And ss we briefly discussed yesterday. We want to start with just Google login, but we can definitely add some `Register` screen and allow account creation later.

apps/core/components/userContext/Firebase.js
JoseDuSV: True, it's not really clear at first glance. I've added short description to this condition to make clear why it is necessary.

apps/core/components/userContext/UserContext.js
JoseDuSV: Sure, updated. I'll probably even move whole try...catch part directly to SignIn/SignOut component to handle UI changes during login. But I'm saving this for next PR to keep this one smaller.

JoseDuSV:
@RobinCsl @tbergq Can you check my updates if it resolves your comments, so we can finally merge this one. Thanks a lot

apps/core/components/userContext/SignInHelpers.web.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+
+import firebase from 'firebase/app';
+
+export const signInWithGoogle = async () => {
+  const provider = new firebase.auth.GoogleAuthProvider();
```
aalises: Is localization taken into account here? I don't know if that's is something that has to be taken into account or we just use the default. To use the device locale we could also set it with `firebase.auth().useDeviceLanguage()`

JoseDuSV:
@aalises I just updated description slightly and force-pushed it again. Netlify seems to be happy this time, so I think we can merge it. Thanks 🚀

FilipMessa:
Looks good to me, just one question do we need optional firebase for FE? What is the background for this decision?

JoseDuSV:
> Looks good to me, just one question do we need optional firebase for FE? What is the background for this decision?

@FilipMessa It's optional mainly to make initial setup easier and do not force developers to create Firebase app to start the project. This way you can try it quicker and fill-in remaining `.env` vars later like mentioned in updated docs. But if it's not really clear, I can try to review description and update it. Maybe just let me know where you will also expect this information if not in new Firebase related section.

FilipMessa:
@JoseDuSV 
OK, I understand. I know that we had a plan to make it optional in GraphQL but I missed idea behind the why we want to do this on the FE. But I got it now thanks.

PR #604, opened 2019-03-28
Add Slug field to search queries location 

### LINKED ISSUE #603: Add slug field into the LocationItineraryInput GraphQL schema
expected schema
```
input LocationItineraryInput {
  ids: [String]
  slug: [String]
}
```

aalises:
Add the slug field to the location so you can search by ids or slugs. Closes #603

### DIFF
--- a/apps/graphql/src/apps/itinerary/Itinerary.js
+++ b/apps/graphql/src/apps/itinerary/Itinerary.js
@@ -15,7 +15,7 @@ type Passengers = {|
   +infants?: number,
 |};
 
-type SearchLocation = {|
+export type SearchLocation = {|
   +ids?: string[],
   +slugs?: string[],
 |};
--- a/apps/graphql/src/apps/itinerary/Itinerary.js
+++ b/apps/graphql/src/apps/itinerary/Itinerary.js
@@ -16,7 +16,8 @@ type Passengers = {|
 |};
 
 type SearchLocation = {|
-  +ids: string[],
+  +ids?: string[],
+  +slugs?: string[],
 |};
 
 export type ItinerariesOneWaySearchParameters = {|
--- a/apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
+++ b/apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
@@ -35,7 +35,7 @@ export const parseParameters = (
     ? input.itinerary.inboundDate
     : null;
 
-  const flyFrom = unmaskID(origin.ids).join();
+  const flyFrom = origin.ids && unmaskID(origin.ids).join();
   const flyTo =
     destination && destination.ids ? unmaskID(destination.ids).join() : null;
 
--- a/apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
+++ b/apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
@@ -12,14 +12,20 @@ import {
   mapDate,
   getItineraryType,
   mapSectors,
-  unmaskID,
+  slugsNeedParsing,
+  unmaskReplaceIds,
 } from '../helpers/Itineraries';
 import type {
   ItinerariesReturnSearchParameters,
   ItinerariesOneWaySearchParameters,
-  ApiResponseType,
+  ApiResponseType as ApiResponseSearch,
   Itinerary,
+  SearchLocation,
 } from '../Itinerary';
+import type {
+  ApiResponse as ApiResponseLocation,
+  ApiLocation,
+} from '../../location/Location';
 
 const stripTimeZoneOffset = (date: Date) =>
   DateFNS.addMinutes(date, date.getTimezoneOffset());
@@ -35,9 +41,9 @@ export const parseParameters = (
     ? input.itinerary.inboundDate
     : null;
 
-  const flyFrom = origin.ids && unmaskID(origin.ids).join();
-  const flyTo =
-    destination && destination.ids ? unmaskID(destination.ids).join() : null;
+  const destinationIds = destination?.ids;
+  const flyFrom = (origin.ids ?? []).join();
+  const flyTo = destinationIds !== null ? (destinationIds ?? []).join() : null;
 
   const commonSearchParams = {
     fly_from: flyFrom,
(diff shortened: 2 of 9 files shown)


apps/graphql/src/apps/itinerary/dataloaders/__tests__/Itineraries.test.js
```diff
@@ -13,10 +13,10 @@ const searchParamsOneWay = {
   },
   itinerary: {
     origin: {
-      ids: ['TG9jYXRpb246cHJhZ3VlX2N6'],
```
tbergq: Why has this changed?

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
```diff
@@ -36,6 +41,17 @@ export const mapVehicle = (type: ?string, uniqueNo: ?string) => ({
   uniqueNo: uniqueNo ?? null,
 });
 
+export const slugsNeedParsing = (location: ?SearchLocation): boolean => {
+  return !!(!location?.ids && location?.slugs);
```
tbergq: This is really hard to read, could it rather be something like 
`return location?.ids == null && location?.slugs != null`?

apps/graphql/src/apps/itinerary/dataloaders/__tests__/Itineraries.test.js
aalises: @tbergq Before, the unmasking of the IDs was done on the `parseParameters`. 

For simplicity we moved it out of the `parseParameters` to the resolvers (so you can unmask the IDs only in the case they are opaque i.e when inputting the IDs, but `Tequila Locations API` when searching for `slugs` gives you back an unmasked ID so we don't always unmask them, hence the reason of doing the conditional unmasking on the resolver rather than on the parseParameters). 

So now the input of the of the `parseParameters` are always unmasked IDs.

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
aalises: Done!

apps/graphql/src/apps/itinerary/dataloaders/__tests__/Itineraries.test.js
tbergq: Ok, got it 🙂

apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
```diff
@@ -35,9 +41,8 @@ export const parseParameters = (
     ? input.itinerary.inboundDate
     : null;
 
-  const flyFrom = unmaskID(origin.ids).join();
-  const flyTo =
-    destination && destination.ids ? unmaskID(destination.ids).join() : null;
+  const flyFrom = origin.ids && origin.ids.join();
+  const flyTo = destination && destination.ids ? destination.ids.join() : null;
```
tbergq: Also, I think we should use optional chaining and nullish coaleshing, so something like: 

```
const destinationIds = destination?.ids;

const flyFrom = (origin.ids ?? []).join();
const flyTo = destinationIds != null ? destinationIds.join() : null;
```

apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
```diff
@@ -71,11 +76,41 @@ const addReturnSearchQueryParams = inboundDate => {
   };
 };
 
+const parseSlugsToIds = async (
+  location: SearchLocation,
+): Promise<SearchLocation> => {
+  if (!location.slugs) return location;
+
+  const results: $ReadOnlyArray<ApiResponseLocation> = await Promise.all(
+    location.slugs.map(async slug => fetch(`/locations/slug?term=${slug}`)),
```
tbergq: this async keyword is not necessary, since fetch already returns a promise

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
```diff
@@ -36,6 +41,17 @@ export const mapVehicle = (type: ?string, uniqueNo: ?string) => ({
...
 });
 
+export const slugsNeedParsing = (location: ?SearchLocation): boolean => {
+  return location?.ids == null && location?.slugs != null;
+};
+
+export const unmaskReplaceIds = (location: ?SearchLocation): SearchLocation => {
+  return {
+    ...location,
+    ...(location && location.ids && { ids: unmaskID(location.ids) }),
```
tbergq: You could also use nullish coalescing and optional chaining here

apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
aalises: Done!

apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
aalises: Done!

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
aalises: Done! Usually when I am fiddling with those I get some Flow infer errors, still have to get used to that.

tbergq: Yeah, we are early adopters, so there still might be some things that will improve. 

😄

apps/graphql/src/apps/itinerary/Itinerary.js
```diff
@@ -15,16 +15,17 @@ type Passengers = {|
...
 |};
 
 export type ItinerariesOneWaySearchParameters = {|
   +order?: Order,
   +sort?: Sort,
   +itinerary: {|
-    +origin: SearchLocation,
-    +destination?: SearchLocation,
+    origin: SearchLocation,
+    destination?: SearchLocation,
```
tbergq: Why is this not read-only?

apps/graphql/src/apps/itinerary/Itinerary.js
```diff
@@ -34,8 +35,8 @@ export type ItinerariesReturnSearchParameters = {|
...
```
tbergq: same here?

apps/graphql/src/apps/itinerary/Itinerary.js
aalises: Because we are reassigning them to parse the IDs on `fetchItineraries` 

```
params.itinerary.origin = slugsNeedParsing(origin)
        ? await parseSlugsToIds(origin)
        : unmaskReplaceIds(origin);
```
If I put it into different variables and then compose them to pass them to `parseParameters` get a lot of weird Flow errors...

apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
```diff
@@ -71,20 +77,59 @@ const addReturnSearchQueryParams = inboundDate => {
...
+      let { origin, destination } = params.itinerary;
+
+      origin = slugsNeedParsing(origin)
+        ? await parseSlugsToIds(origin)
+        : unmaskReplaceIds(origin);
+
+      destination =
+        destination && slugsNeedParsing(destination)
+          ? await parseSlugsToIds(destination)
+          : unmaskReplaceIds(destination);
```
tbergq: Ok, it is better, the only thing I don't like is that you still overwrite the original object.

Please do `const parsedOrigin` and `const parsedDestination` instead. Then the PR should be ready to be merged.

aalises: Changed! Yes it's a little bit ugly, it would be much nicer if the `Tequila API` returned also the opaque IDs so we could move it back to `parseParameters` and not have to do this ID tweaking...

tbergq: Well, I still don't understand the use case for this, why do we need this query?

apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
```diff
@@ -71,20 +77,59 @@ const addReturnSearchQueryParams = inboundDate => {
   };
 };
 
+const parseSlugsToIds = async (
```
tbergq: Also, just realising that this should be done in a locations dataloader, so we can optimise this data-fetching. 
Now we have no caching for fetching these locations, and we could end up hitting the rest-api again when it is not necessary.

apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
JoseDuSV: @tbergq We just talked about it with Albert and this part is not longer relevant. After ve merged everything to `RouteStop` in #606 we don't need this extra parsing and everything (slug included) is solved by Location query now.

aalises:
Closing this because as discussed with @JoseDuSV  and @tbergq , this is no longer relevant.

PR #606, opened 2019-03-29
Refactor: Duplicate types in Itinerary and Booking

JoseDuSV:
Summary:
- Merged Itinerary `endTime + destination` & `startTime + origin` to `arrival` & `departure` RouteStop types to make schema more consistent with bookings.
- Moved duplicate Itinerary and Booking flow types to `common` folder.

Before
Screenshot 2019-03-29 at 17 14 30

After
Screenshot 2019-03-29 at 17 14 49

### DIFF
--- a/apps/graphql/src/apps/common/CommonTypes.js
+++ b/apps/graphql/src/apps/common/CommonTypes.js
@@ -1,0 +1,43 @@
+// @flow
+
+export type DateType = {|
+  +local: ?(string | number),
+  +utc: ?(string | number),
+|};
+
+export type Price = {|
+  +amount: ?number,
+  +currency: ?string,
+|};
+
+export type RouteStop = {|
+  +cityId?: ?string,
+  +time: ?DateType,
+  +code: ?string,
+|};
+
+export type Sector = {|
+  +duration?: ?number,
+  +segments: ?Array<Segment>,
+  +stopoverDuration?: ?number,
+  +departure: ?RouteStop,
+  +arrival: ?RouteStop,
+|};
+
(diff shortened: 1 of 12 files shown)


apps/graphql/src/apps/common/CommonTypes.js
```diff
@@ -0,0 +1,43 @@
+// @flow
+
+export type DateType = {|
+  +local: ?(string | number),
```
jvaclavik: I know it was here before, but I think we shouldn't do it this way. I had problems with types like this one because `format` from new `date-fns` can't accept number and you have to call different function for it. And even without this issue it's still confusing. 

What do you think? I would propose to create new issue for it.

tbergq: The issue is that we get date as string from Tequila, and mocked data for MMB has it as int.

JoseDuSV: @jvaclavik I totally agree and I also tried to update it, but then I find out what Trond mentioned and don't want this PR to grow too big with other date related necessary changes.
But we can definitely create new issue for it.

jvaclavik: > The issue is that we get date as string from Tequila, and mocked data for MMB has it as int.

I understand the reason, but according to me those types should be separated.

tbergq: Well, they will be passed to the same resolver, so that is why it has this type. 

Maybe we should just wrap both sanitizers with new Date(local)

PR #610, opened 2019-04-01
Add size prop to Button component

### LINKED ISSUE #590: [Button] Add size prop like on Orbit components
The button currently has a fixed height of 44; I would need to have the same size props as on Orbit components: https://kiwicom.github.io/orbit-components/?selectedKind=Button&selectedStory=Playground&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs

KarinaDavtyan:
Summary: change button and icon height, fontSize and paddings/margins depending on the size prop passed to the component

Closes #590


packages/universal-components/src/Button/Button.stories.js
```diff
@@ -37,20 +37,21 @@ storiesOf('Button', module)
       ],
       'primary',
     );
+    const size = select('Size', ['small', 'normal', 'large'], 'normal');
     const disabled = boolean('Disabled', false);
     let sublabel;
     if (Platform.OS !== 'web') {
-      sublabel = text('Sublabel', 'Sublabel');
```
RobinCsl: Reason for this change?

packages/universal-components/storybook/preview-head.html
```diff
@@ -1,11 +1,14 @@
-<link href="https://fonts.googleapis.com/css?family=Roboto:400,400i,500,500i,700" rel="stylesheet">
+<link
```
RobinCsl: Why is this file included?

packages/universal-components/src/Button/Button.stories.js
KarinaDavtyan: For faster development, not to remove it after each change. Also changed icons. Should I change it back?

packages/universal-components/storybook/preview-head.html
KarinaDavtyan: Because prettier kept insisting on changing it, even if I reverted the changes it was showing an error in vscode

JoseDuSV: Are you using some kind of Prettier plugin? In our project it shouldn't care about `html` files because we don't use any parser for it.

KarinaDavtyan: I am using Prettier - Code formatter.Apparently it was using its own settings for html but now disabled it for html files

packages/universal-components/src/Button/__tests__/__snapshots__/index.test.js.snap
```diff
@@ -32,22 +32,31 @@ exports[`Button - native should match snapshot diff 1`] = `
 +             \\"justifyContent\\": \\"space-between\\",
             },
             false,
+            Object {
+              \\"height\\": 44,
+-             \\"paddingEnd\\": 16,
+-             \\"paddingStart\\": 16,
++             \\"paddingEnd\\": 12,
++             \\"paddingStart\\": 12,
```
JoseDuSV: This is ok and you can remove your note from description. It's difference snapshot. So padding is not added twice but it is simply just difference between button with and without icons.

KarinaDavtyan: Thank you for explanation! It was so confusing

packages/universal-components/src/Button/Button.stories.js
JoseDuSV: I like when initial Playground state shows all component possibilities. So I'm for setting it back. But we can keep new variable icons 😉. Thanks

KarinaDavtyan: Its back :)

packages/universal-components/src/Button/styles/shared.js
```diff
@@ -19,3 +19,54 @@ export const wrapperColor = {
...
+
+export const size = {
+  small: {
+    height: parseInt(defaultTokens.heightButtonSmall, 10),
+    fontSize: parseInt(defaultTokens.fontSizeButtonSmall, 10),
+    paddingStart: {
+      paddingButton: parseInt(defaultTokens.spaceSmall, 10),
+      paddingButtonWithIcons: parseInt(defaultTokens.spaceXSmall, 10),
+      paddingButtonWithRightIcon: parseInt(defaultTokens.spaceSmall, 10),
+      paddingButtonWithLeftIcon: parseInt(defaultTokens.spaceMedium, 10),
```
JoseDuSV: Is this value correct? It looks bigger than it should be.

Screenshot 2019-04-02 at 11 12 48

packages/universal-components/src/Button/styles/shared.js
```diff
@@ -19,3 +19,54 @@ export const wrapperColor = {
...
+      paddingButtonWithIcons: parseInt(defaultTokens.spaceXSmall, 10),
+      paddingButtonWithRightIcon: parseInt(defaultTokens.spaceSmall, 10),
+      paddingButtonWithLeftIcon: parseInt(defaultTokens.spaceSmall, 10),
+    },
+  },
+  normal: {
+    height: parseInt(defaultTokens.heightButtonNormal, 10),
+    fontSize: parseInt(defaultTokens.fontSizeButtonNormal, 10),
+    paddingStart: {
+      paddingButton: parseInt(defaultTokens.spaceMedium, 10),
```
JoseDuSV: It looks like there are multiple repeating values. Cannot we simplify it just be setting something like `padding` and `paddingWithIcon` constants for each size and then just use it on each side according to presence of icon.
Or is there a case when for example left padding can be different with just left icon and with both. Now I see this case only for small button, but according to my other comment I'm not sure if it's intentional.

packages/universal-components/src/Button/ButtonInner.js
```diff
@@ -20,8 +21,26 @@ type Props = {|
...
 |};
 
+const paddingTokens = {
+  paddingButton: 'paddingButton',
+  paddingButtonWithIcons: 'paddingButtonWithIcons',
+  paddingButtonWithLeftIcon: 'paddingButtonWithLeftIcon',
+  paddingButtonWithRightIcon: 'paddingButtonWithRightIcon',
+};
+
+const getPadding = (leftIcon, rightIcon) => {
```
JoseDuSV: It's not returning actual padding so something like `getPaddingToken` or  `getPaddingType` will be probably more suitable.

packages/universal-components/src/Button/ButtonInner.js
```diff
@@ -52,29 +79,37 @@ export default function ButtonInner({
     testID = `ButtonInner${testIDProps}`;
   }
 
+  const padding = getPadding(leftIcon, rightIcon);
```
JoseDuSV: It was slightly confusing for me at first glance. Because I was expecting actual padding value.
Probably something like `paddingType/Token` can be more suitable.

packages/universal-components/src/Button/styles/shared.js
KarinaDavtyan: It should be, I was taking them from design-tokens and referencing orbit button values
Which value exactly it is?

packages/universal-components/src/Button/styles/shared.js
KarinaDavtyan: Yes, the small button values werent right. Just updated them.
Unfortunately there are no repeating values
Would it be a good idea ask about RN values in `plz-orbit`? Wasnt it discussed before, by any chance, dont want to duplicate question.
 It took the way too much time converting padding values into RN one by one😩

packages/universal-components/src/Button/styles/shared.js
KarinaDavtyan: Was it as small button?

packages/universal-components/src/Button/ButtonInner.js
KarinaDavtyan: Sorry for poor naming, will change it now

KarinaDavtyan: Done!

packages/universal-components/src/Button/styles/shared.js
JoseDuSV: Yes, it is small button. It just looked suspicious to me that there are just two size variants for other sizes and small is mixing three values. 
Left padding in Orbit storybook looks smaller for similar case.
Screenshot 2019-04-02 at 12 34 31

KarinaDavtyan: Should be good now, after correcting values

packages/universal-components/src/Button/styles/shared.js
JoseDuSV: But after update repeating values pattern is similar also for small size. So I think it can be simplified and we don't need four tokens but just setting padding according to icon presence should be fine.
I've quickly tried to simplify it a bit, you can check `feat/button-size-prop-icons-padding` branch and let me know what do you think about this approach and if there can be any possible drawbacks. 

I totally understand you pain with design tokens. But adding RN values to Orbit is still ongoing discussion. Current state is that if the value is present in tokens we should use it (with that annoying repeating parsing if it's in pixels). And if there is some repeating value missing from tokens, you can add it to `@kiwicom/universal-components/DesignTokens`.

KarinaDavtyan: I understand what you meant now, with this approach it looks much better

KarinaDavtyan: How could we apply these changes here, or are you planning to do PR after merging this one?

JoseDuSV: Good question. Lets just merge this one and I'll create new small PR with update and let you know about code review.

PR #613, opened 2019-04-02
Upgrade @kiwicom/eslint to latest version

tbergq:


aalises:
Maybe I am missing something, but shouldn't we update the dependency as well somewhere? Or where are we getting the latest version.

tbergq:
Yeah, that probably explains why the lint failed on the CI, I didn't even notice, seems my package.json was not committed 😱

tbergq:
@aalises 🚀

PR #616, opened 2019-04-02
Fix: Unify names for search trip types (OneWay, Return)

### LINKED ISSUE #615: Consolidate naming of oneway, OneWay, oneWay
I noticed that we are mixing up some naming (`OneWay`, `oneway`, `oneWay`). We should keep it consistent. I'm just not sure which form would be the best one.

Screenshot 2019-04-02 at 11 21 47
Screenshot 2019-04-02 at 11 21 32
Screenshot 2019-04-02 at 11 21 15

Similar problem is also with `return` and `Return`

aalises:
Unified names to `OneWay`, `Return` across all the app. Closes #615


apps/core/components/fromTo/__tests__/FromToIcon.test.js
```diff
@@ -22,14 +22,14 @@ it('show correct icons', () => {
...
-  const oneWay = render(
+  const OneWay = render(
     // $FlowExpectedError: It should be ok to pass string
     <FromToIcon data={getData('BOOKING_ONE_WAY')} iconColor="blue" />,
   );
 
   expect(multicity.getByType(Icon).props.name).toBe('flight-multicity');
   expect(bookingReturn.getByType(Icon).props.name).toBe('flight-return');
-  expect(oneWay.getByType(Icon).props.name).toBe('flight-direct');
+  expect(OneWay.getByType(Icon).props.name).toBe('flight-direct');
```
tbergq: why should this be uppercased, while multicity and bookingReturn is not?

tbergq:
How about adding these string in `@kiwicom/margarita-utils` so we don't end up in this mess again? 😊

aalises:
Great Idea @tbergq , I will put them in a utils folder named `constants` so we have them there.

jvaclavik:
That was fast! I created issue two hours ago and there is already a pull request :-)

apps/core/components/fromTo/__tests__/FromToIcon.test.js
aalises: Done!

packages/utils/constants/TripTypes.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+
+export type TripTypes = 'Return' | 'OneWay' | 'MultiCity';
```
JoseDuSV: I was just thinking if `@kiwicom/margarita-config` package will not be more suitable for this.
There is quite thin border between those two packages.
But as I see it now, this package is probably more suitable for constants like this.
I'm just going to create new issue for `utils` folder structure revision, as you probably noticed while adding this, it is not really unified.

aalises: Yes! also there is a babel error when trying to import and use `margarita-utils` on the `graphql` package. It is not transpiled (`margarita-config` is as it uses `modules.export`)

tbergq: The issue is not wether it uses modules.exports or not. Nothing should use modules.export, it is a sign that something is wrong, and we should fix it rather than making temporary patches that will give us further problems in the future. 

Luckily we already solved this in `incubator/universe`, so I will prepare a PR to fix this.

tbergq: after #621 is merged, it should work just fine.

aalises: Moved to `margarita-config` to overcome this problem!

PR #617, opened 2019-04-02
Upgrade @kiwicom/relay to latest version

tbergq:


apps/core/scenes/results/Results.js
```diff
@@ -103,6 +79,14 @@ export default class Results extends React.Component<Props> {
       DateFNS.format(DateFNS.parseISO(date), ORDINAL_DAY_MONTH_FORMAT);
 
     const tripType: TripTypes = returnDateFrom ? 'Return' : 'OneWay';
+    const QueryComponent =
+      tripType === 'Return' ? ReturnResultsQuery : OneWayResultQuery;
```
aalises: Thanks for the refactoring! Cleaner this way. One thing, maybe we should rename `ReturnResultsQuery` and `OneWayResultQuery` to be either both `<Type>ResultsQuery` or `<Type>ResultQuery` for consistency.

tbergq: Not sure I understand exactly what you want to change, could you please elaborate?

aalises: It's just a naming convention, Return **Results**Query and OneWay**Result**Query, maybe putting both to use either Result or Results for consistency would be nice 💃

tbergq: Yes, you are right, I didn't catch that. I will name them both with plural

tbergq:
@JoseDuSV @aalises 🚀

PR #618, opened 2019-04-02
Feat: Added code to Transporter type

JoseDuSV:
Summary: So far only name was used for Transporter but in a wrong way in multiple cases, because instead of actual carrier name the code was mapped to this value.
Because both values are necessary for result detail, `code` value was added and mapping with implementation was updated.

MMB itinerary detail before:
Screenshot 2019-04-02 at 13 57 35

After:
Screenshot 2019-04-02 at 13 58 02


schema.graphql
```diff
@@ -508,6 +508,7 @@ enum SortSearchInput {
 
 type Transporter {
```
JoseDuSV: I was also considering to switch type name from `Transporter` to `Carrier`. 
Carrier is used across glossary https://kiwi.wiki/kiwi/handbook/#/comms/glossary and it will also be unified with new kiwi search schema.

What do you think?

packages/universal-components/src/CarrierLogo/CarrierLogoHelpers.js
```diff
@@ -9,18 +9,23 @@ export const getCarrierImageUri = (
   pixelRatio: number,
   carrierData: CarrierData,
 ): string => {
+  const type =
+    carrierData.type === undefined
+      ? CARRIER_TYPE_OPTIONS.AIRLINE
+      : carrierData.type;
```
tbergq: any reason for not doing `const type = carrierData.type ?? CARRIER_TYPE_OPTIONS.AIRLINE`

Is it because you want to set type = null if carrierData.type === null?

schema.graphql
tbergq: I think that is a good idea yes. We should use the names from the glossary

jvaclavik: +1

packages/universal-components/src/CarrierLogo/CarrierLogoHelpers.js
JoseDuSV: Thanks, updated. I just blindly extracted this part and moved it few lines up from url constructor without thinking about it. It should fall to default value also for `null` case, so this version is definitely better and shorter as a bonus.

schema.graphql
JoseDuSV: Ok, thanks, switching this to WIP and will re-request reviews in a moment after renaming.

JoseDuSV: Updated, renamed.

tbergq:
@JoseDuSV is it ready to be merged?

JoseDuSV:
> @JoseDuSV is it ready to be merged?

Yes, just rebased it and it is ready. Thanks

PR #621, opened 2019-04-03
Add @kiwicom/monorepo to make graphql work with monorepo packages

tbergq:


yarn.lock
```diff
@@ -78,7 +78,7 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/core@^7.0.0", "@babel/core@^7.1.0", "@babel/core@^7.1.6", "@babel/core@^7.2.0", "@babel/core@^7.3.4":
+"@babel/core@^7.0.0", "@babel/core@^7.1.0", "@babel/core@^7.1.6", "@babel/core@^7.2.0":
```
JoseDuSV: Thanks, looks good. The only thing is, that after running `yarn install` I'm getting 12 mostly Babel related differences in `yarn.lock`. Looks like the problem we faced before.

Screenshot 2019-04-03 at 10 12 01

tbergq: Can you try again? Should be fixed now

JoseDuSV: Yes, it's ok now. 🚀

tbergq: 👍 I will merge it when CI finishes

JoseDuSV: Definitely, thanks.

PR #623, opened 2019-04-03
Refactor: Changed Structure of Margarita Utils

### LINKED ISSUE #619: Revise @kiwicom/margarita-utils
Revise `@kiwicom/margarita-utils` package content.

- [x] **Unify folders structure** - Currently we have few utils folders directly in package root and other few nested inside `src` and `helpers` subfolders. Maybe everything can be in root. Or we can use some logical subfolders like `formatters`, `constants`, `context`...

- [x] **Unify files naming** - We are mixing capital and small letters in file naming. For example: `device/Device.js` versus `helpers/capitalize.js` 

- [x] **Remove unused utils** - Check if all utils are actually used anywhere and remove unused. For example there are multiple formatters inside `localization` folder which we moved from https://github.com/kiwicom/mobile , but we have started using `date-fns` for date formatting, so we should consider if it still makes sense to keep it there.

aalises:
Closes #619 . Refactor into different folders under `src`... , get rid of unused utils (mostly date formatters). Everything is `PascalCase` now

Screenshot 2019-04-03 at 15 12 40


packages/utils/index.js
```diff
@@ -1,26 +1,22 @@
 // @flow
 
 // Helpers
-export { capitalize } from './helpers/capitalize';
+export { Capitalize } from './helpers/Capitalize';
```
JoseDuSV: Thanks, looks definitely better. I'm just not that sure about those few formatter functions exported with capital letter. 
For me it looks more like class or constant at first glance.
And I'll be more for keeping it with lower case, like in case of `withLayoutContext` (but keep filename with capital).
What do you think?

aalises: Yep, It also looks weird to me. Will change it now.

JoseDuSV: Maybe `Capitalize` can be also moved to `formatters`.

aalises: Sure! that way we get can get rid of the helpers

packages/utils/index.js
```diff
@@ -1,26 +1,22 @@
...
-} from './layoutContext/LayoutContext';
+export { LAYOUT } from './context/LayoutConstants';
+export type { LayoutContextState } from './context/LayoutContext';
+export { withLayoutContext } from './context/LayoutContext';
+export { default as LayoutContextProvider } from './context/LayoutContext';
 
 // src
-export { default as withContext } from './src/withContext/withContext';
-export { default as formatDate } from './src/formatDate';
+export { default as withContext } from './context/withContext';
```
JoseDuSV: We should keep `LayoutContext` and `withContext` in separate subfolders.

Also, I know that I've suggested it as possible structure in description. But now when I see it I'll be more for something slightly more universal. I think it can be moved directly into `helpers`.

Or maybe even better we can consider moving everything into `src` subfolder (into separate subfolders). Because we don't have that many utils (looks like 7 for now) and border between packages is sometimes quite thin (for example `FormatDate` in `formatters`, but `CurrencyFormatter` in `localization` can be confusing) and it can be often hard to choose where new util should be put. I'll let it up to you.

aalises: Moved everything to `src` @JoseDuSV

PR #626, opened 2019-04-03
Add style prop to Button

KarinaDavtyan:


packages/universal-components/src/Button/Button.stories.js
```diff
@@ -62,8 +63,11 @@ storiesOf('Button', module)
     }
     const circled = boolean('circled', false);
     const label = text('label', 'Playground button');
+    const style = object('style', {});
```
JoseDuSV: I think we can remove style object from Storybook, it doesn't seems to work (or I'm just trying values in wrong format 😄). We don't have it there for any other component.

KarinaDavtyan: Yeah this one is tricky, on web it works like `{"borderRadius": 10}` but didnt manage to make it work on mobile

JoseDuSV: Ok, so let just remove it. As I mentioned it's not available for any other styleable component.

KarinaDavtyan: Ok, removed!

JoseDuSV: Thanks

packages/universal-components/src/Button/Button.stories.js
```diff
@@ -8,6 +8,7 @@ import {
   select,
   text,
   number,
+  object,
```
JoseDuSV: @KarinaDavtyan Just remove this one and we can merge it 🚀

KarinaDavtyan: Done

packages/universal-components/src/Button/__tests__/index.test.js
```diff
@@ -24,7 +24,7 @@ const type = 'info';
 const testID = 'testID';
 const circled = false;
 const size = 'normal';
-
+const style = {};
```
RobinCsl: why was this change necessary? it seems like it's polluting the snapshots?

KarinaDavtyan: I assumed that all the props should be in the snapshot, should I remove it in separate PR?

JoseDuSV: It's that repeating question about what we are testing and how. One of the tests is for example `should have passed props` so it probably makes sense there (if we really want to test it).
And same for snapshots. Currently we are using `snapshotDiff` between Button with all props and without. Same as before, question is if it is really useful for something.

RobinCsl: Yes, we would need to sit down and decide what we believe is the best practice.

PR #629, opened 2019-04-04
Upgrade dependencies

tbergq:
Remove unused babel plugins from graphql


apps/graphql/package.json
```diff
@@ -24,10 +24,6 @@
     "webpack": "^4.29.1"
   },
   "devDependencies": {
-    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.2.0",
```
jvaclavik: Will we be able to use coalescing operator etc.?

tbergq: Of course. It is already a part of Kiwi.com-Babel-preset

tbergq:
@JoseDuSV 🚀

PR #631, opened 2019-04-04
Graphql: Make itinerary return type an interface

### LINKED ISSUE #551: Refactor the GraphQL searchItineraries schema to interface
The `searchItineraries` query return the GraphQL connection (  [Itinerary] ) as the output type, but it shows that GraphQL Interface is the better approach, It will provide the better way how to recognize Itinerary type (return, one-way, multicity) For inspiration check the https://kiwi-search-schema.now.sh/voyager

tbergq:
closes #551

### DIFF
--- a/apps/graphql/src/apps/itinerary/types/outputs/ItineraryInterface.js
+++ b/apps/graphql/src/apps/itinerary/types/outputs/ItineraryInterface.js
@@ -1,0 +1,75 @@
+// @flow
+
+import {
+  GraphQLInterfaceType,
+  GraphQLString,
+  GraphQLList,
+  GraphQLBoolean,
+} from 'graphql';
+import GlobalID from '@kiwicom/graphql-global-id';
+import differenceInMinutes from 'date-fns/differenceInMinutes';
+
+import type { Sector } from '../../../common/CommonTypes';
+import type { Itinerary } from '../../Itinerary';
+import GraphQLPrice from '../../../common/types/outputs/Price';
+import GraphQLSector from '../../../common/types/outputs/Sector';
+import GraphQLRouteStop from '../../../common/types/outputs/RouteStop';
+
+const getStopoverDuration = (
+  sector: GraphQLSector,
+  previousSector: ?GraphQLSector,
+): number | null => {
+  if (previousSector == null) {
+    return null;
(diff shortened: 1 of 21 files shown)


apps/graphql/src/apps/itinerary/types/outputs/ItineraryInterface.js
```diff
@@ -0,0 +1,75 @@
...
+  return differenceInMinutes(
+    new Date(currentSectorDeparture),
+    new Date(previousSectorArrival),
+  );
+};
+
+export const commonFields = {
+  id: GlobalID(({ id }) => id),
+  price: { type: GraphQLPrice },
+  sectors: {
```
JoseDuSV: We should then refactor this to use just `sector` for `OneWay` and `inbound/outbound` for `Return`. Like in case of booking and https://kiwi-search-schema.now.sh/voyager . But this is definitely for different PR.

Otherwise looks good. Thanks

tbergq: Yes, my ideas was to do it step by step. This PR is big enough as it is

JoseDuSV: Definitely, so lets merge it 😉

apps/graphql/src/apps/itinerary/types/outputs/ItineraryReturn.js
```diff
@@ -0,0 +1,18 @@
...
+import { GraphQLObjectType } from 'graphql';
+import { TRIP_TYPES } from '@kiwicom/margarita-config';
+
+import type { Itinerary } from '../../Itinerary';
+import ItineraryInterface, { commonFields } from './ItineraryInterface';
+
+export default new GraphQLObjectType({
+  name: 'ItineraryReturn',
+  description:
+    'Itinerary from A to B with no return, but with possible stopovers',
```
aalises: The description isn't changed from the `OneWay` itinerary.

tbergq: Copy paste mistake. I will fix it ASAP

PR #634, opened 2019-04-04
Itineraries: Split sectors into sector on OneWay

tbergq:
and inbound, outbound for Return


apps/graphql/src/apps/itinerary/helpers/stopoverDuration.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+import differenceInMinutes from 'date-fns/differenceInMinutes';
+
+import GraphQLSector from '../../common/types/outputs/Sector';
+
+const getStopoverDuration = (
+  sector: GraphQLSector, // TODO: Check type
+  previousSector: ?GraphQLSector,
+): number | null => {
```
jvaclavik: Is there any difference between `number | null` and `?number`?

tbergq: Yes, number | null means it can be a number or null while ?number means it can be number, null or undefined

tbergq: Just noticed, I wanted to improve that type, so making it WIP again.

tbergq: Yes, that type was bad. Improved.

PR #637, opened 2019-04-05
Add loading state to Button component

KarinaDavtyan:
By passing `isLoading` prop Buttons state changes to loading.
Also added `color` prop in Loader component as it is required in design

closes #635

PR #646, opened 2019-04-08
Fix: Button loader visibility

JoseDuSV:
Summary: Button loader is present in layout even if it's hidden.

Screenshot 2019-04-08 at 12 42 36
Screenshot 2019-04-08 at 12 50 43

Note: Maybe we can later refactor `ButtonInner` into two components to make insides slightly cleaner.
I've been also thinking about solving this issue directly inside `Loader` (simply returning `null` when `animating` is `false`). But at the and I think that handling it inside button is slightly better. There can be cases where hidden `Loader` inside layout can be required to prevent layout jumping while it is switched on/off.

### DIFF
--- a/packages/universal-components/src/Button/ButtonContent.js
+++ b/packages/universal-components/src/Button/ButtonContent.js
@@ -1,0 +1,92 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import { Icon } from '../Icon';
+import { StyleSheet } from '../PlatformStyleSheet';
+import ButtonTitle from './ButtonTitle';
+import type { ButtonType, ButtonSize } from './ButtonTypes';
+import { textColor } from './styles';
+import type { IconSize } from '../Icon/IconTypes';
+
+type Props = {|
+  +type?: ButtonType,
+  +children?: React.Node,
+  +leftIcon?: React.Element<typeof Icon> | null,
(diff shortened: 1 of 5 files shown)


tbergq:
The general idea here was not to need this ternary in the component, but rather rely on the [animating](https://facebook.github.io/react-native/docs/activityindicator#animating) prop of `ActivityIndicator`

JoseDuSV:
> The general idea here was not to need this ternary in the component, but rather rely on the [animating](https://facebook.github.io/react-native/docs/activityindicator#animating) prop of `ActivityIndicator`

I've found this discussion https://github.com/kiwicom/margarita/pull/637#discussion_r272562212 but as I've mentioned in PR note. I'm slightly more for solution directly inside button.
At the end we still need to have `isLoading` condition there for label content so ternary adds just one extra line. But I agree that readability slightly suffers (and we should then at least extract label part into separate component).

But if you want to have it solved inside `Loader` I don't have a big problem with it. Maybe we should then just rename `animating` prop to `isVisible`, because it can be confusing that component is not part of layout when `animating` is set to false.

tbergq:
Sounds ok, Please do extract the label into a separate component and return null in the loader when it is not showing.

JoseDuSV:
@tbergq @KarinaDavtyan 

Ok, updated:
- Extracted button content into separate component to make `ButtonInner` slightly lighter.
- Renamed `Loader` `animating` prop to `isVisible` and component now returns `null` if set to `false`.

PR #649, opened 2019-04-08
Refactor: Passenger Form moved to Modal

JoseDuSV:
Summary: Removed PassengerForm screen and moved form content to modal according to actual designs.

Note: This structure will also save us little bit of trouble with passing passenger info between screens. And it will be easier to refactor web version to something closer to kiwi.com later (with form directly on page).

Apr-08-2019 16-55-16


tbergq:
@JoseDuSV some tests are unhappy

JoseDuSV:
> @JoseDuSV some tests are unhappy

👍 Updated, they are happy now.

PR #655, opened 2019-04-10
Fix RangeDatePicker arrow sync on iOS

### LINKED ISSUE #639: Fix arrow animation in RangeDatePicker
Animation is desynchronised after selection. When arrow is clicked both animations should be restarted.

jvaclavik:
On iOS was still some problem with sync, now it should be fine.
(Fixes #639)

### DIFF
--- a/packages/universal-components/src/RangeDatePicker/AnimatedDayItemArrow.js
+++ b/packages/universal-components/src/RangeDatePicker/AnimatedDayItemArrow.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react';
-import { Animated, Easing, View } from 'react-native';
+import { Animated, Easing, View, InteractionManager } from 'react-native';
 
 import { StyleSheet, type StylePropType } from '../PlatformStyleSheet';
 import { designTokens } from '../DesignTokens';
@@ -25,7 +25,7 @@ const RenderArrows = ({ scale }) => (
       style={[
         styles.easingArrow,
         {
-          transform: [{ scale }],
+          transform: [{ scale: scale ?? 0 }],
         },
       ]}
     />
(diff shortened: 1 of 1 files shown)


packages/universal-components/src/RangeDatePicker/AnimatedDayItemArrow.js
```diff
@@ -34,35 +34,48 @@ const RenderArrows = ({ scale }) => (
...
-      }),
-    ]).start(() => {
-      if (AnimatedDayItemArrow.animating) {
+    if (AnimatedDayItemArrow.scale === undefined) {
+      AnimatedDayItemArrow.scale = new Animated.Value(0);
+    }
+    if (
+      AnimatedDayItemArrow.scale &&
+      AnimatedDayItemArrow.scale._value === 0 &&
+      AnimatedDayItemArrow.numberOfInstances !== 0
```
JoseDuSV: Animation freezes when picker is opened for second time. It looks like because of this condition.
anim-freeze

packages/universal-components/src/RangeDatePicker/AnimatedDayItemArrow.js
```diff
@@ -34,35 +34,48 @@ const RenderArrows = ({ scale }) => (
...
-      }
-    });
+      });
+    }
   }
 
   componentDidMount() {
-    AnimatedDayItemArrow.animating = true;
-    AnimatedDayItemArrow.handleAnimation();
+    if (!AnimatedDayItemArrow.animating) {
```
JoseDuSV: I think that those conditions and `animating`-`numberOfInstances` combinations starting to be quite complex and slightly hard to read.

Cannot we just simply stop/hide animation when user is touching/dragging picker elements and start it again when he stops. I think it's done in similar way in current Kiwi mobile app.

packages/universal-components/src/RangeDatePicker/AnimatedDayItemArrow.js
```diff
@@ -34,35 +34,56 @@ const RenderArrows = ({ scale }) => (
...
+          AnimatedDayItemArrow.handleAnimation();
+        }
+      });
+    }
   }
 
   componentDidMount() {
-    AnimatedDayItemArrow.animating = true;
-    AnimatedDayItemArrow.handleAnimation();
+    if (AnimatedDayItemArrow.numberOfInstances === undefined) {
```
JoseDuSV: Definitely looks better now. Can't say I like all those conditions, but I can probably live with it if it's necessary for synchronisation 😄  .

Just `ios` version still acts weirdly. If I open picker for second time, animation goes to 100% then freezes for about second and then continues as expected.

arrow freeze

jvaclavik: OK, should be fixed. Animation start is now wrapped in `InteractionManager.runAfterInteractions`

JoseDuSV: Looks good now! Thanks

PR #656, opened 2019-04-11
Feat: improved .env generation on CI

RobinCsl:
Summary: Before, it was enough to write `echo GRAPHQL_URL=$GRAPHQL_URL > .env` to generate necessary .env file for `react-native-dotenv` to work correctly. With https://github.com/kiwicom/margarita/pull/602, it will be necessary to add many environment variables and the previous approach will not scale.


tbergq:
Why not use `dotenv` package to load the variables we want, and write them to .env file?

scripts/create-dotenv.js
```diff
@@ -0,0 +1,70 @@
...
+    if (error) {
+      // file doesn't exist yet
+      fs.writeFile(ENV_FILE, output, cb);
+    } else {
+      console.log(`${ENV_FILE} already exists; aborting.`);
+    }
+  });
+};
+
+export default async function run() {
```
tbergq: why do we export this function? 

Also, you could use IIFE here.

```
(async () => {
  const secrets = await _getSecrets();
 ....
})()
```

PR #657, opened 2019-04-11
Feat: adding back netlify.toml and deploy script

RobinCsl:
Summary: This is to better version control the build commands for Netlify, instead of relying on their UI, to which I only have access. This should be merged after https://github.com/kiwicom/margarita/pull/656.
    It adds a deploy script that, based on an environment variable SITE_ID (that has to be set on Netlify's UI, sadly), will deploy the right site. This is a current limitation of `netlify.toml` as it does not support deploying multiple sites from one repository out of the box. (see https://github.com/netlify/netlify-cms/issues/1325 for more information)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,5 @@ coverage/
 # IDE
 .idea
 .vscode
+
+deploy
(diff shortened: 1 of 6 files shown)


RobinCsl:
Actually, since there are several Netlify websites watching this repo, I need to change this in a better way, otherwise the Storybook page will be Margarita page too...

PR #658, opened 2019-04-11
Web: Push search parameters to url

tbergq:
Recreate search parameters when user clicks back


apps/core/scenes/search/Search.js
```diff
@@ -58,39 +70,54 @@ class Search extends React.Component<Props> {
...
+      const returnDateFrom = format(
+        this.props.returnDateFrom,
+        BASIC_ISO_DATE_FORMAT,
+      );
+      const returnDateTo = format(
+        this.props.returnDateTo,
+        BASIC_ISO_DATE_FORMAT,
+      );
+      if (onSubmit != null) {
+        onSubmit({
```
RobinCsl: So this will only be executed on web for now, since there is no onSubmit on mobile, right?

tbergq: Yes, and the context is preserved on mobile, so no need to do it any other way there.

Issue #537, opened 2019-03-18
Implement confirm_payment from Booking API


JoseDuSV:
Tequila Booking API `confirm_payment` request confirms that the user payment for booking was successful.
Booking needs to be saved first by `save_booking` request ( related issue #536 ).

Implementation will depend on how we decide to mock payment flow for now.
Before starting on this issue we should discuss with Tequila team current state of `Payments API`. So we can mock payments flow to be easily replaced/updated once Payments API is ready.

Booking API documentation:
https://tequila.kiwi.com/portal/resources/booking_api
https://tequila.kiwi.com/portal/resources/booking_and_payment_flow

aalises:
Schema proposal. In this case it's pretty simple.

**Mutation name (POST):**
```
confirmPayment
```

**Payload**
```
type ConfirmPaymentPayload {
  bookingID: String
  transactionID: String!
}
``` 

**Output**
```
type ConfirmPaymentResponse {
  status: Int!
  price: Float!
}
```
@JoseDuSV

PR #659, opened 2019-04-12
Feature: confirm_payment GraphQL Mutation for Booking Workflow

aalises:
Implement the `confirm_payment` mutation of Tequila Booking API.(https://tequila.kiwi.com/portal/resources/booking_api).

Relates to #537

### DIFF
--- a/apps/graphql/src/apps/booking/types/outputs/ConfirmPayment.js
+++ b/apps/graphql/src/apps/booking/types/outputs/ConfirmPayment.js
@@ -1,0 +1,19 @@
+// @flow
+
+import { GraphQLObjectType, GraphQLBoolean, GraphQLFloat } from 'graphql';
+
+const ConfirmPaymentOutput = new GraphQLObjectType({
+  name: 'ConfirmPaymentOutput',
+  fields: {
+    isSuccess: {
+      description: 'Indicates whether the payment was successful',
+      type: GraphQLBoolean,
+    },
+    price: {
+      description: 'The average price of the booking per passenger',
+      type: GraphQLFloat,
+    },
+  },
+});
+
+export default ConfirmPaymentOutput;
--- a/apps/graphql/src/Schema.js
+++ b/apps/graphql/src/Schema.js
@@ -7,7 +7,8 @@ import SearchReturnItineraries from './apps/itinerary/queries/SearchReturnItiner
 import LocationsByTerm from './apps/location/queries/LocationsByTerm';
 import CustomerBookings from './apps/booking/queries/CustomerBookings';
 import BookingDetail from './apps/booking/queries/BookingDetail';
-import SaveBooking from './apps/booking/queries/SaveBooking';
+import SaveBooking from './apps/booking/mutations/SaveBooking';
+import ConfirmPayment from './apps/booking/mutations/ConfirmPayment';
 import CheckItinerary from './apps/itinerary/queries/CheckItinerary';
 import BookingOneWay from './apps/booking/types/outputs/BookingOneWay';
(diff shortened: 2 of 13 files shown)


apps/graphql/src/apps/booking/queries/ConfirmPayment.js
```diff
@@ -0,0 +1,29 @@
+// @flow
```
tbergq: maybe we should have a folder `/bookings/mutations` where this file should be. It is not a query, but a mutation right?

apps/graphql/src/apps/booking/queries/SaveBooking.js
```diff
@@ -12,7 +12,7 @@ type Args = {|
 const SaveBooking = {
   name: 'SaveBooking',
   description:
-    'Mutation (POST Request) that saves a given booking. Initiates the reservation process on the backend.',
+    'Saves a given booking. Initiates the reservation process on the backend. (POST Request)',
```
tbergq: I guess this also goes for this file, you don't have to do it within this PR, but it should be moved to mutations.

apps/graphql/src/apps/booking/queries/__tests__/ConfirmPaymentMutation.test.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import { generateTestsFromFixtures } from '@kiwicom/test-utils';
```
tbergq: The idea is to have one test file read all fixtures. It should not be necessary to create this file again within the same folder there is already a `queries.test.js` here. Now, I guess this file will also be moved to mutations folder, so it is ok.

apps/graphql/src/apps/booking/queries/__tests__/queries.test.js
```diff
@@ -7,6 +7,6 @@ import Itineraries from '../../../itinerary/__datasets__/Itineraries.json';
 
 fetch.mockResponses([JSON.stringify(Itineraries)]);
 
-generateTestsFromFixtures(`${__dirname}/__fixtures__`, input =>
+generateTestsFromFixtures(`${__dirname}/__fixtures__/queries`, input =>
```
tbergq: Please don't do this, we don't have to copy paste this file every time we add a new test. The idea is that it reads all fixtures and creates snapshots.

apps/graphql/src/apps/booking/types/outputs/ConfirmPayment.js
```diff
@@ -0,0 +1,20 @@
...
+
+import { GraphQLObjectType, GraphQLInt, GraphQLFloat } from 'graphql';
+
+const ConfirmPaymentOutput = new GraphQLObjectType({
+  name: 'ConfirmPaymentOutput',
+  fields: {
+    status: {
+      description:
+        'Should be 0 (zero) that means the confirm_payment was successful, any other result means that something went wrong',
+      type: GraphQLInt,
```
tbergq: Is there a meaning behind the other status, or could we just as well return isSuccess: Boolean here?

aalises: I don't know, according to the documentation... 

Screenshot 2019-04-12 at 14 58 01

tbf it would be better to just have a boolean :/

tbergq: So, let's return a boolean instead of this cryptic number 😊

apps/graphql/src/apps/booking/queries/__tests__/queries.test.js
aalises: Done!, the separation is done in `mutations`, `queries`.

PR #662, opened 2019-04-12
LandingPage: move Head to _document.js

KarinaDavtyan:
Summary: title was removed due to https://github.com/zeit/next.js/blob/master/errors/no-document-title.md, and global styles are added now as dangerouslySetInnerHTML as in /web

### DIFF
--- a/apps/landingPage/pages/_document.js
+++ b/apps/landingPage/pages/_document.js
@@ -1,8 +1,15 @@
 // @flow
 
 import React from 'react';
-import Document from 'next/document';
-import { ServerStyleSheet } from 'styled-components';
+import Document, { Head, Main, NextScript } from 'next/document';
+import { ServerStyleSheet, createGlobalStyle } from 'styled-components';
+
+const GlobalStyle = createGlobalStyle`
+  body {
+    margin: 0;
+    -webkit-font-smoothing: antialiased;
+  }
+`;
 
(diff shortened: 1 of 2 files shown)


aalises:
You can still add the title if you move the head to the `_app.js` file (maybe creating a separate `Header` component for the head and styles would be nice). I don't know how important it is right now to have a title for the good ole' SEO...  😆

KarinaDavtyan:
> You can still add the title if you move the head to the `_app.js` file (maybe creating a separate `Header` component for the head and styles would be nice). I don't know how important it is right now to have a title for the good ole' SEO... 😆

Yeah I dont think that adding title is important right now (would be nice to remove placeholders first 😅)

apps/landingPage/pages/_document.js
```diff
@@ -1,9 +1,18 @@
 // @flow
 
 import React from 'react';
-import Document from 'next/document';
+import Document, { Head, Main, NextScript } from 'next/document';
 import { ServerStyleSheet } from 'styled-components';
 
+const globalStyle = {
```
tbergq: any reason we cannot use [createGlobalStyle](https://www.styled-components.com/docs/api#createglobalstyle)?

KarinaDavtyan: I guess not, seem to work fine with it.

PR #669, opened 2019-04-15
Create new device and localization packages

tbergq:
move some functions out of utils into new packages.
Utils should be simple functions we can reuse between web, mobile and graphql.
We could not use utils in grahpql now, since it depended on react-native
which does not run in nodejs.


tbergq:
I don't like the `shared-utils` vs `FE-utils` at least. Utils should be something that doesn't depend on platform in my ears. 

Device definitely depends on platform. You might be right about `GetDeviceLocale`, it might belong to `device` 🤔

PR #671, opened 2019-04-15
MMB: Capitalize passenger titles

### LINKED ISSUE #667: App crashes when going to manage my booking detail screen
IMG_63B7995C025E-1

tbergq:
From current mmb, we get titles as lowercase.
Passenger title enum expects title to be capitalized.

closes #667

### DIFF
--- a/apps/graphql/src/apps/manageMyBooking/dataloaders/Bookings.js
+++ b/apps/graphql/src/apps/manageMyBooking/dataloaders/Bookings.js
@@ -2,6 +2,7 @@
 
 import { head, last } from 'ramda';
 import differenceInMinutes from 'date-fns/differenceInMinutes';
+import { capitalize } from '@kiwicom/margarita-utils';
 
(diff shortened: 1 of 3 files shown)


apps/graphql/src/apps/manageMyBooking/dataloaders/Bookings.js
```diff
@@ -51,6 +53,8 @@ export const sanitizeBooking = (booking: BookingApiResult) => {
 
 const sanitizePassengers = (booking: BookingApiResult): Array<Passenger> =>
   booking.passengers.map(passenger => {
+    const passengerTitle = capitalize(passenger.title);
+    const title: PassengerTitle = (passengerTitle: any);
```
JoseDuSV: It will be nice to get rid of this part, but for now I don't really have anything better 😄. And since we are just working with mocked data for now it doesn't make much sense to dig deeper into response types when it will be updated once/if we connect MMB to real endpoint.

PR #675, opened 2019-04-16
Feat: CheckItinerary params

JoseDuSV:
Summary: Replaced mock values with real params for CheckItinerary query.
Which means that result is matching value from results list.

### DIFF
--- a/apps/core/scenes/resultDetail/ResultDetail.js
+++ b/apps/core/scenes/resultDetail/ResultDetail.js
@@ -8,6 +8,8 @@ import ResultDetailInner from './ResultDetailInner';
 
 type Props = {|
   +bookingToken: ?string,
+  +adults: ?number,
+  +infants: ?number,
 |};
 
 export default class ResultDetail extends React.Component<Props> {
@@ -16,6 +18,7 @@ export default class ResultDetail extends React.Component<Props> {
   };
 
   render() {
+    const { bookingToken, adults, infants } = this.props;
     return (
       <QueryRenderer
         query={graphql`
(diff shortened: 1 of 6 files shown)


apps/core/scenes/resultDetail/ResultDetail.js
```diff
@@ -25,9 +28,12 @@ export default class ResultDetail extends React.Component<Props> {
         `}
         variables={{
           input: {
-            bookingToken: this.props.bookingToken,
+            bookingToken,
             bags: 0, // @TODO - use real data from search context
-            passengers: { adults: 1 }, // @TODO - use real data
+            passengers: {
+              adults: adults ? parseInt(adults, 10) : 1,
```
JoseDuSV: I'm not that happy about this part. Mainly because we are mixing values from url and search context in few places. I want to do some refactoring in different PR and use SearchContext as main source and parser of parameters for Search/Results/ResultDetail . I just need to sync with @aalises  because he will probably need something similar for search input above web results.

tbergq: I guess this is string | number because it is passed differently from web and mobile?

How about doing the parseInt on the parent, so the type here can be just number?

JoseDuSV: @tbergq Good idea. Moved parsing to parent for web version (where params are string) which makes it a bit cleaner. Thanks

apps/web/pages/result-detail.js
```diff
@@ -9,10 +9,16 @@ import { withPageRouter } from '../components/withPageRouter';
 import Layout from '../components/Layout';
 
 const resultDetail = ({ router }) => {
+  const { bookingToken, adults, infants } = router.query;
+  const params = {
+    bookingToken,
+    adults: adults ? parseInt(adults, 10) : 1,
+    infants: infants ? parseInt(infants, 10) : 0,
+  };
```
tbergq: What about

```
const params = {
    bookingToken,
    adults: parseInt(adults ?? 1, 10),
    infants: parseInt(infants ?? 0, 10),
  };
```

JoseDuSV: 👍 Definitely better. Updated

PR #677, opened 2019-04-16
Feature: Refactor Search Form into separate component, add it to results page

aalises:
Screenshot 2019-04-16 at 12 05 39


apps/core/scenes/results/Results.js
```diff
@@ -130,19 +131,26 @@ class Results extends React.Component<Props> {
...
-            localizedDate: getFormattedDate([dateFrom, dateTo]),
-          }}
-          arrival={{
-            city: travelToName,
-            localizedDate: returnDateFrom
-              ? getFormattedDate([returnDateFrom, returnDateTo])
-              : '',
-          }}
-        />
+        {Platform.OS === 'mobile ' ? (
```
JoseDuSV: Extra space in `mobile` is causing this.
Screenshot 2019-04-16 at 11 40 18

apps/core/components/searchForm/SearchForm.js
```diff
@@ -0,0 +1,224 @@
+// @flow
+
+// @flow
```
JoseDuSV: Extra flow declaration, should be removed.

aalises: Yes! Just did it

jvaclavik:
Nice, it looks better than I expected 👍

apps/core/scenes/results/Results.js
```diff
@@ -130,19 +131,26 @@ class Results extends React.Component<Props> {
...
-            localizedDate: getFormattedDate([dateFrom, dateTo]),
-          }}
-          arrival={{
-            city: travelToName,
-            localizedDate: returnDateFrom
-              ? getFormattedDate([returnDateFrom, returnDateTo])
-              : '',
-          }}
-        />
+        {Platform.OS === 'mobile' ? (
```
JoseDuSV: It is still there. I just realised that there is no `mobile` platform. We only have `web`, `ios` and `android`.
So best solution will be to reverse it and use `Platform.OS === 'web'`.

aalises: @JoseDuSV Done!

apps/core/components/searchForm/SearchForm.js
```diff
@@ -0,0 +1,231 @@
...
+    const {
+      travelFrom,
+      travelTo,
+      tripType,
+      adults,
+      infants,
+      bags,
+      onSubmit,
+    } = this.props;
+    if (travelFrom == null) {
```
JoseDuSV: We broke this condition with some other update. But I just spotted it. 
Can you please extend it to `travelFrom == null || travelFrom.length === 0`. The `travelFrom` is now array (and can be empty) it was just single object before.

apps/core/scenes/search/Search.js
```diff
@@ -150,27 +27,8 @@ class Search extends React.Component<Props> {
...
-                onPress={this.handleSubmitPress}
-                label="Search"
-                rightIcon={
-                  Platform.OS === 'web' ? <Icon name="chevron-right" /> : null
-                }
-              />
-            </View>
-          </View>
-          <View style={styles.bottomSpacer} />
+          <SearchForm />
```
JoseDuSV: `onSubmit` web call stopped working. We will probably need to update and revise this part a bit.
But can you for now do just `<SearchForm onSubmit={this.props.onSubmit} />` to pass function down to make it work again (it is for now defined inside `web/pages/index.js` ).

apps/core/components/searchForm/SearchForm.js
tbergq: No lint that catches this 😱

apps/core/scenes/search/Search.js
```diff
@@ -150,27 +27,8 @@ class Search extends React.Component<Props> {
...
-              <Button
-                onPress={this.handleSubmitPress}
-                label="Search"
-                rightIcon={
-                  Platform.OS === 'web' ? <Icon name="chevron-right" /> : null
-                }
-              />
-            </View>
-          </View>
-          <View style={styles.bottomSpacer} />
```
JoseDuSV: I think we should keep this spacer as part of the Search screen. It simply adds bottom padding on web to prevent sticking inputs to bottom edge on small screens and centers content on mobile.
It is related only to this screen and doesn't need to be present anywhere else. On results screen we have it disabled for now, but quite unintentionally by current set of rules.

apps/core/scenes/results/Results.js
```diff
@@ -181,6 +192,11 @@ const styles = StyleSheet.create({
     },
     justifyContent: 'flex-start',
   },
+  searchFormContainer: {
+    paddingHorizontal: 125,
+    paddingVertical: 20,
+    minWidth: 500,
```
jvaclavik: Isn't it pity to set minWidth when design is responsive?

apps/core/scenes/search/Search.js
```diff
@@ -150,27 +43,9 @@ class Search extends React.Component<Props> {
...
+          <SearchForm onSubmit={this.props.onSubmit} />
+          {Platform.OS !== 'web' && <View style={styles.bottomSpacer} />}
```
JoseDuSV: Last thing, otherwise looks good. `bottomSpacer` actually have some web related style. It's not that important and simply just as I've mentioned prevents inputs to stick to the bottom edge on really small screens.
The option is to remove this condition or that (in this case unused) style starting on row 96.

apps/core/scenes/results/Results.js
aalises: Yes! Looks pretty ugly to me, but if not that happens: 

Screenshot 2019-04-16 at 16 15 17

Any idea on how to do that better? @jvaclavik

JoseDuSV: True, we can also save some space by defining some smaller `paddingHorizontal` value for smaller layouts.

Screenshot 2019-04-16 at 16 17 08

apps/core/scenes/search/Search.js
```diff
@@ -234,51 +93,11 @@ const styles = StyleSheet.create({
   },
   bottomSpacer: {
     flex: 1,
-    web: {
```
JoseDuSV: Sorry about this 😄. But can we at the end return this style and remove that new `{Platform.OS !== 'web'` condition for `bottomSpacer`.

After I tested some smaller resolutions I think that it looks better when there is at least little bit of padding at the bottom.
I this version (in extreme cases) when I scroll down it looks like this and I cannot scroll further.
Screenshot 2019-04-16 at 16 40 56

It's better to let user scroll little bit more and add some breathing space to button at the bottom like in current master.

Screenshot 2019-04-16 at 16 41 36

PR #679, opened 2019-04-16
Add carrier name in trip detail

KarinaDavtyan:


apps/graphql/src/apps/itinerary/helpers/Itineraries.js
```diff
@@ -37,10 +38,19 @@ export const mapVehicle = (type: ?string, uniqueNo: ?string) => ({
...
-export const sanitizeCarrier = (routeItem: ?ApiRouteItem) => ({
-  name: null, // @TODO - value is currently missing from API endpoint response
-  code: routeItem?.airline ?? null,
-});
+export const sanitizeCarrier = (routeItem: ?ApiRouteItem) => {
+  // @TODO - value is currently missing from API endpoint response
+  // JSON file is used from https://github.com/BesrourMS/Airlines/blob/master/airlines.json
+  const iata = routeItem?.airline;
+  let name = '';
+  for (let index = 0; index < airlines.length; index++) {
```
aalises: You can probably do something here like (accounting for null values ofc):

```js
name = airlines.filter(airline => airline['iata'] == iata)[0].name
```

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
```diff
@@ -37,10 +38,19 @@ export const mapVehicle = (type: ?string, uniqueNo: ?string) => ({
...
 });
 
-export const sanitizeCarrier = (routeItem: ?ApiRouteItem) => ({
-  name: null, // @TODO - value is currently missing from API endpoint response
-  code: routeItem?.airline ?? null,
-});
+export const sanitizeCarrier = (routeItem: ?ApiRouteItem) => {
+  // @TODO - value is currently missing from API endpoint response
+  // JSON file is used from https://github.com/BesrourMS/Airlines/blob/master/airlines.json
+  const iata = routeItem?.airline;
```
aalises: iata seems a little bit obscure especially for me that didn't know what it was! we could just name it `airlineCode` or something like that and then 
```
return {
    name: name ?? null,
    code: airlineCode ?? null,
  };
```

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
tbergq: I think even `airlines.find(...)` since find will terminate on the first occurrence, and there should be only one. filter will loop through the entire array.

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
KarinaDavtyan: Alright, changed to airlineCode

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
KarinaDavtyan: Updated to `airlines.find(...)`

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
```diff
@@ -37,10 +38,17 @@ export const mapVehicle = (type: ?string, uniqueNo: ?string) => ({
...
-export const sanitizeCarrier = (routeItem: ?ApiRouteItem) => ({
-  name: null, // @TODO - value is currently missing from API endpoint response
-  code: routeItem?.airline ?? null,
-});
+export const sanitizeCarrier = (routeItem: ?ApiRouteItem) => {
+  // @TODO - value is currently missing from API endpoint response
+  // JSON file is used from https://github.com/BesrourMS/Airlines/blob/master/airlines.json
+  const airlineCode = routeItem?.airline;
+  const airline = airlines.find(airline => airline.iata === airlineCode);
+  const name = airline ? airline.name : null;
```
tbergq: const name = airline?.name

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
```diff
@@ -37,10 +38,17 @@ export const mapVehicle = (type: ?string, uniqueNo: ?string) => ({
...
+  return {
+    name: name,
```
tbergq: name,

KarinaDavtyan: 😓

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
aalises: That's the last thing and I think it's good to go :) 🚀

KarinaDavtyan: thanks

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
```diff
@@ -37,10 +38,17 @@ export const mapVehicle = (type: ?string, uniqueNo: ?string) => ({
...
-export const sanitizeCarrier = (routeItem: ?ApiRouteItem) => ({
-  name: null, // @TODO - value is currently missing from API endpoint response
-  code: routeItem?.airline ?? null,
-});
+export const sanitizeCarrier = (routeItem: ?ApiRouteItem) => {
+  // @TODO - value is currently missing from API endpoint response
+  // JSON file is used from https://github.com/BesrourMS/Airlines/blob/master/airlines.json
+  const airlineCode = routeItem?.airline;
+  const airline = airlines.find(airline => airline.iata === airlineCode);
+  const name = airline?.name;
```
aalises: The tests are failing because when there is no `name` the expression returns undefined (and the test expects null). You can either change the test or `airline?.name ?? null`

KarinaDavtyan: Sorry :(
I will update test as airline name also got updated

KarinaDavtyan: Or would be better if I just update airline name part?

tbergq: You don't need to do `airline?.name ?? null` since optional chaining already returns null or undefined. `airline?.name` should be enough

aalises: @tbergq In that case the test should be updated because if there is an airline with an undefined name it returns undefined and the test was failing because it expected null

tbergq: then we should update the test of course

KarinaDavtyan: And test updated :)

PR #684, opened 2019-04-17
Feat: Added sector details to ResultDetail

JoseDuSV:
Summary:
- Extracted `SectorDetailsWrapper` from MMB sector details to make it reusable for results.
- Added expandable sector details to result. According to designs.
https://www.figma.com/file/OUhBJ9rPaTbRljptC0lrRrUF/Multistep

sector-details

Top part of the header will follow in different PR.

Screenshot 2019-04-16 at 18 40 43

### DIFF
--- a/apps/core/components/sectorDetails/ItinerarySectorDetails.js
+++ b/apps/core/components/sectorDetails/ItinerarySectorDetails.js
@@ -1,0 +1,48 @@
+// @flow
+
+import * as React from 'react';
+import { Card, StyleSheet } from '@kiwicom/universal-components';
+import { createFragmentContainer, graphql } from '@kiwicom/margarita-relay';
+import {
+  ItineraryTypeRenderer,
+  SectorDetailsWrapper,
+} from '@kiwicom/margarita-components';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import ItineraryOneWay from '../ItineraryCard/itineraryDetail/ItineraryOneWay';
+import ItineraryReturn from '../ItineraryCard/itineraryDetail/ItineraryReturn';
+import type { ItinerarySectorDetails_itinerary as ItineraryType } from './__generated__/ItinerarySectorDetails_itinerary.graphql';
+
+type Props = {|
+  +itinerary: ?ItineraryType,
+|};
+
+function ItinerarySectorDetails(props: Props) {
+  return (
+    <Card style={styles.container}>
+      <SectorDetailsWrapper>
+        <ItineraryTypeRenderer
+          typename={props.itinerary?.__typename}
+          oneWayComponent={<ItineraryOneWay itinerary={props.itinerary} />}
+          returnComponent={<ItineraryReturn itinerary={props.itinerary} />}
+        />
+      </SectorDetailsWrapper>
+    </Card>
+  );
+}
+
+const styles = StyleSheet.create({
+  container: {
+    marginBottom: parseFloat(defaultTokens.spaceSmall),
+  },
+});
+
+export default createFragmentContainer(ItinerarySectorDetails, {
+  itinerary: graphql`
+    fragment ItinerarySectorDetails_itinerary on ItineraryInterface {
+      __typename
+      ...ItineraryOneWay_itinerary
+      ...ItineraryReturn_itinerary
+    }
+  `,
+});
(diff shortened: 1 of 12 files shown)


apps/core/components/sectorDetails/ItinerarySectorDetails.js
```diff
@@ -0,0 +1,52 @@
...
+  SectorDetailsWrapper,
+} from '@kiwicom/margarita-components';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import ItineraryOneWay from '../ItineraryCard/itineraryDetail/ItineraryOneWay';
+import ItineraryReturn from '../ItineraryCard/itineraryDetail/ItineraryReturn';
+// @TODO - gemerated type seems to be broken - import type { ItinerarySectorDetails_data as ItineraryType } from './__generated__/ItinerarySectorDetails_data.graphql';
+
+type Props = {|
+  +data: any, // @TODO - cannot use genereated ?ItineraryType which seems to be broken,
```
JoseDuSV: Not happy about this part but I cannot make it work. If I try to put content directly into `ResultDetailContent` everything works fine, but once I extract it to fragment, flow start to complaining that `$fragmentRefs is missing in object type [1] but exists in $FragmentRef`.

I'll definitely appreciate any tip and maybe some fresh view will help.

tbergq: Change the fragment to 
```
export default createFragmentContainer(ItinerarySectorDetails, {
  data: graphql`
    fragment ItinerarySectorDetails_data on ItineraryInterface {
      __typename
      ...ItineraryOneWay_itinerary
      ...ItineraryReturn_itinerary
    }
  `,
});
```
and it works

apps/core/components/sectorDetails/ItinerarySectorDetails.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const styles = StyleSheet.create({
+  container: {
+    marginBottom: parseFloat(defaultTokens.spaceSmall),
+  },
+});
+
+export default createFragmentContainer(ItinerarySectorDetails, {
+  data: graphql`
```
tbergq: the reason relay deprecated using graphql directly was to force us programmers to use better naming than calling everything `data`. Lets give it a name like itineraries instead 😊

apps/core/components/sectorDetails/ItinerarySectorDetails.js
JoseDuSV: Thanks, that helped.

apps/core/components/sectorDetails/ItinerarySectorDetails.js
JoseDuSV: True 😊. Updated even for parent container.

packages/components/src/sectorDetailsWrapper/SectorDetailsWrapper.js
```diff
@@ -0,0 +1,86 @@
...
+
+export default class SectorDetailsWrapper extends React.Component<
+  Props,
+  State,
+> {
+  constructor() {
+    super();
+
+    this.state = {
+      expandSegments: false,
```
jvaclavik: I'm not sure about name of this key. What about `areSegmentExpanded` or something like that?

packages/components/src/sectorDetailsWrapper/SectorDetailsWrapper.js
```diff
@@ -0,0 +1,86 @@
...
+        >
+          <Animated.View
+            style={[
+              styles.iconWrapper,
+              {
+                transform: [{ rotate: spin }, { perspective: 1000 }],
+              },
+            ]}
+          >
+            <Icon
```
jvaclavik: Don't we have the same animated icon in BookingDetail?

JoseDuSV: This is the one. Maybe the summary is not that clear. But I extracted this whole wrapper from there and made it reusable.

packages/components/src/sectorDetailsWrapper/SectorDetailsWrapper.js
JoseDuSV: Right. I think we are using `isSomething` across project so I'll change it to `isExpanded`.

JoseDuSV: @jvaclavik  Updated

packages/components/src/sectorDetailsWrapper/SectorDetailsWrapper.js
jvaclavik: I see, in that case it's perfect :)

PR #685, opened 2019-04-17
Eslint: Fix warnings

tbergq:


apps/web/next.config.js
```diff
@@ -47,7 +47,7 @@ module.exports = withImages(
       ];
 
       config.module.rules.push({
-        test: /\.(woff|woff2|eot|ttf|otf)$/,
+        test: /\.woff|woff2|eot|ttf|otf$/,
```
RobinCsl: I think that it's a bit more readable with parens than without. But I don't have a strong objection.

apps/graphql/src/services/helpers/parseAcceptLanguage.js
```diff
@@ -4,12 +4,13 @@ export default function parseAcceptLanguage(acceptLanguage: ?string) {
...
 
   // valid acceptLanguage: en || en_us || en_US || en-US || en-US,en;q=0.9,sl;q=0.8
-  const localeRegex = /^([a-z]{2})(?:[-_]([A-Za-z]{2}))?(,|;|$)/;
+  const localeRegex = /^(?<language>[a-z]{2})(?:[-_](?<territory>[a-z]{2}))?(?:,|;|$)/i;
 
   if (acceptLanguage) {
     const match = localeRegex.exec(acceptLanguage);
     if (match !== null) {
-      [, language, territory = 'US'] = match;
+      language = match.groups?.language ?? 'en';
```
RobinCsl: This is far more readable, however! 🎉

apps/web/next.config.js
mrtnzlml: @tbergq Should be `/\.(?:woff|woff2|eot|ttf|otf)$/` otherwise the first dot is only for `woff` which is different from the previous RegExp behavior. Not a big deal in a real life though.

tbergq: thank you @mrtnzlml I will fix it

tbergq: @mrtnzlml 🚀

packages/components/storybook/webpack.config.js
```diff
@@ -13,12 +13,12 @@ module.exports = {
         },
       },
       {
-        test: /\.(png|jpe?g|gif)$/,
+        test: /\.png|jpe?g|gif$/,
```
RobinCsl: @tbergq it's the same issue here and in other RegEx's as @mrtnzlml called out.

tbergq: 🚀

Issue #682, opened 2019-04-17
Enhance SearchContext and Navigation functionality


JoseDuSV:
Url in **web** version should reflect changes in search parameters on results related pages like `results` and `result-detail`. Basically url should just reflect `SearchContext` props. 

The idea is that we should use just single _“source of truth”_ for screens related to search params (`index`, `results`, `result-detail`).
Currently we are mixing data from `SearchContext` and params from url/navigation. Which is not always clear and causes problems when we for example need to use types like `string | number` for some parameters.

One solution can be to use `SearchContext` for storing and providing all props. And on web just update and parse parameters from url (to `SearchContext`) on redirect or refresh.

We should also consider moving `SearchContext` to `_app.js` for **web** version to make structure equal to mobile, instead of defining it separately inside each page.

aalises:
Closed on #686

PR #686, opened 2019-04-17
Enhancement: Use Context instead of URL for Results.

aalises:
Use Context instead of URL for Results. Relates to #682


tbergq:
If I refresh the browser on the search screen, my search parameters are lost (defaults are restored)

aalises:
> If I refresh the browser on the search screen, my search parameters are lost (defaults are restored)

@tbergq That also happens on master (https://kiwicom-margarita.netlify.com). Don't know which is the intended behavior for that ...

tbergq:
Hmm, Ok, I would definitely expect it to work with refresh also, but we can do that in another PR then.

PR #690, opened 2019-04-18
Feat: Added sector details header to ResultDetail

JoseDuSV:
Summary:
- Moved `SectorInfo` from MMB to components to make it reusable.
- Added interfaces to be able to share data between `BookingIterface` and `ItineraryInterface`.
- Added `SectorInfo` to `ResultDetail`.

Screenshot 2019-04-18 at 10 22 16


apps/graphql/src/apps/common/types/outputs/ReturnInterface.js
```diff
@@ -0,0 +1,25 @@
...
+  name: 'ReturnInterface',
+  description: 'Used for return type Booking and Itinerary',
+  fields: {
+    departure: {
+      type: GraphQLRouteStop,
+    },
+    arrival: {
+      type: GraphQLRouteStop,
+    },
+    outbound: {
```
aalises: Maybe renaming `outbound` and `inbound` to `outboundSector` or `inboundSector` would be cool, it might be confusing as we can also use inbound and outbound for dates...

JoseDuSV: Not sure about this. Personally I like shorter names. I think it's quite clear from schema that it is sector. But if you really find it confusing (maybe I just don't see it after working with this naming for a while 🙂) I can change it.

aalises: Not really! It's okay for me, it's just that I have been tweaking with outbound and inbound dates and took me some seconds to realize those were sectors.

PR #691, opened 2019-04-18
Refactor: Split location dataLoaders (ID and Term)

### LINKED ISSUE #688: Split locations dataloader into two dataloaders
We should have one dataloader for loading by id, and another by term.

aalises:
Now the location data loaders are separated based on the input (IDs or terms) as they call different API endpoints. In the future, we can add more Location dataLoaders based on other parameters (such as slugs) in a similar way.

 Closes #688


apps/core/components/searchForm/SearchForm.js
```diff
@@ -103,8 +104,8 @@ class SearchForm extends React.Component<Props> {
         });
       }
       this.props.navigation.navigate(Routes.RESULTS, {
```
tbergq: Why is this change related to splitting data loaders?

apps/graphql/src/apps/location/dataloaders/LocationsByID.js
```diff
@@ -0,0 +1,39 @@
...
+        .join('&'),
+  );
+  return results.locations.map(location => sanitizeLocations([location]));
+};
+
+export default function locationsByIDLoader() {
+  return new OptimisticDataloader(
+    (
+      ids: $ReadOnlyArray<LocationInputID>,
+    ): Promise<Array<Location[] | Error>> => fetchLocations(ids),
```
tbergq: Would it be possible that this returns `Promise<Array<Location | Error>>` instead. When querying for ID, you just expect to receive one result.

apps/graphql/src/apps/location/dataloaders/LocationsByID.js
```diff
@@ -0,0 +1,39 @@
...
+export type LocationInputID = {|
+  +code: string,
+|};
+
+const fetchLocations = async (params: $ReadOnlyArray<LocationInputID>) => {
+  const results = await fetch(
+    `/locations/id?limit=${params.length}&` +
+      params
+        .map((param: LocationInputID) => {
+          if (param.code !== undefined) {
```
tbergq: Not sure we need this param.code !== undefined anymore

aalises: We don't, on `RouteStop` the `code` can't be undefined

apps/graphql/src/apps/common/types/outputs/RouteStop.js
```diff
@@ -28,7 +28,7 @@ export default new GraphQLObjectType({
         if (code == null) {
           return null;
         }
-        const stops = await dataLoader.locations.load({ code });
+        const stops = await dataLoader.locations.byID.load({ code });
         if (!Array.isArray(stops) || stops.length < 1) {
           return null;
         }
```
tbergq: I see that the return type of the dataloader is changed, which is good, I see that this doesn't to give any flow-error which is bad. I want to get rid of this as well. 

```
 if (!Array.isArray(stops) || stops.length < 1) {
          return null;
  }
```

PR #693, opened 2019-04-18
Fix: Set Navbar background color

JoseDuSV:
To prevent underlying elements to be visible.

Before
Screenshot 2019-04-18 at 17 12 15

After
Screenshot 2019-04-18 at 17 11 55


aalises:
~Looks good to me~ Looks white to me 👍

PR #695, opened 2019-04-23
Keep parameters on reloading (from URL)

aalises:
The parameters are now kept on reloading (obtained from the URL and parsed into the context), on search and results. They are only updated when pressing search though...

### DIFF
--- a/apps/core/scenes/results/Results.js
+++ b/apps/core/scenes/results/Results.js
@@ -36,6 +36,7 @@ import {
   type Location,
 } from '../search/SearchContext';
 import SortTabsWrapper from '../search/SortTabsWrapper';
+import { type SearchParameters } from '../search/Search';
 
 type Props = {|
   +navigation: Navigation,
@@ -52,9 +53,16 @@ type Props = {|
   +bags: number,
   +sortBy: string,
   +layout: number,
+  +routerQuery: SearchParameters,
+  +setStateFromQueryParams: SearchParameters => void,
 |};
 
 class Results extends React.Component<Props> {
+  componentDidMount() {
+    const { setStateFromQueryParams, routerQuery } = this.props;
+    setStateFromQueryParams(routerQuery);
+  }
+
   handleBookPress = (bookingToken: ?string) => {
     const { adults, infants } = this.props;
     this.props.navigation.navigate(Routes.RESULT_DETAIL, {
@@ -229,6 +237,7 @@ const select = ({
   infants,
   bags,
   sortBy,
+  actions: { setStateFromQueryParams },
 }: SearchContextState) => ({
   travelFrom,
   travelTo,
@@ -241,6 +250,7 @@ const select = ({
   infants,
   bags,
   sortBy,
+  setStateFromQueryParams,
 });
 
 const layoutSelect = ({ layout }: LayoutContextState) => ({
--- a/apps/core/scenes/search/SearchContext.js
+++ b/apps/core/scenes/search/SearchContext.js
@@ -4,6 +4,7 @@ import * as React from 'react';
 import { withContext, noop } from '@kiwicom/margarita-utils';
 import { type TripTypes, TRIP_TYPES } from '@kiwicom/margarita-config';
 import * as DateFNS from 'date-fns';
+import qs from 'qs';
 
 import { MODAL_TYPE } from './SearchConstants';
 
@@ -22,6 +23,28 @@ type Props = {|
   +bags?: string,
 |};
 
+type ParseFieldsParams = {|
+  +dateFrom?: ?string,
+  +dateTo?: ?string,
+  +returnDateFrom?: ?string,
+  +returnDateTo?: ?string,
+  +adults?: ?string,
+  +infants?: ?string,
+  +bags?: ?string,
+  +travelFrom?: ?string,
+  +travelTo?: ?string,
+|};
+
+type ParseFieldsReturn = {|
+  +dateFrom?: Date,
+  +dateTo?: Date,
+  +returnDateFrom?: Date,
+  +returnDateTo?: Date,
+  +adults?: number,
+  +infants?: number,
+  +bags?: number,
+|};
+
 export type ModalTypes = $Keys<typeof MODAL_TYPE>;
 export type PassengersData = {|
   adults: number,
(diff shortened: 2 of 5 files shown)


apps/web/pages/_app.js
```diff
@@ -8,15 +8,28 @@ import {
...
+
   render() {
-    const { Component, pageProps } = this.props;
+    const { Component, pageProps, router } = this.props;
     return (
       <Container>
         <UserContextProvider>
           <AlertContextProvider>
-            <SearchContextProvider>
+            <SearchContextProvider
```
tbergq: I don't think this should be done on each page. It should only be done on the results page.

aalises: We did it this way as it is how it's done on the `mobile` version. Also we need it on `Search` and `Results`. I have now added a check to only do this if there is `router.query`

tbergq: Sorry, I was specific enough here. 

I am ok with having the provider here. But maybe the provider should have a method, setStateFromQueryParams or something like that, and we just call that method from the search and result screen.

aalises: That looks great, putting it as an action, will do it now

apps/core/scenes/results/Results.js
```diff
@@ -52,9 +52,16 @@ type Props = {|
   +bags: number,
   +sortBy: string,
   +layout: number,
+  +router: any,
```
jvaclavik: Can't we use here something more specific?

aalises: Already changed it for the `router.query` (which is an object, not very specific either)

apps/core/scenes/results/Results.js
```diff
@@ -52,9 +52,16 @@ type Props = {|
   +bags: number,
   +sortBy: string,
   +layout: number,
+  +routerQuery: Object,
```
tbergq: Can you create a type for it. We should know what params we can get

apps/core/scenes/results/Results.js
```diff
@@ -52,9 +52,16 @@ type Props = {|
...
+  setStateFromQueryParams: string => void,
 |};
 
 class Results extends React.Component<Props> {
+  componentWillMount() {
```
tbergq: This lifecycle method is deprecated

apps/core/scenes/results/Results.js
aalises: Yes, I put it as `SearchParameter`

apps/core/scenes/results/Results.js
aalises: I placed it in `ComponentDidMount` now

apps/core/scenes/search/SearchContext.js
```diff
@@ -163,10 +153,23 @@ export default class SearchContextProvider extends React.Component<
         clearLocation: this.clearLocation,
         addLocation: this.addLocation,
         setLocation: this.setLocation,
+        setStateFromQueryParams: this.setStateFromQueryParams,
       },
     };
   }
 
+  parseFields = (params: Object): Object => {
```
tbergq: You should be able to give proper types for params and the return type as well

apps/core/scenes/search/SearchContext.js
```diff
@@ -178,6 +181,18 @@ export default class SearchContextProvider extends React.Component<
     });
   };
 
+  setStateFromQueryParams = (params: Object) => {
+    const parsedParams = qs.parse(params);
+    if (Object.keys(parsedParams).length) {
```
tbergq: Why this check? Is it not sufficient to test `parsedParms != null`?

aalises: Because `parsedParams` is never null, but `{}`

apps/core/scenes/search/SearchContext.js
```diff
@@ -178,6 +181,18 @@ export default class SearchContextProvider extends React.Component<
     });
   };
 
+  setStateFromQueryParams = (params: Object) => {
+    const parsedParams = qs.parse(params);
```
tbergq: qs.parse takes a string, so is params an object or as string? Could you give a type to parsedParams, so that we know what we are working with.

apps/core/scenes/search/SearchContext.js
tbergq: After you give a type to parsedParams, you should not need the if. Just do `Object.values(qs.parse(parsedParams?.travelFrom))`

apps/core/scenes/search/SearchContext.js
aalises: `Flow` won't let me as we are spreading `{ [key]: value }` so it does not recognize the spreaded keys (d'uh)

apps/core/scenes/search/SearchContext.js
aalises: Hmmm `parsedParams` is never null, we need this to only set the state if it actually has something...

apps/core/scenes/search/SearchContext.js
aalises: Done!

apps/core/scenes/search/SearchContext.js
tbergq: But as far as I can see, none of these calls will crash if paredParams = {}

aalises: Well that assigns `travelFrom` and `travelTo` to null instead of the default state. We can conditionally render them though but still need the Object keys check...

apps/web/pages/results.js
```diff
@@ -8,11 +8,11 @@ import { StyleSheet } from '@kiwicom/universal-components';
 import { withPageRouter } from '../components/withPageRouter';
 import Layout from '../components/Layout';
 
-const results = () => {
+const results = ({ router }) => {
```
tbergq: Also, please: 
```
import { type Router } from 'next/router';
type Props = {
  +router: Router,
};

const results = ({ router }: Props)
...

export default withPageRouter<{}>(results);
```

apps/core/scenes/search/SearchContext.js
tbergq: Well, if `dateFrom` is set but not `travelFrom`, travelFrom will still be null. The check with Object.keys.length does not unsure that travelFrom will be set correctly

Issue #689, opened 2019-04-17
Add Bags info to Itinerary type


JoseDuSV:
For baggage section of booking flow we need to extend Itinerary type with bags info.
As minimum for each type we will need:
- price
- limits (dimensions, weight)

Price and some basic limits info seems to be available on `Search API` and more detailed informations are part of the `Booking API`. Docs are just not always clear and sometimes mixing bags and luggage, but it should be possible to dig what we need from responses. 

<img width="374" alt="Screenshot 2019-04-17 at 16 07 55" src="https://user-images.githubusercontent.com/2660330/56294344-23958400-612b-11e9-8e12-ab4ae88ecb31.png">

PR #701, opened 2019-04-24
Feat: Added HoldBagOptions to Itinerary

JoseDuSV:
Extended `Itinerary` object with `HoldBagOptions`.
Tequila bags options are currently slightly limited, because there is only hold bags info available and bookable on Booking endpoints.
This part will be extended later, after more bags options and combinations will be added to endpoints.

Closes: #689 

_Note:_
This part is currently mainly defined/limited by available bags options on booking endpoint, where you can simply just get and book from 1 to 3 hold bags for now.
For now partners should/can use (as it is implemented in this PR): 
- **Search API**: `bags_price`, `baglimit`
- **Booking API**: `bags_price`, `luggage`
(hopefully this mismatching baglimit/luggage naming will be also unified)

After some research I found out that baggage part is currently in active development and should be extended in upcoming weeks. Recent info from baggage team: `Redesign of baggage ordering and the possibility to offer more types of paid baggage - will undergo integration testing in following days`

### DIFF
--- a/apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
+++ b/apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
@@ -11,13 +11,15 @@ import {
   getItineraryType,
   getItineraryDeparture,
   getItineraryArrival,
+  getHoldBagOptions,
+  sanitizeHoldBagProps,
   mapSectors,
   unmaskID,
 } from '../helpers/Itineraries';
 import type {
   ItinerariesReturnSearchParameters,
   ItinerariesOneWaySearchParameters,
-  ApiResponseType,
+  ItinerariesApiResponse,
   Itinerary,
 } from '../Itinerary';
 
@@ -74,7 +76,7 @@ const addReturnSearchQueryParams = inboundDate => {
 const fetchItineraries = async (
   parameters: $ReadOnlyArray<ItinerariesReturnSearchParameters>,
 ) => {
-  const results: $ReadOnlyArray<ApiResponseType> = await Promise.all(
+  const results: $ReadOnlyArray<ItinerariesApiResponse> = await Promise.all(
     parameters.map(params => {
       return fetch(`/v2/search?${qs.stringify(parseParameters(params))}`);
     }),
@@ -84,14 +86,21 @@ const fetchItineraries = async (
   });
 };
 
-const sanitizeItineraries = (response: ApiResponseType): Itinerary[] => {
+const sanitizeItineraries = (response: ItinerariesApiResponse): Itinerary[] => {
   const itineraries = response.data;
 
   return itineraries.map(itinerary => {
+    const currency = response.currency;
     const type = getItineraryType(itinerary.routes);
     const sectors = mapSectors(itinerary.route, itinerary.routes);
     const departure = getItineraryDeparture(sectors);
     const arrival = getItineraryArrival(type, sectors);
+    const holdBagProps = sanitizeHoldBagProps(itinerary.baglimit);
+    const holdBagOptions = getHoldBagOptions(
+      itinerary.bags_price,
+      currency,
+      holdBagProps,
+    );
 
     return {
       id: itinerary.id,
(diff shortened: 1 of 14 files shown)


apps/graphql/src/apps/itinerary/helpers/Itineraries.js
```diff
@@ -181,3 +186,55 @@ const sanitizeSegment = (segment: ?ApiRouteItem): Segment => {
...
+   * NOTE: For now there is only 1-3 available and bookable quantity for hold bags.
+   * This value is defined by current data available on endpoints and will be extended later
+   * with more combinations, after new bags implementation will be ready for booking
+   * endpoint https://docs.kiwi.com/booking/ (planned for Q2 2019)
+   */
+  const possibleHoldBagQuantities = [1, 2, 3];
+  const options = [];
+  const dimensions = parseBagDimensions(bagDimensions);
+  const weight = parseBagWeight(bagWeight);
+  for (const quantity of possibleHoldBagQuantities) {
```
aalises: Maybe `reduce` can be used here to avoid `break;`
```js
  const options = possibleHoldBagQuantities.reduce((acc, quantity) => {
    const priceAmount = prices?.[quantity.toString()];
    if (priceAmount == null) {
      return acc;
    }
    const option = {
      quantity,
      dimensions,
      weight,
      price: {
        currency: currency,
        amount: priceAmount,
      },
    };
    return [...acc, option];
  }, []);

```

apps/graphql/src/apps/itinerary/dataloaders/Itinerary.js
```diff
@@ -54,10 +55,21 @@ const sanitizeItinerary = (response: ItineraryApiResponse): Itinerary => {
...
    */
+  const currency = response.currency;
   const sectors = sanitizeSectors(response.flights);
   const type = getItineraryType(sectors);
   const departure = getItineraryDeparture(sectors);
   const arrival = getItineraryArrival(type, sectors);
+  const holdBagOptions = getHoldBagOptions(
+    response.bags_price,
+    currency,
+    response.luggage?.[4],
```
aalises: The Tequila API giving back an array of numbers and not an object with keys is 😞 ... maybe we can have a type `LuggageOptions` that is:
```
hand_length
hand_width
hand_height
hand_weight
hold_weight
hold_length
hold_width
hold_height
note = note of hold luggage
hand2_length
hand2_width
hand2_height
hand2_weight
hand2_note
hand_note
```
And have a method to map the array of luggage params to this object so we at least have it somewhat documented as fields not as positions (?¿)

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
JoseDuSV: Good idea, thanks, looks definitely better.

apps/graphql/src/apps/itinerary/dataloaders/Itinerary.js
JoseDuSV: I wanted to avoid fixing Tequila problems (because it will probably not end with just this 🙂). But you are right. Extra type and parser function in this case will not hurt and it will be more readable.
I'll update it.

apps/graphql/src/apps/itinerary/dataloaders/Itineraries.js
```diff
@@ -102,8 +114,9 @@ const sanitizeItineraries = (response: ApiResponseType): Itinerary[] => {
       departure,
       arrival,
       sectors,
+      holdBagOptions,
       price: {
-        currency: response.currency,
+        currency: currency,
```
tbergq: currency,

JoseDuSV: 👍 updated

apps/graphql/src/apps/itinerary/dataloaders/Itinerary.js
JoseDuSV: Refactored and added "sanitizer" which makes it little bit more readable.

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
JoseDuSV: Updated

PR #703, opened 2019-04-24
Refactor: Search Modals are no longer tied to the context

### LINKED ISSUE #700: Refactor Modals to not be directly connected to Context
Currently, Each `SearchForm`component (e.g `DatePickers`) is connected to the `SearchContext`. That means that every time that you update a value, it is also directly updated on the context, hence triggering a render on all components connected to it, and also triggering the `Search`API calls.

**Proposal:** Update the `SearchForm` to make all the values be set as the local state of the component. when clicking search, the `OnSubmit` callback is called with all the params, and then, after the validation passes (no empty fields, whatever), they are all set at once on the context.

This way the application is only updated / API calls are only made on clicking search, and not every time a parameter is changed, which creates unnecessary calls.

This way we can also clean up the context a little bit such as not having the modal types there.

WDYT @jvaclavik @tbergq @JoseDuSV @KarinaDavtyan @RobinCsl

aalises:
Summary: Get rid of the `setModalType` and modal types in the search context. Closes #700


apps/core/components/searchForm/PlacePicker/PlacePickerRenderer.js
```diff
@@ -5,22 +5,51 @@ import { QueryRenderer, graphql } from '@kiwicom/margarita-relay';
 
 import type { PlacePickerRendererQueryResponse } from './__generated__/PlacePickerRendererQuery.graphql';
 import PlacePickerContent from './PlacePickerContent';
+import {
+  MODAL_TYPE,
+  type modalType,
```
jvaclavik: I think it should be `ModalType` with uppercase on the first letter

aalises: Changed!

aalises:
> I would expect the entire concept of modal type to disappear. Each field should have its own modal.

Yes, we were also considering this possibility with @JoseDuSV .  It would be cleaner this way no @tbergq ?

JoseDuSV:
@aalises You can check `Picker.ios.js` component which also includes its own modal implementation.
I think it's reasonable to place it directly inside component/input if it's not affected by anything else.

tbergq:
> It would be cleaner this way no @tbergq ?

I would say it is the only(acceptable) way to do it.

aalises:
> I would say it is the only(acceptable) way to do it.

Done!

apps/core/components/searchForm/SearchFormModes.js
```diff
@@ -75,6 +95,25 @@ class SearchFormModes extends React.Component<Props> {
...
+        <Modal isVisible={this.state.isModalOpen} onClose={this.onClose}>
+          {this.state.formMode === FORM_MODE.TRIP_TYPE && (
+            <Select
+              optionsData={TRIP_TYPE}
+              selectedType={tripType}
+              onSelect={this.handleTripTypeSelect}
+              onClosePress={this.onClose}
+            />
+          )}
+          {this.state.formMode === FORM_MODE.PASSENGERS && (
```
JoseDuSV: I think this component is getting little bit bigger and obviously dealing with two different parts.
What about splitting it into separate `TripType` a `Passengers` related components? Each with its own logic and modal implementation?

apps/core/components/searchForm/SearchFormModes.js
```diff
@@ -75,6 +95,25 @@ class SearchFormModes extends React.Component<Props> {
...
         />
+        <Modal isVisible={this.state.isModalOpen} onClose={this.onClose}>
+          {this.state.formMode === FORM_MODE.TRIP_TYPE && (
+            <Select
+              optionsData={TRIP_TYPE}
+              selectedType={tripType}
+              onSelect={this.handleTripTypeSelect}
+              onClosePress={this.onClose}
+            />
+          )}
```
tbergq: Please also refactor this part. One modal for trip type and one modal for passengers input.

apps/core/components/searchForm/SearchFormModes.js
JoseDuSV: I see that Trond already suggested this one ⬆️. So I'll resolve this duplicate comment.

PR #704, opened 2019-04-24
Add gestures to RangeDatePicker

jvaclavik:
## Known issues (will be fixed in another PRs)
- **For now iOS only**. Next platforms will come later, there will be a lot of shared code. It works great on real device, but it's laggy in the simulator.
- There are few optimization issues (it's mentioned with `@TODO` in the code)
- Tests are missing
- There is issue during crossing months (when you cross months and `fingerPosition.y < spaceBetweenMonths + touchBuffer`)
- ~~**Types** - I was unable to find proper types for events from `react-native-gesture-handler` (any idea what it could be?)~~

ezgif com-video-to-gif

### DIFF
--- a/packages/universal-components/src/RangeDatePicker/DayNames.js
+++ b/packages/universal-components/src/RangeDatePicker/DayNames.js
@@ -8,10 +8,11 @@ import { format } from 'date-fns';
 import { StyleSheet } from '../PlatformStyleSheet';
 import { Text } from '../Text';
 import { designTokens } from '../DesignTokens';
-import { getCurrentWeekArray, type WeekStarts } from './libs';
+import { getCurrentWeekArray } from './libs';
+import type { WeekStartsType } from './RangeDatePickerTypes';
 
 type Props = {|
-  weekStartsOn: WeekStarts,
+  +weekStartsOn: WeekStartsType,
 |};
 
 const getDayNames = weekStartsOn =>
--- a/packages/universal-components/src/RangeDatePicker/DraggableItem.ios.js
+++ b/packages/universal-components/src/RangeDatePicker/DraggableItem.ios.js
@@ -1,0 +1,94 @@
+// @flow
+
+import React from 'react';
+import { PanGestureHandler, State } from 'react-native-gesture-handler';
+import { Animated } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import { designTokens } from '../DesignTokens';
+import { StyleSheet, type StylePropType } from '../PlatformStyleSheet';
+import type { GrabbedSideType, XY } from './RangeDatePickerTypes';
+import type { OnDragEvent, OnDropEvent } from '../types';
+
+type Props = {
+  +style?: StylePropType,
+  +onDrag: (OnDragEvent, GrabbedSideType) => void,
+  +onDrop: OnDropEvent => void,
+  +grabbedSide: GrabbedSideType,
+};
+
+export default class DraggableItem extends React.Component<Props> {
+  constructor(props: Props) {
+    super(props);
+    this.translateX = new Animated.Value(0);
+    this.translateY = new Animated.Value(0);
+    this.lastOffset = { x: 0, y: 0 };
+    this.onGestureEvent = Animated.event(
+      [
+        {
+          nativeEvent: {
+            translationX: this.translateX,
+            translationY: this.translateY,
+          },
+        },
+      ],
+      {
+        useNativeDriver: true,
+        listener: this.onDrag,
+      },
+    );
+  }
+
+  lastOffset: XY;
+  translateY: Animated.Value;
+  translateX: Animated.Value;
+  onGestureEvent: OnDragEvent;
+
+  onDrag = (event: OnDragEvent) => {
+    this.props.onDrag(event, this.props.grabbedSide);
+  };
+
+  onHandlerStateChange = (event: OnDropEvent) => {
+    if (event.nativeEvent.oldState === State.ACTIVE) {
+      this.props.onDrop(event);
+      this.lastOffset.x += event.nativeEvent.translationX;
+      this.lastOffset.y += event.nativeEvent.translationY;
+      this.translateX.setOffset(this.lastOffset.x);
+      this.translateX.setValue(0);
+      this.translateY.setOffset(this.lastOffset.y);
+      this.translateY.setValue(0);
+    }
+  };
+
+  render() {
+    return (
+      <PanGestureHandler
+        {...this.props}
+        onGestureEvent={this.onGestureEvent}
+        onHandlerStateChange={this.onHandlerStateChange}
+      >
+        <Animated.View
+          style={[
+            styles.box,
+            {
+              transform: [
+                { translateX: this.translateX },
+                { translateY: this.translateY },
+              ],
+            },
+            this.props.style,
+          ]}
+        />
+      </PanGestureHandler>
+    );
+  }
+}
+
+const styles = StyleSheet.create({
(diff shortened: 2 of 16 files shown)


tbergq:
You have a flow error 🙈

packages/universal-components/src/RangeDatePicker/DayNames.js
```diff
@@ -8,10 +8,11 @@ import { format } from 'date-fns';
 import { StyleSheet } from '../PlatformStyleSheet';
 import { Text } from '../Text';
 import { designTokens } from '../DesignTokens';
-import { getCurrentWeekArray, type WeekStarts } from './libs';
+import { getCurrentWeekArray } from './libs';
+import type { WeekStartsType } from './RangeDatePickerTypes';
 
 type Props = {|
-  weekStartsOn: WeekStarts,
+  weekStartsOn: WeekStartsType,
```
tbergq: could it be read-only?

packages/universal-components/src/RangeDatePicker/DraggableItem.ios.js
```diff
@@ -0,0 +1,101 @@
...
+      },
+    );
+  }
+
+  lastOffset: XY;
+  translateY: Animated.Value;
+  translateX: Animated.Value;
+  onGestureEvent: any;
+
+  onDrag = (event: any) => {
```
tbergq: Could we maybe define some event types, maybe it could be in `universal-components` 🤔 
something like [this](https://github.com/kiwicom/mobile/blob/master/packages/shared/types/Events.js)

packages/universal-components/src/RangeDatePicker/RenderDay.js
```diff
@@ -50,89 +62,150 @@ const RenderArrow = (props: RenderArrowProps) =>
...
+                      weight="bold"
+                      style={[
+                        styles.day,
+                        isDaySelected && styles.selectedDatesText,
+                        isDayInPast(day) && styles.dayInPast,
+                      ]}
+                    >
+                      {format(day, 'd')}
+                    </Text>
+                    {price ?? <DayPrice price={price} />}
```
tbergq: price is of type `?string` if price is equal to empty string, this crashes on android.

packages/universal-components/src/RangeDatePicker/libs.js
```diff
@@ -49,12 +66,217 @@ export const getMonthMatrix = (
...
+
+// @TODO add flow type for event
+export const findRelatedItem = (
+  event: any,
+  grabbedDay: Date,
+  selectedDates: $ReadOnlyArray<Date>,
+  dayItemSize: DayItemSizeType,
+  grabbedSide: GrabbedSideType,
+  weekStartsOn: WeekStartsType,
+  callback: (Array<Date>) => void,
```
tbergq: this method has too many parameters, it will be hard to use in the future. Could it take an object instead? 

You could consider doing it with other functions in this file as well.

packages/universal-components/src/RangeDatePicker/DraggableItem.ios.js
jvaclavik: I was also thinking about this solution.  But the event is quite big object:
```
{
  "_dispatchInstances": FiberNode {
    "tag": 7,
    "key": null,
    "type": "RCTView",
  },
  "_dispatchListeners": [Function bound value],
  "_targetInst": FiberNode {
    "tag": 7,
    "key": null,
    "type": "RCTView",
  },
  "bubbles": undefined,
  "cancelable": undefined,
  "currentTarget": 933,
  "defaultPrevented": undefined,
  "dispatchConfig": Object {
    "registrationName": "onGestureHandlerEvent",
  },
  "eventPhase": undefined,
  "isDefaultPrevented": [Function J],
  "isPropagationStopped": [Function J],
  "isTrusted": undefined,
  "nativeEvent": Object {
    "absoluteX": 294,
    "absoluteY": 285.5,
    "handlerTag": 2,
    "numberOfPointers": 1,
    "state": 4,
    "target": 933,
    "translationX": 111.5,
    "translationY": -12.5,
    "velocityX": 53.02986135148045,
    "velocityY": 0,
    "x": 18.5,
    "y": 17.5,
  },
  "target": 933,
  "timeStamp": 1556178367135,
  "type": undefined,
}
```

Would it be enough type just `nativeEvent` without exact type using `|`?

tbergq: Just add the values we are using, we can expand it as needed.

packages/universal-components/src/RangeDatePicker/DayNames.js
jvaclavik: Should be :) done

packages/universal-components/src/RangeDatePicker/RenderDay.js
jvaclavik: Thanks, I always forget about this one. Done

packages/universal-components/src/RangeDatePicker/libs.js
jvaclavik: Good point, done

packages/universal-components/src/RangeDatePicker/DraggableItem.ios.js
jvaclavik: OK, done

jvaclavik:
> You have a flow error 🙈

Fixed

packages/universal-components/src/RangeDatePicker/RenderDay.js
tbergq: Once you have spent 4 hours trying to find out why it works on ios and not on android, you never forget again 😄 🐘

packages/universal-components/src/RangeDatePicker/DraggableItem.ios.js
```diff
@@ -0,0 +1,94 @@
+// @flow
+
+import React from 'react';
+import { Animated } from 'react-native';
+import { PanGestureHandler, State } from 'react-native-gesture-handler';
```
JoseDuSV: This should be listed in dependencies. It is preinstalled in expo, but will cause problem when someone try to use universal components without expo.

packages/universal-components/src/RangeDatePicker/DraggableItem.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+import React from 'react';
+
+// @TODO Add implementations for web and Android
+export default class DraggableDay extends React.Component<{}> {
```
JoseDuSV: Can we name this `DraggableItem` for consistency?

jvaclavik: Oh, I forgot about this one during the renaming :) thanks

packages/universal-components/src/RangeDatePicker/DraggableItem.ios.js
jvaclavik: OK, done, it's already in Expo, so I added peerDependency to `universal-components`.

JoseDuSV: Ok, thanks. But do we need to add whole `expo` as dependency? Wouldn't be just previous `react-native-gesture-handler` enough?

jvaclavik: Right, if we use expo `react-native-gesture-handler` it shouldn't be installed again, but it's pity to add whole Expo dependency to universal-components. So I added `react-native-gesture-handler` to peerDependency and it should be fine now.

PR #705, opened 2019-04-25
Graphql: Enable gzip compression using apollo-server-express

tbergq:
Reduce size of returned json by 3.5 times


tbergq:
@aalises @RobinCsl @JoseDuSV please have another l👀k. It should be 🚀  now

PR #711, opened 2019-04-29
Feat: Add withHover HOC

### LINKED ISSUE #472: Refactor Hoverable component to HOC
Opened by @JoseDuSV 

> Refactor Hoverable component to HOC that injects isHovered prop.

Comment by @tbergq 

> How about creating a new HOC component, then in the existing Hoverable we can add a console.warn in componentDidMount, saying this component is deprecated, please switch to new HOC component. So we can make upgrading more smooth.

(Original issue https://github.com/kiwicom/universal-components/issues/209)

tbergq:
Summary: withHover HOC will provide us with a better way to check
if a component is hovered. Hoverable implementation was useless
and you had to do the work in each component using it anyways.
Also added a deprecation warning to hoverable component

closes #472

### DIFF
--- a/packages/universal-components/src/WithHover/__tests__/__snapshots__/withHover.test.js.snap
+++ b/packages/universal-components/src/WithHover/__tests__/__snapshots__/withHover.test.js.snap
@@ -1,0 +1,25 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Hoverable - native should match snapshot 1`] = `
+<View
+  onMouseEnter={[Function]}
+  onMouseLeave={[Function]}
+>
+  <View
+    isHovered={false}
+    testID="childView"
+  />
+</View>
+`;
+
+exports[`Hoverable - web should match snapshot 1`] = `
+<View
+  onMouseEnter={[Function]}
+  onMouseLeave={[Function]}
+>
+  <View
+    isHovered={false}
+    testID="childView"
+  />
+</View>
+`;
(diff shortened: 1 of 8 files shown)


packages/universal-components/src/WithHover/__tests__/__snapshots__/withHover.test.js.snap
```diff
@@ -0,0 +1,25 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Hoverable - native should match snapshot 1`] = `
+<View
+  onMouseEnter={[Function]}
```
JoseDuSV: When hover is available just on web, shouldn't this be just `<View>`?

packages/universal-components/src/WithHover/index.js
```diff
@@ -0,0 +1,3 @@
+// @flow
+
+export { default as withHover } from './withHover';
```
JoseDuSV: Can you please also add export to main `src/index.js`.

tbergq: Oops, I forgot that 🙈

packages/universal-components/src/WithHover/__tests__/__snapshots__/withHover.test.js.snap
tbergq: I don't think it matters. There is no mouse on mobile. So it should never be triggered.

tbergq:
@JoseDuSV 🚀

packages/universal-components/src/WithHover/__tests__/__snapshots__/withHover.test.js.snap
JoseDuSV: True, it shouldn't break anything.

PR #712, opened 2019-04-29
Unique GraphQL server file for dev and prod

RobinCsl:
Please note that it's necessary to change your local env to `GRAPHQL_URL=http://localhost:4000/graphql`.


aalises:
@RobinCsl We also need to add the generated `.webpack` to `.gitignore` no?

tbergq:
Yes, we should, and graphql no longer listens on `<local_ip>:4000/graphql` 😞

PR #713, opened 2019-04-29
Upgrade dependencies

tbergq:


apps/graphql/package.json
```diff
@@ -25,14 +25,15 @@
     "webpack": "^4.29.1"
   },
   "devDependencies": {
+    "@babel/plugin-proposal-decorators": "^7.4.4",
```
aalises: Out of curiosity, why do we need this now?

tbergq: 🤷‍♂ to run the gql server, it crashed without it saying it could not find it

PR #715, opened 2019-04-29
Add tests for RangeDatePicker libs

jvaclavik:
- Add tests
- Few naming changes and optimisations


packages/universal-components/src/RangeDatePicker/mocks/monthsWithDistributedWeeks.js
```diff
@@ -0,0 +1,147 @@
+// @flow
```
tbergq: I don't think this is a mock, this is a dataset.

jvaclavik: Renamed

jvaclavik:
rebased

PR #719, opened 2019-04-30
Refactor: Remove wrapping view from withHover

tbergq:
Summary:Wrapping view caused malformed style when applied to components.
Instead inject the onMouseLeave and onMouseEnter through the HOC.

### DIFF
--- a/apps/graphql/package.json
+++ b/apps/graphql/package.json
@@ -30,7 +30,8 @@
     "serverless-offline": "^4.9.4",
     "serverless-webpack": "^5.2.0",
     "uglifyjs-webpack-plugin": "^2.1.2",
-    "webpack": "^4.29.1"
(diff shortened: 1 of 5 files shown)


apps/graphql/package.json
```diff
@@ -30,7 +30,8 @@
     "serverless-offline": "^4.9.4",
     "serverless-webpack": "^5.2.0",
     "uglifyjs-webpack-plugin": "^2.1.2",
-    "webpack": "^4.29.1"
+    "webpack": "^4.29.1",
+    "webpack-node-externals": "^1.7.2"
```
RobinCsl: This is not really related, but okay :D

tbergq: Maybe not, but it still gave eslint error

PR #725, opened 2019-04-30
Upgrade @kiwicom/relay

tbergq:


tbergq:
> That was quick 🚀

⚡️

PR #729, opened 2019-05-02
Core: Add fragment test for ItineraryCard

tbergq:


tbergq:
It will confirm that the component renders and works. It is better than having no test at all. 

You could create a test further down in the graph to get more useful snapshots.

tbergq:
@RobinCsl switched the test for a fragment with smaller output. I could dive further down the graph if you like

RobinCsl:
@tbergq even though output is still quite large, I think it makes more sense this way.
I'm just trying to imagine how people usually update the snapshots without really inquiring what changed, and what would the optimal length of a snapshot for this behaviour not to be dominant.

tbergq:
Yeah, I would consider that at this level, as long as the output is still there, it does not change to be null, it is fine. 

Deeper down in the graph, there should be smaller snapshots better to determine impacts of changes. 

Anyway, here are two examples of how to test, which is what I wanted to show. 

Now we have no tests... 🙈

PR #730, opened 2019-05-02
Feature: Add SortTabs design for mobile

### LINKED ISSUE #678: Search filter options are cut on mobile
Image from iOS (3)

 - Device:  iPhone6

aalises:
Before:
image
Now:
Screenshot 2019-05-02 at 10 51 15

Closes #678

### DIFF
--- a/packages/components/src/sortTabs/SortTab.js
+++ b/packages/components/src/sortTabs/SortTab.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react';
-import { View } from 'react-native';
+import { View, Platform } from 'react-native';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import {
   Text,
@@ -49,6 +49,7 @@ class SortTab extends React.Component<Props> {
     const price = additionalInfo?.price;
     const currency = additionalInfo?.currency;
     const duration = additionalInfo?.duration;
+    const isMobile = Platform.OS !== 'web';
     return (
       <TouchableWithoutFeedback
         onPress={this.handleOnPress}
@@ -64,7 +65,7 @@ class SortTab extends React.Component<Props> {
             isHovered && styles.containerHovered,
           ]}
         >
-          <View>
+          <View style={isMobile && styles.rowContainer}>
             <Text
               style={[styles.label, isActive && styles.labelActive]}
               align="left"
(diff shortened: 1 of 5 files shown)


jvaclavik:
I think that the container for sorting is too big on mobile device. Can't we keep it in one line? Or what about do it as in kiwi.com app - in modal? What do you think?

packages/components/src/sortTabs/SortTab.js
```diff
@@ -110,18 +110,22 @@ class SortTab extends React.Component<Props> {
 export default withHover(SortTab);
 
 const borderRadius = parseInt(defaultTokens.borderRadiusSmall, 10);
-const containerHeight = 75;
+const containerHeight = {
```
JoseDuSV: It definitely looks better. But I'm just slightly worried about size on smaller screens, where there is not much space left for results.

Screenshot 2019-05-02 at 11 36 24

What about reducing content to just one line and remove unnecessary empty space. Something like.

Screenshot 2019-05-02 at 11 39 59

Or maybe even better, we can hide this content into badge which will open options inside modal, like in current mobile app. But we can probably save this iteration for later when/if we will have more filters.

Screenshot_20190502-114845

aalises:
@jvaclavik @JoseDuSV 
Screenshot 2019-05-02 at 14 00 45

jvaclavik:
OK, we can improve it later. 

Few more thing – what about not to have border radius there and also disable vertical margins of the box? Top gray margin also could have only 1px width. 

I also think that green left radius should be thicker. 

Screenshot 2019-05-02 at 14 58 59

aalises:
Say no more @jvaclavik . Also fixed an unrelated issue with `HoverMonitor` that arised on a previous merged PR.

Screenshot 2019-05-02 at 15 12 16

jvaclavik:
One more proposal :D

57078698-077b1080-6cef-11e9-8a2c-ced9575eaf9d

packages/components/src/sortTabs/SortTab.js
```diff
@@ -64,15 +65,20 @@ class SortTab extends React.Component<Props> {
             isHovered && styles.containerHovered,
           ]}
         >
-          <View>
+          <View style={[isMobile && styles.rowContainer]}>
```
jvaclavik: You don't need `[]` here

jvaclavik:
And now there are 3 types of borders: 

Screenshot 2019-05-02 at 15 47 04

I think border of this container could has the same style as border of result items. What do you think?

PR #731, opened 2019-05-02
Refactor: Get rid of Hoverable

aalises:
With the new `withHover` HOC, we do not need `Hoverable` anymore.


packages/universal-components/src/Hoverable/HoverMonitor.js
```diff
@@ -1,32 +0,0 @@
-// @flow
-
-const HoverMonitor = () => {
```
JoseDuSV: I think we need to include something similar to this into new `withHover`.
To prevent components being stuck in hover mode in web version on mobile/touch devices.
What we need is to basically disable hover events dispatching when there is touch detected.

It happens for example on current result cards.

hover

aalises: Oops sure! Added it back 💃

JoseDuSV: Thanks, now it works fine! 🚀

PR #733, opened 2019-05-02
Add gestures to RangeDatePicker (web)

jvaclavik:
ezgif com-optimize (1)

### DIFF
--- a/packages/universal-components/package.json
+++ b/packages/universal-components/package.json
@@ -33,6 +33,7 @@
     "@kiwicom/orbit-design-tokens": "^0.5.0",
     "@storybook/react-native": "^4.1.13",
     "memoize-one": "^5.0.0",
+    "react-draggable": "^3.3.0",
     "react-fast-compare": "^2.0.4",
     "react-native-modal": "^7.0.2",
     "react-native-multi-slider": "https://github.com/kiwicom/react-native-multi-slider.git#beb71bf5eb210fed393cb29b18032105b725611e",
--- a/packages/universal-components/src/RangeDatePicker/DraggableItem.web.js
+++ b/packages/universal-components/src/RangeDatePicker/DraggableItem.web.js
@@ -1,0 +1,72 @@
+// @flow
+
+import React from 'react';
+import { View } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+import { DraggableCore, type DraggableData, type Event } from 'react-draggable';
+
+import { designTokens } from '../DesignTokens';
+import type { GrabbedSideType, DayItemSizeType } from './RangeDatePickerTypes';
+import type { OnDragEvent } from '../types';
+
+type Props = {
+  +onDrag: (OnDragEvent, GrabbedSideType) => void,
+  +onDrop: () => void,
+  +grabbedSide: GrabbedSideType,
+  +dayItemSize: DayItemSizeType,
+};
+
+export default class DraggableItem extends React.Component<Props> {
+  handleDrag = (event: Event, data: DraggableData) => {
+    this.props.onDrag(
+      {
+        nativeEvent: {
+          translationX:
+            data.lastX + data.deltaX - this.props.dayItemSize.width / 2,
+          translationY:
+            data.lastY + data.deltaY - this.props.dayItemSize.height / 2,
+        },
+      },
+      this.props.grabbedSide,
+    );
+  };
+
+  handleStop = () => {
+    this.props.onDrop();
+  };
+
+  render() {
+    return (
+      <DraggableCore onDrag={this.handleDrag} onStop={this.handleStop}>
+        <View
+          style={
+            this.props.grabbedSide === 'left'
+              ? styles.draggableItemLeft
+              : styles.draggableItemRight
+          }
+        />
+      </DraggableCore>
+    );
+  }
+}
+
+const boxStyle = {
+  width: designTokens.widthDraggableCalendarItem,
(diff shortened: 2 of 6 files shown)


packages/universal-components/src/RangeDatePicker/DraggableItem.web.js
```diff
@@ -0,0 +1,57 @@
...
+  };
+
+  handleStop = () => {
+    this.props.onDrop();
+  };
+
+  render() {
+    return (
+      <Draggable onDrag={this.handleDrag} onStop={this.handleStop}>
+        <div
```
jvaclavik: This doesnt work with `<View />`, but I think it shouldn't be a problem

packages/universal-components/src/RangeDatePicker/DraggableItem.web.js
```diff
@@ -0,0 +1,57 @@
+// @flow
+
+import React from 'react';
```
JoseDuSV: Is this component working for you in this PR? When I try it `DraggableItem.js` is used on web instead of this one. It started working after I've renamed `DraggableItem.js` to `DraggableItem.android.js`. Then proper `DraggableItem.web.js` was used on web.

jvaclavik: Thanks for the info, that could be possible. I renamed it back  at the end because I wanted to make Android changes in different PR.

JoseDuSV: 👍 Now it works fine. Thanks!

jvaclavik:
@JoseDuSV Fixed and rebased

packages/universal-components/package.json
```diff
@@ -33,6 +33,7 @@
     "@kiwicom/orbit-design-tokens": "^0.5.0",
     "@storybook/react-native": "^4.1.13",
     "memoize-one": "^5.0.0",
+    "react-draggable": "^3.3.0",
```
JoseDuSV: I have some mixed feelings about this library. I understand that it saves some time than doing our implementation.
But it throws some errors into console after drag.
Screenshot 2019-05-03 at 10 23 34

It looks like related to this issue https://github.com/mzabriskie/react-draggable/issues/390

packages/universal-components/src/RangeDatePicker/DraggableItem.web.js
```diff
@@ -0,0 +1,57 @@
...
+  handleStop = () => {
+    this.props.onDrop();
+  };
+
+  render() {
+    return (
+      <Draggable onDrag={this.handleDrag} onStop={this.handleStop}>
+        <div
+          style={{
+            ...styles.box,
```
JoseDuSV: Do we need to recreate this style on each update? I see that it's mainly because of passed `this.props.style`. But cannot we just use `grabbedSide` prop and handle style internally (have for example some `boxLeft` and `boxRight` style object and use it directly)?

I'm just giving this for consideration, if it will be lot of refactoring I can probably live even with this approach now.

packages/universal-components/package.json
jvaclavik: Hmm, that's weird 😕it should be possible to use `DraggableCore` component which is more low-level than `Draggable`. I think it could solve this problem.

jvaclavik: Should be fixed (with `DraggableCore`).

packages/universal-components/src/RangeDatePicker/DraggableItem.web.js
jvaclavik: OK, done

JoseDuSV: Thanks, definitely better. But I think that we can go even further and get rid of this destructuring and new object creation (basically do it only once during initialisation).

You can for example extract `box` from `styles` and then just spread it into `draggableItemLeft` and `draggableItemRight`. After that you can use `draggableItemLeft/Right` directly.

packages/universal-components/package.json
JoseDuSV: Cool, seems to be ok now. I also like that this lighter component is enough for us. 
That this library is slightly overkill was part of my concern, but I feel better about it now 😉.
Thanks

packages/universal-components/src/RangeDatePicker/DraggableItem.web.js
jvaclavik: Good idea! Done

packages/universal-components/src/RangeDatePicker/DraggableItem.web.js
jvaclavik: Fixed with `DraggableCore`

packages/universal-components/src/RangeDatePicker/DraggableItem.web.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+const boxStyle = {
+  width: designTokens.widthDraggableCalendarItem,
+  height: designTokens.heightCalendarItem,
+  zIndex: parseFloat(defaultTokens.zIndexOnTheTop),
+  position: 'absolute',
+};
+const styles = {
```
JoseDuSV: While we are using `View` component we should use `StyleSheet.create(` to create styles only once. I think it will be perfect after that.

jvaclavik: Done

PR #735, opened 2019-05-02
Feat: Make whole body scrollable when there are search results (On web with mobile res)

### LINKED ISSUE #732: Whole web results page should be scrollable in mobile resolution
Search form currently fills almost whole screen and there is just tiny scrollable space for results (highlighted in green on screenshot).
Screenshot_20190502-115709

Possible solution: Set fixed position just to header (or maybe also hide it after you scroll down) and make whole screen including search form scrollable.

aalises:
Fixes #732 . Make all the body scrollable when there are search results instead of just the flat list, only in web version with mobile resolution. This allows seeing the results (before only a small frame appeared)

### DIFF
--- a/apps/core/scenes/results/Results.js
+++ b/apps/core/scenes/results/Results.js
@@ -165,6 +165,7 @@ class Results extends React.Component<Props> {
       },
     };
     const desktopLayout = this.props.layout >= LAYOUT.desktop;
+    const mobileLayout = this.props.layout < LAYOUT.largeMobile;
     return (
       <SafeAreaView style={styles.container}>
         {Platform.OS === 'web' ? (
@@ -192,7 +193,12 @@ class Results extends React.Component<Props> {
           />
         )}
 
-        <View style={styles.resultContainer}>
+        <View
+          style={[
+            styles.resultContainer,
+            !mobileLayout && styles.flexContainer,
+          ]}
+        >
           <SortTabsWrapper />
           <QueryComponent {...props} />
         </View>
@@ -204,6 +210,8 @@ class Results extends React.Component<Props> {
 const styles = StyleSheet.create({
   resultContainer: {
     backgroundColor: defaultTokens.backgroundBody,
+  },
+  flexContainer: {
     flex: 1,
   },
   container: {
--- a/apps/web/components/Layout.js
+++ b/apps/web/components/Layout.js
@@ -34,7 +34,9 @@ export const Layout = ({ layoutReady, children }: Props) => {
 const styles = StyleSheet.create({
   container: {
     flex: 1,
-    overflow: 'hidden',
+    web: {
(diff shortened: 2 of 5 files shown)


apps/core/scenes/results/Results.js
```diff
@@ -76,7 +76,13 @@ class Results extends React.Component<Props> {
     props: ReturnResultsQueryResponse | OneWayResultsQueryResponse,
   ) => {
     const { searchData } = props;
-    return <ResultsList data={searchData} onBookPress={this.handleBookPress} />;
+    return this.props.layout < LAYOUT.largeMobile ? (
+      <View>
```
JoseDuSV: It works but after some testing I'm slightly worried about performance.
Because by this we are displaying all results at once (instead of adding them during scroll), which can be a quite huge number. And by this website can become really slow for example during resize events.
So we will probably need to find some way how to keep current behaviour when cards are created during scroll but with whole page scrollable.

apps/web/components/Layout.js
```diff
@@ -34,7 +34,7 @@ export const Layout = ({ layoutReady, children }: Props) => {
 const styles = StyleSheet.create({
   container: {
     flex: 1,
-    overflow: 'hidden',
+    overflow: 'scroll',
```
JoseDuSV: Can we use `auto` instead? Scroll is ok on Mac, but on windows it will make scrollbar visible all the time.

apps/core/scenes/results/Results.js
```diff
@@ -76,7 +76,13 @@ class Results extends React.Component<Props> {
...
+        <ResultsList data={searchData} onBookPress={this.handleBookPress} />
+      </View>
```
JoseDuSV: I'm not sure about this "hack". It works, but I'm slightly worried that it is not that readable.
By this we are just using default `View` styles to prevent results from scrolling.
Cannot we use same condition to simply disable `flex: 1` style on `resultContainer` which will have a similar effect and will be more obvious?

aalises: @JoseDuSV Done!

apps/core/scenes/results/Results.js
```diff
@@ -165,6 +165,7 @@ class Results extends React.Component<Props> {
       },
     };
     const desktopLayout = this.props.layout >= LAYOUT.desktop;
+    const isMobile = this.props.layout < LAYOUT.largeMobile;
```
JoseDuSV: Just one last small thing and I believe we can merge this one after that.
Can we name this `mobileLayout` to keep it consistent with const line above? Or maybe use `desktopLayout` directly and make whole page scrollable even for tablet-like resolutions.

FilipMessa:
Do you have some information from Tequila devs, about the plan to implement the pagination?

Issue #736, opened 2019-05-03
Date input field should not be on two lines in Search screen


jvaclavik:
There should be probably three dots (or cropped content?):

<img width="1110" alt="Screenshot 2019-05-03 at 08 55 59" src="https://user-images.githubusercontent.com/858321/57123231-638e7500-6d81-11e9-8382-d5e11c53ba3a.png">

How it works on kiwi.com:
<img width="1265" alt="Screenshot 2019-05-03 at 08 59 50" src="https://user-images.githubusercontent.com/858321/57123390-ec0d1580-6d81-11e9-9c01-d30450177e6f.png">

JoseDuSV:
In this case I'll be slightly more for **crop** , with dots we will lose even more informations.
For me it's better to see at least `Thu 23 Ma` than `Thu 23 ...`.

PR #738, opened 2019-05-03
Fix: Text is now truncated on datepickers (only one line)

aalises:
Relates to #736

### DIFF
--- a/packages/components/src/tripInput/TripInput.js
+++ b/packages/components/src/tripInput/TripInput.js
@@ -38,7 +38,11 @@ export default function TripInput({
       <View style={[styles.container, style]}>
         {Platform.OS !== 'web' && <View style={styles.icon}>{inputIcon}</View>}
         <Text style={styles.label}>{label.length > 0 ? `${label}: ` : ''}</Text>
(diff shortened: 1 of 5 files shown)


packages/components/src/tripInput/TripInput.js
```diff
@@ -38,7 +38,11 @@ export default function TripInput({
       <View style={[styles.container, style]}>
         {Platform.OS !== 'web' && <View style={styles.icon}>{inputIcon}</View>}
         <Text style={styles.label}>{label.length > 0 ? `${label}: ` : ''}</Text>
-        {value !== '' && <Text style={styles.value}>{value}</Text>}
+        {value !== '' && (
+          <Text numberOfLines={1} ellipsizeMode="clip" style={styles.value}>
```
JoseDuSV: It doesn't seems to work on mobile. Text is not clipped when I try to set some smaller width to `TripInput`.

Screenshot 2019-05-03 at 12 50 24

aalises: @JoseDuSV That might be because the ellipsis is done on the internal text container..

PR #741, opened 2019-05-03
Feature: Put new mobile SortTabs design on web with mobile resolution

### LINKED ISSUE #740: Mobile SortTabs design should be also used for mobile web resolution
Currently we are using new compact tabs ( https://github.com/kiwicom/margarita/pull/730 ) only in mobile version.
Tabs should be switched to this design also on web with mobile layout.

Current state

Screenshot_20190503-150114

aalises:
The new SortTabs with the mobile design is also applied to the mobile resolution of the web version. Closes #740 .

### DIFF
--- a/packages/components/src/sortTabs/SortTab.js
+++ b/packages/components/src/sortTabs/SortTab.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react';
-import { View, Platform } from 'react-native';
+import { View } from 'react-native';
 import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import {
   Text,
@@ -26,6 +26,7 @@ export type Props = {|
   +isHovered: boolean,
   +onMouseLeave: () => void,
   +onMouseEnter: () => void,
+  +mobileLayout: boolean,
 |};
 
 class SortTab extends React.Component<Props> {
@@ -45,11 +46,11 @@ class SortTab extends React.Component<Props> {
       isHovered,
       onMouseEnter,
       onMouseLeave,
+      mobileLayout,
     } = this.props;
     const price = additionalInfo?.price;
     const currency = additionalInfo?.currency;
     const duration = additionalInfo?.duration;
-    const isMobile = Platform.OS !== 'web';
     return (
       <TouchableWithoutFeedback
         onPress={this.handleOnPress}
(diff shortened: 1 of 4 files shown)


packages/components/src/sortTabs/SortTab.js
```diff
@@ -155,20 +159,15 @@ const styles = StyleSheet.create({
...
-      borderStartColor: defaultTokens.paletteProductNormal,
-    },
-    android: {
-      borderStartWidth: 4,
-      borderStartColor: defaultTokens.paletteProductNormal,
-    },
+    borderBottomWidth: 2,
+    borderBottomColor: defaultTokens.paletteProductNormal,
+  },
+  containerActiveMobile: {
```
JoseDuSV: Looks like it will be necessary to also tweak container height on web.

Screenshot 2019-05-03 at 15 31 37

aalises: Screenshot 2019-05-06 at 09 05 40

Done!

packages/components/src/sortTabs/SortTab.js
```diff
@@ -76,7 +81,7 @@ class SortTab extends React.Component<Props> {
             <View
               style={[
                 styles.priceDurationContainer,
-                isMobile && styles.priceTagMobile,
+                mobileLayout && styles.priceTagMobile,
```
JoseDuSV: Can we unify naming of this container? We are currently using `priceDuration` and `priceTag` for the same component.

packages/components/src/sortTabs/SortTab.js
```diff
@@ -59,13 +60,17 @@ class SortTab extends React.Component<Props> {
         <View
           style={[
             styles.container,
-            isFirst && styles.containerFirst,
-            isLast && styles.containerLast,
-            isActive && styles.containerActive,
+            mobileLayout ? styles.heightMobile : styles.heightWeb,
```
JoseDuSV: I'm slightly worried how big this component become and if those conditions will be readable after a while.
What about splitting it into smaller pieces? For example this one can be something like `SortTabWrapper` which will render `View` with different styles based on `mobileLayout` prop. So it can be slightly more readable. Please just consider it, if it will actually make component more complex (lot of props passing etc.) I can probably live even with those complex conditions.

packages/components/src/sortTabs/SortTab.js
aalises: Done!

packages/components/src/sortTabs/SortTab.js
aalises: I think that for now, it's okay, there are not many conditions and they won't grow that much...

PR #747, opened 2019-05-06
Add limit parameter on Search

### LINKED ISSUE #746: Add limit parameter on Search API

From PR https://github.com/kiwicom/margarita/pull/735 , some performance problems arise when loading way too many results on screen and the body is scrollable (e.g on resizing the windows) on web version with mobile resolution.

**Describe the solution you'd like**

For solving it, we should consider implementing a sensible `limit` parameter on the search API call, for example, 50 results.

cc  @JoseDuSV

aalises:
Added limit parameter on search with a value of `50`. Closes #746

### DIFF
--- a/apps/core/scenes/search/SearchContext.js
+++ b/apps/core/scenes/search/SearchContext.js
@@ -2,7 +2,12 @@
 
 import * as React from 'react';
 import { withContext, noop } from '@kiwicom/margarita-utils';
-import { type TripTypes, TRIP_TYPES } from '@kiwicom/margarita-config';
+import {
+  type TripTypes,
+  TRIP_TYPES,
+  SEARCH_RESULTS_LIMIT,
+  DEFAULT_SEARCH_SORTING,
+} from '@kiwicom/margarita-config';
 import * as DateFNS from 'date-fns';
 import qs from 'qs';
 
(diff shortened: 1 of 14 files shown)


FilipMessa:
What is the motivation behind this?

aalises:
@FilipMessa So now on web version, when on a mobile resolution, the whole body is scrollable. That creates performance issues on resizing, for example, as we are re-rendering 200 cards. This is a *temporary* solution as discussed with @JoseDuSV for optimization purposes. Ideally, when pagination is implemented in Tequila, we should implement it with pagination with a load more button for example.

apps/core/scenes/search/SearchContext.js
```diff
@@ -89,7 +91,7 @@ type State = {|
 const defaultDepartureDate = DateFNS.addDays(new Date(), 1);
 const defaultReturnDate = DateFNS.addDays(defaultDepartureDate, 2);
 const defaultSortBy = 'QUALITY';
-
+const defaultLimit = 50;
```
jvaclavik: Maybe we should move default values to `margarita-config`, what do you think?

aalises: @jvaclavik done!

PR #748, opened 2019-05-06
Style and refactor RangeDatePicker

jvaclavik:
- Unify `DayItemArrow` and `AnimatedDayItemArrow`. There is no reason to have it in separated components
- Separate arrow from Android to component called `SimpleArrow`
- Folder started to be messy, so I refactored the structure a bit
- Better UX on the web (discussed about it with Denis Rojcyk)
  - On the web version, arrows are also clickable, not only draggable (it's like on kiwi.com)
  - Arrow and draggable points are not visible if the click would go to the past (it's like on kiwi.com)
  - Update design of the arrows
Screenshot 2019-05-06 at 15 54 39


packages/universal-components/src/RangeDatePicker/RangeDatePicker.js
```diff
@@ -73,6 +73,11 @@ const styles = StyleSheet.create({
     backgroundColor: 'white',
     borderRadius: parseFloat(defaultTokens.borderRadiusBadge),
     overflow: 'hidden',
+    flex: 1,
+    web: {
+      marginVertical: 12,
```
KarinaDavtyan: Could be replaced for defaultTokens.spaceSmall

packages/universal-components/src/RangeDatePicker/components/DayItemArrow.web.js
```diff
@@ -0,0 +1,74 @@
...
+import { View } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import { StyleSheet, type StylePropType } from '../../PlatformStyleSheet';
+import { ExtendedTouchable } from '../../ExtendedTouchable';
+import { SimpleArrow } from './SimpleArrow';
+
+export type Props = {|
+  +style?: StylePropType,
+  +onPress?: () => void,
```
FilipMessa: The onPress should be required, I suppose

packages/universal-components/src/RangeDatePicker/components/DayItemArrow.web.js
```diff
@@ -0,0 +1,74 @@
...
+    zIndex: parseFloat(defaultTokens.zIndexDefault),
+    backgroundColor: defaultTokens.backgroundBadgeDark,
+    borderRadius: parseFloat(defaultTokens.borderRadiusSmall),
+    height: 40,
+    width: 12,
+    position: 'absolute',
+  },
+  containerShadowLeft: {
+    web: {
+      boxShadow: '3px 0 2px -1px rgba(0,0,0,0.3)',
```
FilipMessa: Aren't these shadows values in designTokone?

packages/universal-components/src/RangeDatePicker/components/RenderDay.js
```diff
@@ -147,10 +136,12 @@ export default class RenderDay extends React.Component<Props, State> {
       >
         <View>
           {isRangePicker &&
+            isPossibleToChangeDateThisDirection &&
             (isStartOfSelectedDates || this.state.isDragging) && (
```
FilipMessa: This big condition statement scares me :), there must be a way how to write this in a more readable way.

packages/universal-components/src/RangeDatePicker/components/DayItemArrow.web.js
jvaclavik: Actually I'm not sure about that. I'm going to remove this `onPress` when I will implement gestures on Android. Then it would be duplicity, because you could grab invisible `DraggableItem` or click on `DayItemArrow` which is rendered behind the `DraggableItem`.

packages/universal-components/src/RangeDatePicker/components/DayItemArrow.web.js
jvaclavik: No

packages/universal-components/src/RangeDatePicker/components/RenderDay.js
jvaclavik: I tried to make it better. But it's still quite complicated, it has to solve lot of cases.

packages/universal-components/src/RangeDatePicker/RangeDatePicker.js

packages/universal-components/src/RangeDatePicker/components/DayItemArrow.web.js
jvaclavik: I'm not sure if it should be there, but it seems to me very component-specific.

packages/universal-components/src/RangeDatePicker/components/DayItemArrow.web.js
FilipMessa: OK make sense to me

PR #750, opened 2019-05-07
Add ControlContainer to RangeDatePicker

jvaclavik:
Screenshot 2019-05-07 at 12 38 18

### DIFF
--- a/packages/universal-components/src/RangeDatePicker/components/ControlContainer.js
+++ b/packages/universal-components/src/RangeDatePicker/components/ControlContainer.js
@@ -1,0 +1,35 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { format, isSameDay } from 'date-fns';
+
+import { StyleSheet } from '../../PlatformStyleSheet';
+import ControlContainerItem from './ControlContainerItem';
+
+type Props = {|
+  +label: string,
+  +dates: $ReadOnlyArray<Date>,
+  +daysInDestination?: ?$ReadOnlyArray<Date>,
+  +dateFormat: string,
+|};
+
+export default class ControlContainer extends React.Component<Props> {
+  render() {
+    const { label, dates, dateFormat } = this.props;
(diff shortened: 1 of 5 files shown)


packages/universal-components/src/RangeDatePicker/components/ControlContainer.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { format, isSameDay } from 'date-fns';
+
+import { StyleSheet } from '../../PlatformStyleSheet';
+import ControlContainerItem from './ControlContainerItem';
+
+export const LONG_DAY_MONTH_FORMAT = 'EEE d MMM'; // @TODO add to config
```
jvaclavik: `universal-components` should not have dependency to `margarita-config`. We should probably create new config just for universal components

JoseDuSV: And cannot this string be specified in component props? Something similar like texts/labels passed in `labels` to make it even more variable.

jvaclavik: It could be specified in props, but I use this component inside of many other components and it would have to bubble across many layers which does not seem me right.

JoseDuSV: As I see you're already passing `label` string in props, which looks like similar case to me.
But if it will be a problem we can probably just add `constants.js` to component root and put this string there (like for example in case of `TagsInput` component). I don't think we will need to reuse it in any other component right now.

jvaclavik: You are right! I didn't realise that this component is rendered directly in `RangeDatePicker`. In this case it's totally fine, so I added it to the props.

JoseDuSV: 👍 Thanks, it looks fine now.

PR #754, opened 2019-05-09
Chore: add create-dotenv for GraphQL deploy

aalises:


.circleci/config.yml
```diff
@@ -102,6 +102,9 @@ jobs:
           command: |
             sudo npm i -g serverless
             yarn install
+      - run:
+          name: Add .env file (react-native-dotenv)
```
JoseDuSV: Can you please just remove that `(react-native-dotenv)` it is not relevant here. We can keep just `Add .env file`

aalises: sure!

PR #760, opened 2019-05-10
Chore: Generate custom .env file for GraphQL

aalises:
Generating only the necessary fields on the `.env` to not overflow the limit set by AWS Lambda on environment variables (4kb)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,8 +103,8 @@ jobs:
             sudo npm i -g serverless
             yarn install
       - run:
-          name: Add .env file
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -103,8 +103,8 @@ jobs:
             sudo npm i -g serverless
             yarn install
       - run:
-          name: Add .env file
-          command: yarn create-dotenv
```
RobinCsl: Could you explain what's the difference here?
This script is meant to dump everything into a .env file.

PR #761, opened 2019-05-10
Add nightsInDestination to RangeDatePicker

jvaclavik:
Ready for CR!

ezgif com-optimize (6)

### DIFF
--- a/apps/core/scenes/results/Results.js
+++ b/apps/core/scenes/results/Results.js
@@ -53,6 +53,9 @@ type Props = {|
   +sortBy: string,
   +layout: number,
   +limit: number,
+  +nightsInDestinationFrom: string,
+  +nightsInDestinationTo: string,
+  +isNightsInDestinationSelected: boolean,
   +routerQuery: SearchParameters,
   +setStateFromQueryParams: SearchParameters => void,
 |};
@@ -91,6 +94,9 @@ class Results extends React.Component<Props> {
       infants,
       sortBy,
       limit,
+      nightsInDestinationFrom,
+      nightsInDestinationTo,
+      isNightsInDestinationSelected,
     } = this.props;
     return {
       passengers: {
@@ -111,12 +117,19 @@ class Results extends React.Component<Props> {
           end: DateFNS.format(dateTo, BASIC_ISO_DATE_FORMAT),
         },
         ...(type === TRIP_TYPES.RETURN
-          ? {
-              inboundDate: {
-                start: DateFNS.format(returnDateFrom, BASIC_ISO_DATE_FORMAT),
-                end: DateFNS.format(returnDateTo, BASIC_ISO_DATE_FORMAT),
-              },
-            }
+          ? !isNightsInDestinationSelected
+            ? {
+                inboundDate: {
+                  start: DateFNS.format(returnDateFrom, BASIC_ISO_DATE_FORMAT),
+                  end: DateFNS.format(returnDateTo, BASIC_ISO_DATE_FORMAT),
+                },
+              }
+            : {
+                nightsInDestination: {
+                  from: parseInt(nightsInDestinationFrom, 10),
+                  to: parseInt(nightsInDestinationTo, 10),
+                },
+              }
           : {}),
       },
     };
@@ -248,6 +261,9 @@ const select = ({
   infants,
   sortBy,
   limit,
+  nightsInDestinationFrom,
+  nightsInDestinationTo,
+  isNightsInDestinationSelected,
   actions: { setStateFromQueryParams },
 }: SearchContextState) => ({
   travelFrom,
@@ -261,6 +277,9 @@ const select = ({
   infants,
   sortBy,
   limit,
+  nightsInDestinationFrom,
+  nightsInDestinationTo,
+  isNightsInDestinationSelected,
   setStateFromQueryParams,
 });
 
--- a/packages/universal-components/src/RangeDatePicker/NightsInDestination.js
+++ b/packages/universal-components/src/RangeDatePicker/NightsInDestination.js
@@ -1,0 +1,49 @@
+// @flow
+
+import * as React from 'react';
+import { View } from 'react-native';
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+import { Slider } from '../Slider';
+import { StyleSheet } from '../PlatformStyleSheet';
+
+type Props = {|
+  +onNightsInDestinationChange?: ?(Array<number>) => void,
+  +nightsInDestination?: ?$ReadOnlyArray<number>,
+  +nightsInDestinationLabel?: string,
+|};
+
+export default class NightsInDestination extends React.Component<Props> {
+  render() {
+    return (
+      this.props.nightsInDestination != null &&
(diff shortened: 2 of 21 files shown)


apps/core/scenes/search/SearchContext.js
```diff
@@ -93,6 +105,8 @@ type State = {|
   },
 |};
 
+const defaultNightsInDestinationFrom = '3';
```
aalises: Maybe we could move these default values to `@margarita/config` (as the default sort or the result limits)

apps/core/scenes/results/Results.js
```diff
@@ -54,6 +54,9 @@ type Props = {|
   +sortBy: string,
   +layout: number,
   +limit: number,
+  +nightsInDestinationFrom: string,
```
JoseDuSV: Does it needs to be a string? I see that you are then parsing it to number on multiple places. And same prop in `ResultsScreen` already have `number` type, so it will be nice to unify it.
Same with `nightsInDestinationTo`.

packages/universal-components/src/RangeDatePicker/NightsInDestination.js
```diff
@@ -0,0 +1,48 @@
...
+        <View style={styles.container}>
+          <View style={styles.container} />
+          <View style={styles.slider}>
+            <Slider
+              minValue={1}
+              maxValue={30}
+              startValue={this.props.nightsInDestination[0]}
+              endValue={this.props.nightsInDestination[1]}
+              step={1}
+              label="Nights in destination"
```
JoseDuSV: We should pass this label as prop like in other components inside `universal`.

packages/universal-components/src/RangeDatePicker/RangeDatePicker.js
```diff
@@ -45,34 +72,48 @@ export default class RangeDatePicker extends React.Component<Props> {
...
-            weekStartsOn={weekStartsOn}
-            isRangePicker={isRangePicker ?? true}
-          />
+          {!isNightsInDestinationSelected || !isNightsInDestinationVisible ? (
+            <RangeDatePickerContent
+              selectedDates={this.props.dates}
+              onDayPress={this.handleChangeDate}
+              numberOfRenderedMonths={numberOfRenderedMonths}
+              weekStartsOn={weekStartsOn}
+              isRangePicker={isRangePicker ?? true}
```
JoseDuSV: Can this prop be false if this component is called `RangeDatePicker`?

packages/universal-components/src/RangeDatePicker/RangeDatePicker.js
```diff
@@ -45,34 +72,48 @@ export default class RangeDatePicker extends React.Component<Props> {
...
-                dates={dates}
-                dateFormat={dateFormat}
-              />
-            </View>
+            {isControlContainerVisible && (
+              <View style={[styles.controlContainer]}>
+                <ControlContainer
+                  dates={dates}
+                  dateFormat={dateFormat}
+                  labelNightsInDestination={'Nights in destination'}
```
JoseDuSV: Label should be passed in prop.

packages/universal-components/src/RangeDatePicker/components/ControlContainerItem.js
```diff
@@ -5,48 +5,82 @@ import { View } from 'react-native';
...
+  flexContainer: {
+    flex: 1,
+  },
   outerContainer: {
     flex: 1,
   },
   borderedContainer: {
     borderWidth: parseFloat(defaultTokens.borderWidthInput),
-    borderColor: defaultTokens.borderColorButtonInfoBordered,
+    borderColor: defaultTokens.colorTextSecondary, // @TODO better name in defaultTokens
```
JoseDuSV: What about using more universal `paletteInkLight` instead of text color value?

packages/universal-components/src/RangeDatePicker/NightsInDestination.js
```diff
@@ -0,0 +1,48 @@
...
+
+export default class NightsInDestination extends React.Component<Props> {
+  render() {
+    return (
+      this.props.nightsInDestination != null &&
+      this.props.onNightsInDestinationChange != null && (
+        <View style={styles.container}>
+          <View style={styles.container} />
+          <View style={styles.slider}>
+            <Slider
```
JoseDuSV: This is probably `Slider` problem. So it definitely doesn't need to be part of this PR.
But currently I cannot select values with 1 step distance (for example `11-12` range).
Can we at least create issue for it?

limit-problem

Another thing which can be saved for separate issue/PR. Selecting just single value instead of range should be possible (like on Kiwi.com).

single-val

packages/universal-components/src/RangeDatePicker/RangeDatePicker.js
jvaclavik: I see :D 

If:
```
isRangePicker == true // it's possible to choose date range 
isRangePicker == false // it's possible to choose only one date (will be used for Hotels in Kiwi app)
```

Maybe there should be different naming and the best would be to rename whole `RangeDatePicker` to `DatePicker`, but it should be different PR. And I'm also not convinced if it worth for it. And there is also different `DatePicker` in `universal-components`

Maybe we could just rename this prop. For example `isSingleDateOnly`. What do you think?

JoseDuSV: Ok, thanks for explanation. I don't think that component renaming is necessary.
Just changing prop should be fine. We can have `type` like in case of `Slider`. Or something like you proposed is also fine.
Adding storybook with playground to these new components later will also help. So user will be able to try what each prop actually does.

jvaclavik: Yes, totally agree. There is already issue for adding RangeDatePicker to storybook (#749). I will focus on it after landing page, which has right now bigger priority.

packages/universal-components/src/RangeDatePicker/RangeDatePicker.js

packages/universal-components/src/RangeDatePicker/components/ControlContainerItem.js
jvaclavik: Done

apps/core/scenes/results/Results.js
jvaclavik: Done

apps/core/scenes/search/SearchContext.js
jvaclavik: Done

packages/universal-components/src/RangeDatePicker/NightsInDestination.js
jvaclavik: Done

packages/universal-components/src/RangeDatePicker/NightsInDestination.js
jvaclavik: #762  
#807

apps/core/scenes/results/Results.js
jvaclavik: Done

PR #767, opened 2019-05-13
Feature: Add anywhere text when no location selected on PlacePicker

### LINKED ISSUE #766: Add anywhere option value into the SearchForm's place pickers
As a user, I can choose anywhere option from the SearchForm's PlacePicker.
If as a user, I don't select any location,  search location input field will be set to anywhere. 

ezgif com-optimize

It should look like this:
Screen Shot 2019-05-13 at 15 50 39

aalises:
Anywhere is placed instead on the PlacePickers when no place is selected. Due to limitations on `Tequila`, both places can't be empty at the same time but one of them can (either destination or origin), so a check that both aren't empty is performed on submit. Closes #766 

May-13-2019 17-25-33

### DIFF
--- a/packages/components/src/tripInput/TripInput.js
+++ b/packages/components/src/tripInput/TripInput.js
@@ -22,6 +22,10 @@ type Props = {|
   +placeholder?: string,
 |};
 
+const getEmptyLocationPlaceholder = (field: string): string => {
+  return field === 'From' ? '' : 'Anywhere';
+};
+
 export default function TripInput({
   style,
   icon,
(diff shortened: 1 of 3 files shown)


FilipMessa:
That state when origin and destination are equal is anywhere, is not very nice from UX. Probably we can do that like it is on kiwi.com, and require that origin must be chosen, and just destination can have value anywhere. 
What do you think? @aalises

aalises:
@FilipMessa That was how it was done before, so just reverting back the condition on `SearchForm` would do it, shall I?

FilipMessa:
@aalises Yes, but I would like to keep anywhere label for the destination location.

aalises:
May-14-2019 10-59-48

packages/components/src/tripInput/TripInput.js
```diff
@@ -22,6 +22,10 @@ type Props = {|
   +placeholder?: string,
 |};
 
+const getEmptyLocationPlaceholder = (field: string): string => {
```
FilipMessa: What about `getAnywherePlaceholder`? Just an idea

aalises: Would also be nice! but in the case of `From` it's not really anywhere... hahaha, we can change it later!

PR #770, opened 2019-05-14
Landing page: Add menu and style features

jvaclavik:


aalises:
Header looks a little bit weird when resizing on web :o
eeeh

apps/landingPage/components/Content.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+import styled from 'styled-components';
+
+import { BREAKPOINTS } from '../mediaQueriesConfig';
+
+export default styled.div`
+  max-width: 80vw;
+  position: relative;
+  @media (min-width: ${BREAKPOINTS.BIG_MOBILE}px) {
```
KarinaDavtyan:
```suggestion
  @media (min-width: ${defaultTokens.widthBreakpointLargeMobile) {
```

apps/landingPage/components/Footer.js
```diff
@@ -79,20 +105,28 @@ const FooterContainer = styled.div`
   }
 `;
 
+const LinksWrapper = styled.div`
+  display: flex;
+  justify-content: space-around;
+  padding: 40px 0 10px 0;
+  flex-wrap: wrap;
+  @media (min-width: ${BREAKPOINTS.TABLET}px) {
```
KarinaDavtyan: Maybe removing for now if its empty

apps/landingPage/components/Header.js
```diff
@@ -50,115 +50,108 @@ const logos = [
...
-    </Container>
+    <>
+      <Container>
+        <Content>
+          <UpperHeader>
+            <Logo src="/static/logo_white.png" alt="kiwi.com logo" />
+          </UpperHeader>
+          <Columns>
+            <Headline>
+              <Motto inverted={true} element="div">
```
KarinaDavtyan: Why do we need to remove <Headline /> from orbit here?

KarinaDavtyan:
Maybe `Content.js`, `AccentedText.js`, `Highlight.js` should go to `styledComponent/` (or any other similar name) folder? I think it would be more convenient to have separate folder for just styled components

apps/landingPage/components/Content.js
jvaclavik: OK

apps/landingPage/components/Header.js
jvaclavik: We need different font and different styling

apps/landingPage/components/Footer.js
jvaclavik: Done

jvaclavik:
> Header looks a little bit weird when resizing on web :o
> eeeh

It will be solved in another PR

PR #777, opened 2019-05-15
Refactor LogoCards

KarinaDavtyan:
Remove logos config from the Header. So know all the LogoCards related code is in the components/LogoCards


apps/landingPage/components/LogoCards.js
```diff
@@ -0,0 +1,77 @@
...
+import { defaultTokens } from '@kiwicom/orbit-design-tokens';
+
+export default function LogoCard() {
+  return (
+    <Container>
+      {logos.map(logo => (
+        <CardWrapper key={logo.title}>
+          {logo.icon}
+          <Column>
+            <Heading type="title2" spaceAfter="smallest">
```
jvaclavik: Can you use for `Heading` and `Text` "Circlular pro" font?

apps/landingPage/components/LogoCards.js
```diff
@@ -0,0 +1,77 @@
...
+            </Heading>
+            <Text weight="bold">{logo.text}</Text>
+          </Column>
+        </CardWrapper>
+      ))}
+    </Container>
+  );
+}
+
+const phoneWidthBuffer = 280;
```
jvaclavik: I guess this cont should be shared, we will use it also in different component

Issue #771, opened 2019-05-14
Add a Passengers to Booking


FilipMessa:
On the result detail page, we currently have mocked passengers data.
We need removed mocked and add the possibility to add new passengers and edit already added.

1. After click on the Add button, open the PassengerForm
2. A user can fill fields and Save (context or state)
3. Render Added Passenger
4. A user can edit or delete passenger

<img width="1659" alt="Screen Shot 2019-05-14 at 13 53 48" src="https://user-images.githubusercontent.com/6354326/57695889-bbfe2600-764f-11e9-81a8-b336bf0fd2de.png">

PR #779, opened 2019-05-15
Feature: Add and Delete passengers from a Booking

aalises:
Adding and deleting Passengers is now possible. Relates to #771  

*Validation: Name and ID are mandatory, and the ID has to be unique*. 

May-15-2019 15-03-03

### DIFF
--- a/apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
+++ b/apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
@@ -1,7 +1,13 @@
 // @flow
 
 import * as React from 'react';
-import { PassengerCards } from '@kiwicom/margarita-components';
+import {
+  PassengerCards,
+  type PassengerCardType,
+  withAlertContext,
+  type AlertContent,
+  type AlertContextState,
+} from '@kiwicom/margarita-components';
 import { createFragmentContainer, graphql } from '@kiwicom/margarita-relay';
 
 import type { ResultDetailPassenger_itinerary as ResultDetailPassengerType } from './__generated__/ResultDetailPassenger_itinerary.graphql';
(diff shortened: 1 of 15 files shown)


packages/components/src/passengerCard/__tests__/__snapshots__/PassengerCard.test.js.snap
```diff
@@ -146,7 +146,7 @@ Object {
         <PassengerCardDetail
           label="Date of birth"
           style="normal"
-          value="22/04/1980"
+          value="1980-04-12"
```
JoseDuSV: Is this change related to format used in MMB? I think that so far we are using `DD/MM/YYYY` everywhere and it will be reasonable to keep it unified.

aalises: @JoseDuSV Sure! I just didn't recall which format was in the config 😆So I put the first one

FilipMessa:
And what about the edit? It will be in the next PR?

aalises:
@FilipMessa Yes! That was the plan, to get this right as the edit will be done similarly.

FilipMessa:
OK, so in that case, there should not be a comment for close the issue. 
I'll check the code

apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
```diff
@@ -1,53 +1,108 @@
...
 |};
 
 type State = {|
   +isFormVisible: boolean,
+  +passengerData: Array<PassengerCardType>,
 |};
 
-// @TODO - temp passengers info, load default values from user profile or set empty for uregistered
-const passengerCards = [
+const defaultPassengers = [
```
FilipMessa: this should not be mocked in future som pls add some TODO comment there, THANKS

apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
```diff
@@ -1,53 +1,108 @@
...
-    this.setState({ isFormVisible: false });
+  handleFormSaveRequest = passenger => {
+    const existsID: boolean = !!this.state.passengerData.find(
+      el => el.id === passenger.id,
+    );
+    const validateForm = () => {
+      if (!(passenger.id && passenger.name)) {
+        return 'ID and Name are mandatory';
+      } else if (existsID) {
+        return 'A passenger with the same ID has already been added. Please choose another one';
```
FilipMessa:
```suggestion
        return 'A passenger with the same ID has already been added.';
```

packages/components/src/passengerCard/PassengerCard.js
```diff
@@ -64,13 +72,15 @@ class PassengerCard extends React.Component<Props> {
       passengerCount,
       bags,
       actionIconName,
+      secondaryActionIconName,
       onActionPress,
+      onSecondaryActionPress,
```
FilipMessa: Are we using this card for something else? Can we just rename it `onEditPress` and `onDeletePress`?

aalises: All done!

FilipMessa: Thanks!

PR #788, opened 2019-05-16
Feature: Implement edit passenger feature

### LINKED ISSUE #771: Add a Passengers to Booking

aalises:
Closes #771 
May-16-2019 13-12-56

### DIFF
--- a/apps/core/components/passengerForm/PassengerForm.js
+++ b/apps/core/components/passengerForm/PassengerForm.js
@@ -27,7 +27,9 @@ import type { PassengerForm_itinerary as PassengerFormType } from './__generated
 type Props = {|
   +itinerary: ?PassengerFormType,
   +isVisible: boolean,
+  +isEditing: boolean,
   +onRequestClose: () => void,
+  +prefillData: ?PassengerCardType,
   +onRequestSave: PassengerCardType => void,
   +setAlertContent: (alertContent: AlertContent | null) => void,
 |};
@@ -48,22 +50,42 @@ const nationalityData = [
   },
 ];
 
+const initialFormState = {
+  gender: 'male',
+  name: null,
+  lastName: null,
+  id: null,
+  nationality: null,
+  dateOfBirth: null,
+  bags: null,
+  passengerCount: 1,
+};
+
 type State = {
   ...PassengerCardType,
   lastName: ?string,
+  hasPrefilledState: boolean,
 };
 
 class PassengerForm extends React.Component<Props, State> {
   state = {
-    gender: 'male',
-    name: null,
-    lastName: null,
-    id: null,
-    nationality: null,
-    dateOfBirth: null,
-    bags: null,
-    passengerCount: 1,
+    ...initialFormState,
+    hasPrefilledState: false,
   };
+
+  static getDerivedStateFromProps(props: Props, state: State) {
+    // reset form into initial state when closing
+    if (!props.isVisible) {
+      return { ...initialFormState, hasPrefilledState: false };
+    }
+    const editModeOpened = props.isEditing && !state.hasPrefilledState;
+    // If is in edit mode preload passenger data into the form state
+    if (editModeOpened) {
+      return { ...props.prefillData, hasPrefilledState: true };
+    }
+
+    return state;
+  }
 
   handleGenderChange = (gender: any) => {
     this.setState({ gender });
@@ -104,7 +126,8 @@ class PassengerForm extends React.Component<Props, State> {
       id,
       dateOfBirth,
       gender,
-      name: `${name ?? ''} ${lastName ?? ''}`,
+      name,
+      lastName,
       bags,
       passengerCount: 1,
     };
@@ -138,12 +161,14 @@ class PassengerForm extends React.Component<Props, State> {
               label="Given names"
               autoCorrect={false}
               type="text"
+              value={this.state.name}
               formLabelContainerStyle={styles.inputLabel}
             />
             <TextInput
               onChangeText={this.handleLastNameChange}
               label="Last name"
               autoCorrect={false}
+              value={this.state.lastName}
               type="text"
               formLabelContainerStyle={styles.inputLabel}
             />
@@ -152,6 +177,7 @@ class PassengerForm extends React.Component<Props, State> {
               label="Passport or ID number"
               autoCorrect={false}
               type="text"
+              value={this.state.id}
               formLabelContainerStyle={styles.inputLabel}
             />
             <DateInput
--- a/apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
+++ b/apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
@@ -21,12 +21,14 @@ type Props = {|
 type State = {|
   +isFormVisible: boolean,
   +passengerData: Array<PassengerCardType>,
+  +currentEditID: ?string,
 |};
 
 // @TODO: This data should not be mocked in the future. Remove it.
 const defaultPassengers = [
   {
-    name: 'John Doe',
+    name: 'John',
+    lastName: 'Doe',
     gender: 'male',
     nationality: 'Russian',
     dateOfBirth: new Date('1980-06-22'),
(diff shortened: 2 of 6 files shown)


packages/components/src/passengerCard/PassengerCard.js
```diff
@@ -135,9 +135,13 @@ class PassengerCard extends React.Component<Props> {
                 {bags &&
                   bags.map((bag, idx) => (
                     <BagInformation
-                      key={`${bag.dimensions}, ${idx}`}
+                      key={`${
```
JoseDuSV: Cannot we use just index/idx for key?

packages/components/src/passengerCard/PassengerCard.js
```diff
@@ -135,9 +135,13 @@ class PassengerCard extends React.Component<Props> {
...
+                        bag.dimensions ? `${bag.dimensions},` : ''
+                      }, ${idx}`}
                       count={bag.quantity}
-                      type={`${bag.dimensions}, ${bag.weight}`}
+                      type={`${
```
JoseDuSV: It is little bit hard to read. What about at least extracting it to top of the render function next to quite similar `passengerWithTitle`?

aalises: Done!

apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -191,7 +212,11 @@ class PassengerForm extends React.Component<Props, State> {
           </View>
           <View style={styles.spacer} />
           <View style={styles.menuButtonWrap}>
-            <Button block={true} label="Save" onPress={this.handleSavePress} />
+            <Button
+              block={true}
+              label={this.props.isEditing ? 'Edit' : 'Save'}
```
JoseDuSV: Not sure about this. Original `Save` is ok for me in both cases. When I see `Edit` button I would expect to be taken to some other "edit screen" not that it will save my updated values.
But maybe I just get used too much to that old `Save` label. 😄

aalises: What about `Save Edit` instead of `Edit`? 😆

packages/components/src/passengerCard/PassengerCard.js
JoseDuSV: Sorry I didn't noticed that it is actually nested inside `map`. In this case we can probably move this new parser even slightly up out of render function. What do you think?

aalises: Done!

apps/core/components/passengerForm/PassengerForm.js
JoseDuSV: Not sure 🙂, I'm for good old `Save` personally. But I'm also definityle open to other opinions. Lets ask at least one more person 😉

aalises: Let's keep `Save` then, reverted.

apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -48,23 +50,39 @@ const nationalityData = [
...
-    nationality: null,
-    dateOfBirth: null,
-    bags: null,
-    passengerCount: 1,
+    ...initialFormState,
+    hasPrefilled: false,
   };
 
+  static getDerivedStateFromProps(props, state) {
+    // Reset the form if we add a passenger after an edit process
```
JoseDuSV: Looks like reset doesn't always behave correctly. When I try to add second user in row, form is not cleared.

pass-save

aalises: Good catch! will look at it now

aalises: Fixed!

FilipMessa:
When a user clicks on the edit the first name and last name are merged in the first name field and last-name field is empty. 

ezgif com-optimize (1)

aalises:
> When a user clicks on the edit the first name and last name are merged in the first name field and last-name field is empty.
> 
> ezgif com-optimize (1)

Done @FilipMessa !

apps/core/components/passengerForm/PassengerForm.js
JoseDuSV: 👍Problem is gone, thanks!

FilipMessa:
When a user clicks on edit passenger, the baggage option is always changed to no baggage value.
bug-baggage

aalises:
@FilipMessa  Yep, this is because the `BaggageBundles`  are not prefilled :( couldn't manage to do it as we are dealing with Relay Fragments in the components (`BagOptions`) so passing a prop with the `initialSelected` wouldn't do it... any idea?

apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -48,23 +50,38 @@ const nationalityData = [
...
-    gender: 'male',
-    name: null,
-    lastName: null,
-    id: null,
-    nationality: null,
-    dateOfBirth: null,
-    bags: null,
-    passengerCount: 1,
+    ...initialFormState,
+    hasPrefilled: false,
```
FilipMessa:
```suggestion
    hasPrefilledState: false,
```
Will be better for understanding what's going on

apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -48,23 +50,38 @@ const nationalityData = [
...
   };
 
+  static getDerivedStateFromProps(props, state) {
+    if (!props.isVisible) {
```
FilipMessa: Maby add some comment here:
```
// reset form into initial state
```

apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -48,23 +50,38 @@ const nationalityData = [
...
+      return { ...initialFormState, hasPrefilled: false };
+    }
+    return props.isEditing && !state.hasPrefilled
```
FilipMessa: Also here please add some explanation comment
something like
```
 // If is in edit mode load passenger data into the form state

```

TBH: It took me some time to understand what's going there, so I believe that there exist some more elegant solution.

apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
```diff
@@ -79,16 +82,18 @@ class ResultDetailPassenger extends React.Component<Props, State> {
...
     passengerData: defaultPassengers,
+    currentEditID: '',
   };
 
   handleFormSaveRequest = passenger => {
     const existsID: boolean = !!this.state.passengerData.find(
       el => el.id === passenger.id,
     );
+
     const validateForm = () => {
```
FilipMessa: Can we declare the validation function outside? It will be more readable.

apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
```diff
@@ -79,16 +82,18 @@ class ResultDetailPassenger extends React.Component<Props, State> {
...
       if (!(passenger.id && passenger.name)) {
         return 'ID and Name are mandatory';
-      } else if (existsID) {
+      } else if (existsID && this.state.currentEditID !== passenger.id) {
```
FilipMessa:
```suggestion
      }  
      if (existsID && this.state.currentEditID !== passenger.id) {
```

I suppose that the `else` is not necessary.

apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
```diff
@@ -100,16 +105,19 @@ class ResultDetailPassenger extends React.Component<Props, State> {
       });
     } else {
       this.setState(state => ({
-        passengerData: [...state.passengerData, passenger],
+        // If we are editing, replace the passenger, else add it
+        passengerData: !state.currentEditID
```
FilipMessa: What about making some helper function from this to make it more readable?

FilipMessa:
@aalises related to the baggage issue, I understand so we can handle this in another PR at first will need to removed mocked passengers.

Issue #787, opened 2019-05-16
Call the `checkItinerary` GraphQL query with the real passengers' data.


FilipMessa:
The `checkItinerary` is used for check booking validity. The query expects data about a number of badges and a number of passengers.

If the user adds more passengers that in search form we need to check booking validity again.

PR #791, opened 2019-05-20
Refactor: Remove mock passenger data

### LINKED ISSUE #790: Result Detail - remove mocked passengers
Just remove mocked passengers

aalises:
Closes #790 . Also, added validation that at least one passenger is required before going into the `Payment` screen of the booking process.


FilipMessa:
I think this is not the best way how to do this. 
I don't like that we need the passengers' count (`passengerNumber`) in the state.

I think we should move the state probably into the new Context (bookingContext), It will be handy because we will need to also call the 'checkItinerary' query if the count of the passengers has changed (`ResultDetailInner`) to verify if the booking is valid. #787

aalises:
Sure @FilipMessa , so let's do it in a different PR then

PR #796, opened 2019-05-20
Feature: Add booking context / Payment screen passenger number validation

### LINKED ISSUE #794: Booking: Disable continue button if there are no passengers (create bookingContext)
When there are no passengers on booking, disable the `Continue` button that leads to the payment stage. 

Screenshot 2019-05-20 at 14 44 31

To do that, we need a new context (`bookingContext`), that will store the number of passengers so we can then check when it changes (also useful to refetch `checkItinerary`) #787

aalises:
Also,  on booking flow, if no passengers available, do not let the user to the payment stage. Closes  #794

### DIFF
--- a/apps/core/components/bookingContext/BookingContext.js
+++ b/apps/core/components/bookingContext/BookingContext.js
@@ -1,0 +1,67 @@
+// @flow
+
+import * as React from 'react';
+import { withContext, noop } from '@kiwicom/margarita-utils';
+
+import { type BaggageBundleType } from '../passengerForm/baggageBundles/__generated__/BaggageBundle_bagOption.graphql';
+
+type Props = {|
+  +children: React.Node,
+|};
+
+export type PassengerType = {|
+  +name: ?string,
+  +lastName: ?string,
+  +gender: 'female' | 'male' | 'other',
+  +nationality: ?string,
+  +dateOfBirth: ?Date,
+  +id: ?string,
+  +insurance?: ?string,
+  +bags: null | Array<BaggageBundleType>,
+  +passengerCount: number,
+  +visaRequired?: ?boolean,
+|};
+type State = {|
+  +passengers: Array<PassengerType>,
+  +actions: {|
+    +setPassengers: (Array<PassengerType>) => void,
+  |},
+|};
+
+const defaultState = {
+  passengers: [],
+  actions: {
+    setPassengers: noop,
+  },
+};
+
+const { Provider, Consumer } = React.createContext<State>(defaultState);
+
+export default class BookingContextProvider extends React.Component<
+  Props,
+  State,
+> {
+  constructor(props: Props) {
+    super(props);
+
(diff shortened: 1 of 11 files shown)


apps/core/components/bookingContext/BookingContext.js
```diff
@@ -0,0 +1,54 @@
...
+
+    this.state = {
+      ...defaultState,
+      actions: {
+        setPassengers: this.setPassengers,
+      },
+    };
+  }
+
+  setPassengers = (passengers: Array<PassengerCardType>) => {
```
FilipMessa: Maby just the `PassengerType` will be better now? what do you think?

apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
```diff
@@ -44,15 +48,14 @@ class ResultDetailPassenger extends React.Component<Props, State> {
     return '';
   };
 
-  getNewPassengerData = passenger => {
+  getNewpassengers = passenger => {
```
FilipMessa: It should be `getNewPassengers`

apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
```diff
@@ -88,14 +87,14 @@ class ResultDetailPassenger extends React.Component<Props, State> {
   };
 
   render() {
-    const { passengerData, isFormVisible, currentEditID } = this.state;
+    const { isFormVisible, currentEditID } = this.state;
     const prefillData = currentEditID
-      ? passengerData.find(el => el.id === currentEditID)
+      ? this.props.passengers.find(el => el.id === currentEditID)
```
FilipMessa: What about deconstruction? JUST IDEA

aalises: Done!

apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
aalises: Done!

apps/core/components/bookingContext/BookingContext.js
aalises: You mean creating `type PassengerType = Array<PassengerCardType>`?

FilipMessa: No `PaseangerCardType` is too coupled with the `PassengerCard` component and now we want to use it in many places, and I think that type should be more related to Context than the Component because the state is the source of truth.

But it is just a proposal. To have the `passengerType` inside the context.

aalises: Done! now `PassengerType` is in the context and is used by `PassengerCard`

PR #797, opened 2019-05-20
universal-components: The InputText can works like controlled component

FilipMessa:


aalises:
LGTM! 🚀  Why do we need that now? @FilipMessa

FilipMessa:
@aalises
I need to clear contact form data after user sign out himself.

PR #802, opened 2019-05-21
Feat/save contact info

### LINKED ISSUE #775: As a user, I want to save my contact information.
1. Save a user contact data into the firebase under his account
2. If a user contact information is available, prefill contact form with that.
3. User can edit his contact information

Screen Shot 2019-05-15 at 10 36 29

FilipMessa:
close #775

ezgif com-optimize (2)


apps/core/components/userContext/UserContext.js
```diff
@@ -81,14 +108,19 @@ export default class UserContextProvider extends React.Component<Props, State> {
...
 
-  mapUserData = (authUser: Object) => {
-    return {
-      userEmail: authUser?.email,
-      userDisplayName: authUser?.displayName,
-      profileImage: authUser?.photoURL,
-    };
+  getPhoneNumberFromFirebase = async (userId: string) => {
+    const userRef = firebase.database().ref(`users/${userId}/phoneNumber`);
+    const phoneNumbersnapshot = await userRef.once('value');
```
aalises: `const phoneNumberSnapshot` with  capital Snapshot. Also, `userRef` is not really the user reference but the `phoneNumberRef`. We can also get the user reference and then `const phoneNumbersnapshot = await userRef.child('phoneNumber').once('value');` like we do on the `setUserPhoneNumber`, but I am perfectly okay with that 🚀.

apps/core/components/userContext/UserContext.js
```diff
@@ -99,12 +131,28 @@ export default class UserContextProvider extends React.Component<Props, State> {
...
-      await firebase.auth().signOut(); // @TODO refactro to GoogleLoginHelper
+      await firebase.auth().signOut();
     } catch (e) {
       // @TODO Handle sign-in error - show notification or message
     }
   };
 
+  setUserPhoneNumber = (phoneNumber: PhoneNumber) => {
+    const { userId } = this.state;
+    const userRef = userId && firebase.database().ref(`users/${userId}`);
```
aalises: Maybe we could use a helper method here for getting the `phoneNumber` reference and also use it on `getPhoneNumberFromFirebase`, something like:
```
getPhoneNumberReference = (userId: ?string) => {
  return userId && firebase.database().ref(`users/${userId}/phoneNumber`);
}
```

packages/components/src/contactDetailsForm/ContactDetailsForm.js
```diff
@@ -13,15 +13,23 @@ import { defaultTokens } from '@kiwicom/orbit-design-tokens';
 import HeaderWithIcon from '../headerWithIcon/HeaderWithIcon';
 
 type Props = {|
-  onChangeEmail: string => void,
-  onChangePhoneNumber: string => void,
-  onChangeCountryCode: (?string) => void,
+  +disableEmail: ?boolean,
```
aalises: `disableEmail` cannot be `null` as `disabled` on TextInput cannot be null (but it's optional) and also we are generating it with `Boolean(this.props.userId)` which can never be null. So maybe change that type to be `+disableEmail?: boolean` and set a default disabled if the prop is not there, or even `+disableEmail: boolean`

apps/core/components/userContext/UserContext.js
FilipMessa: good idea 👍

packages/components/src/contactDetailsForm/ContactDetailsForm.js
FilipMessa: good point 👍

apps/core/components/userContext/UserContext.js
FilipMessa: done

apps/core/components/userContext/UserContext.js
FilipMessa: done

packages/components/src/contactDetailsForm/ContactDetailsForm.js
FilipMessa: done

apps/core/scenes/resultDetail/resultDetailContent/ResultDetailContent.js
```diff
@@ -59,12 +92,17 @@ class ResultDetailContent extends React.Component<Props, State> {
...
       this.props.itinerary?.price?.currency,
     );
+
     return (
       <>
         <ContentContainer>
           <ItinerarySectorDetails itinerary={this.props.itinerary} />
           <ResultDetailPassenger itinerary={this.props.itinerary} />
           <ContactDetailsForm
+            disableEmail={Boolean(this.props.userId)}
```
JoseDuSV: Looks good. Maybe just one small thing to consider. We can probably allow user to set different contact email for booking, just prefill their email from login as default value. But if it will create some complications we can even keep it as it is right now.

FilipMessa: I was also thinking about that but in the current state of the application, it is nice to have, and it will better do it in separate PR. So I will create an issue for that.

JoseDuSV: Ok, so lets merge this one 🚀

PR #813, opened 2019-05-22
Feature: Call checkItinerary on passenger change (ResultDetails)

### LINKED ISSUE #787: Call the `checkItinerary` GraphQL query with the real passengers' data.

aalises:
Closes #787

### DIFF
--- a/apps/core/scenes/resultDetail/ResultDetail.js
+++ b/apps/core/scenes/resultDetail/ResultDetail.js
@@ -14,7 +14,9 @@ type Props = {|
 
 export default class ResultDetail extends React.Component<Props> {
   renderInner = (data: ResultDetailQueryResponse) => {
-    return <ResultDetailInner data={data} />;
+    return (
+      <ResultDetailInner bookingToken={this.props.bookingToken} data={data} />
+    );
   };
 
   render() {
--- a/apps/core/scenes/resultDetail/ResultDetailInner.js
+++ b/apps/core/scenes/resultDetail/ResultDetailInner.js
@@ -8,12 +8,19 @@ import {
 } from '@kiwicom/margarita-relay';
 import { IllustrationWithInformation } from '@kiwicom/margarita-components';
 
+import {
+  withBookingContext,
+  type BookingContextState,
+  type PassengerType,
+} from '../../contexts/bookingContext/BookingContext.js';
 import type { ResultDetailInner_data as ResultDetailInnerType } from './__generated__/ResultDetailInner_data.graphql';
 import ResultDetailContent from './resultDetailContent/ResultDetailContent';
 
 type Props = {|
   +data: ?ResultDetailInnerType,
   +relay: RelayRefetchProp,
+  +passengers: Array<PassengerType>,
+  +bookingToken: ?string,
 |};
 
 /**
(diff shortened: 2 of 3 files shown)


apps/core/scenes/resultDetail/ResultDetail.js
```diff
@@ -30,7 +30,7 @@ export default class ResultDetail extends React.Component<Props> {
           input: {
             bookingToken,
             bags: 0, //  @TODO - use data from passengers forms
-            passengers: { adults, infants },
+            passengers: { adults: 1, infants: 0 },
```
FilipMessa: This is basically used for the first check after the user clicks on the book button on the result list page, it should not be hardcoded. This data should data what user added to search from.

aalises: True that! will put it back now!

aalises: Done!

FilipMessa: It still not working as is expected, when I set adults on 6 and try to book the flight, then check the Graphql API call log, I realized that it calls with the wrong count of passenger it is always with one. Also now in url of result detail page is missing information about the count of passengers, like was before. 

Screen Shot 2019-05-22 at 16 30 49

aalises: @FilipMessa Forgot to add it back to the router navigation, now it should work.

FilipMessa: I am on it 🚶

apps/core/scenes/resultDetail/ResultDetailInner.js
```diff
@@ -31,12 +38,30 @@ class ResultDetailInner extends React.Component<Props> {
...
-      this.stopItineraryCheck();
-      this.tryStartItineraryCheck();
+    const { passengers, data, relay, bookingToken } = this.props;
+    const numPassengers = passengers.length;
+    const passengersChanged = prevProps.passengers.length !== numPassengers;
+    const checkedChanged =
+      data?.checkItinerary?.isChecked !==
+      prevProps.data?.checkItinerary?.isChecked;
+
+    if (checkedChanged || passengersChanged) {
```
FilipMessa: We need to find better architecture because let's say a user search flight for three passengers at the start. When a user is on the result-detail page first checkItinerary run for 3 passengers. But when a user adds a first passenger in to form, we call the checkItinerary with one, if itinerary will be not checked successfully we will call `this.tryStartItineraryCheck()` until is checked, but here is a problem because of the `tryStartItineraryCheck` call `relay.refetch()` with params from search so it will be for 3 passengers now.
This probably won't be a big problem but what if the user will be searched itinerary just for one passenger and decided added 4 passengers? The tryStartItineraryCheck will be called with one but we need to check if a flight is valid for 4 passengers. 
And the price should change also.

FilipMessa: We should somehow simplify this flow. 🤔

aalises: Yes. This `tryStartItineraryCheck` is weird ahahaha, we should maybe revisit it in another PR? or this one, like: `Fix passengers itinerary check`

FilipMessa: Currently, this PR doesn't solve the problem with check itinerary. These two checks with different params just lead to inconsistency. 

I am trying to figure out a simple solution on how to do deal with that.

FilipMessa:
Closed because of new solution which will be delivered in another PR

PR #818, opened 2019-05-22
Refactor: remove default state settings from props

FilipMessa:
I realized that we set the default state for the SearchContext from props, but I can't find any use case for that, and also props sending to SearchContext are different that flowTypes expected.

**PLEASE double check if I didn't miss something.**

### DIFF
--- a/apps/core/contexts/searchContext/SearchContext.js
+++ b/apps/core/contexts/searchContext/SearchContext.js
@@ -13,17 +13,6 @@ import qs from 'qs';
 
 type Props = {|
   +children: React.Node,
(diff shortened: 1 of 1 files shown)


aalises:
Yes, they were added here (https://github.com/kiwicom/margarita/pull/658/files) because before the persistence of the search parameters was done on reload by getting them from the router query params and passing them as props to the context to repopulate it.

However, here (https://github.com/kiwicom/margarita/pull/695) was changed to set it from an action in the context which is called on `ComponentDidMount` on the components that use it (`Results` and `Search`). So the props are indeed useless now :)

PR #820, opened 2019-05-23
Add Google Analytics to landing page

jvaclavik:


apps/landingPage/pages/_document.js
```diff
@@ -88,6 +89,11 @@ export default class CustomDocument extends Document {
         <body>
           <Main />
           <NextScript />
+          <script
+            async
+            src="https://www.googletagmanager.com/gtag/js?id=UA-29345084-41"
+          />
+          <script>ReactGA.initialize('UA-29345084-41');</script>
```
RobinCsl: This will also work in DEV environment I believe ;-)

jvaclavik:
Actually I'm not sure how it should be proper implemented and tested. I can't test it on localhost. Could you help @RobinCsl please?

KarinaDavtyan:
[Slack conversation](https://skypicker.slack.com/archives/C1XN8BFMZ/p1558453763263800) regarding what should we use in #plz-analytics for reference

jvaclavik:
It should be fine now, I figured out how to test it :)

apps/landingPage/containers/LandingPage.js
```diff
@@ -28,6 +30,8 @@ export default class LandingPage extends React.Component<{}, State> {
   };
 
   componentDidMount() {
+    ReactGA.initialize(GOOGLE_ANALYTICS_TRACKING_ID);
```
KarinaDavtyan: What about adding condition here to avoid tracking localhost? `document.location.hostname != 'localhost'` ?

apps/landingPage/config/GA.js
```diff
@@ -0,0 +1,76 @@
+// @flow
+
+import ReactGA from 'react-ga';
+
+export const GOOGLE_ANALYTICS_TRACKING_ID = 'UA-29345084-41';
+
+export const EventLandingPageReadMore = () => {
+  ReactGA.event({
```
KarinaDavtyan: Not sure if we need to track it.

apps/landingPage/config/GA.js
```diff
@@ -0,0 +1,76 @@
...
+    category: 'Github',
+    action: 'star',
+    label: 'Give us a star',
+  });
+};
+
+export const EventKiwiJobs = () => {
+  ReactGA.event({
+    category: 'kiwi',
```
KarinaDavtyan: Maybe `kiwi.com`?

apps/landingPage/config/GA.js
jvaclavik: I think it's important. We want to know what is more attractive for users (github or read more). We don't have any other way how to do it

apps/landingPage/containers/LandingPage.js
jvaclavik: OK, done

KarinaDavtyan: Thanks

apps/landingPage/config/GA.js
jvaclavik: ok

apps/landingPage/config/GA.js
KarinaDavtyan: Alright :)

PR #834, opened 2019-05-27
Fix/persist params update search

### LINKED ISSUE #778: BUG: After refreshing the result list page, lost newly selected location.
**To Reproduce**
1. go to the result list page
2. change a location 
3. refresh the page

May-15-2019 12-38-17

KarinaDavtyan:
Closes #778

### DIFF
--- a/apps/core/components/searchForm/SearchForm.js
+++ b/apps/core/components/searchForm/SearchForm.js
@@ -66,7 +66,7 @@ class SearchForm extends React.Component<Props> {
     return '';
   };
 
-  handleSubmitPress = () => {
+  getSearchParams = () => {
     const {
       travelFrom,
       travelTo,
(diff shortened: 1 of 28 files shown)


apps/core/components/searchForm/SearchForm.js
```diff
@@ -129,14 +130,59 @@ class SearchForm extends React.Component<Props> {
...
+      if (onSubmit != null) {
+        onSubmit({
+          dateFrom,
+          dateTo,
+          returnDateFrom,
+          returnDateTo,
+          tripType,
+          travelFrom,
+          travelTo,
+          sort: 'QUALITY',
```
FilipMessa: Why you don't use the `sortBy`  from the SearchContext?

apps/web/pages/results.js
```diff
@@ -14,10 +15,16 @@ type Props = {
 };
 
 const results = ({ router }: Props) => {
+  function onSubmit(query) {
```
FilipMessa: This naming was a little bit confusing for me, for the first time, now I understand.
But I am thinking if here want to be better call that something like `changeUrl` or something like and passes it as an on submit into the: 
```javascript 
<SeachForm onSubmit={this.props.changeUrl} />
``` 

For my, it will be more readable. It is just Idea what do you think.

FilipMessa:
Just last thing when I removed the location for the Destination placekicker, it set up to Anywhere, but after the refresh, the last chosen Location set back.

But it can be handled in another PR.

apps/core/components/searchForm/SearchForm.js
KarinaDavtyan: To be honest, I didnt put much thought here. I just copy pasted required logic from handle SubmitPress.
Thats weird 🧐, will replace in both places. But also will try to refactor to avoid copy pasting

apps/web/pages/results.js
KarinaDavtyan: Totally agree, naming is a bit confusing here

KarinaDavtyan:
> Great job, 👍
> I like that, I just wrote a few notes, and maybe will be great just try to find a way how to avoid this, long pass on Submit through props.
> Maby we could handle the `onSubimt (changeURL)` in the SearchContext.
> Because you can access to `router` inside the `/margarita/apps/web/pages/_app.js`.
> 
> But your solution is OK this is just an Idea and also I don't know if it will be work fine.

I ll try, it should be yet improved for sure.

FilipMessa:
any update?

KarinaDavtyan:
> any update?

@FilipMessa Sorry, will fix it now. Was on holidays

FilipMessa:
> @FilipMessa Sorry, will fix it now. Was on holidays

@KarinaDavtyan Hello Karina, sure no problem, thanks

apps/core/components/searchForm/SearchForm.js
KarinaDavtyan: Changed

apps/web/pages/results.js
KarinaDavtyan: done

apps/core/scenes/results/Results.js
```diff
@@ -56,6 +56,9 @@ type Props = {|
   +nightsInDestinationTo: string,
   +isNightsInDestinationSelected: boolean,
   +setBookingToken: string => void,
+  +routerQuery: SearchParameters,
+  +setStateFromQueryParams: SearchParameters => void,
```
FilipMessa: The `routerQuery` and `setStateFromQueryParams` do not use anywhere, I guess.

apps/core/scenes/search/Search.js
```diff
@@ -22,12 +22,15 @@ export type SearchParameters = {|
   +tripType: string,
   +travelFrom: ?$ReadOnlyArray<Location>,
   +travelTo: ?$ReadOnlyArray<Location>,
-  +sort: string,
+  +sortBy: string,
```
FilipMessa: Pls change this name also in the `parseURLqueryToState` (PARSER_CONFIG).
`/margarita/apps/core/contexts/searchContext/helpers.js`

apps/core/scenes/results/Results.js
KarinaDavtyan: Hmm its not, probably forgot to remove as I dont remember already reason behind putting there in a first place.
Thanks! Removing

apps/core/scenes/search/Search.js
KarinaDavtyan: Changed

FilipMessa:
I found Issue on mobile when I changed the flight type I was redirected immediately on the result page. 
Jun-17-2019 11-31-37

KarinaDavtyan:
> I found Issue on mobile when I changed the flight type I was redirected immediately on the result page.
> Jun-17-2019 11-31-37

Hmm interesting, will look into that. Thanks for spotting it!

KarinaDavtyan:
> I found Issue on mobile when I changed the flight type I was redirected immediately on the result page.

@FilipMessa should be behaving properly now. But I ve discovered other issue, when it goes to /results with `oneWay` param, it still shows as it was `return`.
Would it be in scope of this PR to fix it? (the same behavior is on master)

ezgif com-video-to-gif

FilipMessa:
@KarinaDavtyan Thanks I will check, and related bug what you found it should be in another PR.

FilipMessa:
@KarinaDavtyan LGTM now, BTW will you take care about this bug or?

KarinaDavtyan:
> @KarinaDavtyan LGTM now, BTW will you take care about this bug or?

@FilipMessa Yes I will

PR #835, opened 2019-05-27
Possibility of picking history dates in RangeDatePicker

### LINKED ISSUE #828: Pick date from history in new Range Date Picker
It would be possible to reuse it for choosing birth date

jvaclavik:
- RangeDatePicker can be rendered from desired date (not only from today) by `renderFirstMonthFrom` props
- it's possible to use prop `isChoosingPastDatesEnabled` to enable selecting dates in history

Closes #828

### DIFF
--- a/packages/universal-components/src/RangeDatePicker/RangeDatePicker.js
+++ b/packages/universal-components/src/RangeDatePicker/RangeDatePicker.js
@@ -13,8 +13,6 @@ import NightsInDestination from './NightsInDestination';
 import ControlContainer from './components/ControlContainer';
 
 export default class RangeDatePicker extends React.Component<Props> {
-  // @TODO load price for days
-
   static defaultProps = {
     isNightsInDestinationVisible: false,
     isNightsInDestinationSelected: false,
@@ -52,6 +50,7 @@ export default class RangeDatePicker extends React.Component<Props> {
 
   render() {
     const {
+      isChoosingPastDatesEnabled,
       isVisible,
       label,
       buttonLabels,
@@ -65,6 +64,7 @@ export default class RangeDatePicker extends React.Component<Props> {
       nightsInDestinationLabel,
       isNightsInDestinationSelected,
       isControlContainerVisible,
+      renderFirstMonthFrom,
     } = this.props;
 
     return (
@@ -76,11 +76,13 @@ export default class RangeDatePicker extends React.Component<Props> {
         <View style={styles.content}>
           {!isNightsInDestinationSelected || !isNightsInDestinationVisible ? (
             <RangeDatePickerContent
+              isChoosingPastDatesEnabled={isChoosingPastDatesEnabled ?? false}
               selectedDates={this.props.dates}
               onDayPress={this.handleChangeDate}
               numberOfRenderedMonths={numberOfRenderedMonths}
-              weekStartsOn={weekStartsOn}
+              weekStartsOn={weekStartsOn ?? 1}
               isRangePicker={isRangePicker ?? true}
+              renderFirstMonthFrom={renderFirstMonthFrom ?? new Date()}
             />
           ) : (
             <NightsInDestination
--- a/packages/universal-components/src/RangeDatePicker/RangeDatePickerContent.js
+++ b/packages/universal-components/src/RangeDatePicker/RangeDatePickerContent.js
@@ -19,6 +19,8 @@ type Props = {|
   +numberOfRenderedMonths: number,
   +weekStartsOn: WeekStartsType,
   +isRangePicker: boolean,
+  +renderFirstMonthFrom: Date,
+  +isChoosingPastDatesEnabled: boolean,
 |};
 
 type State = {|
@@ -32,7 +34,12 @@ export default class RangeDatePickerContent extends React.Component<
   constructor(props: Props) {
     super(props);
     this.state = {
-      nextMonths: getMonths(props.numberOfRenderedMonths, props.weekStartsOn),
+      nextMonths: getMonths({
+        numberOfRenderedMonths: props.numberOfRenderedMonths,
+        weekStartsOn: props.weekStartsOn,
+        whichMonthsToRender: 'next',
+        renderFirstMonthFrom: props.renderFirstMonthFrom,
+      }),
     };
   }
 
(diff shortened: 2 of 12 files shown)


packages/universal-components/src/RangeDatePicker/RangeDatePicker.js
```diff
@@ -76,11 +78,13 @@ export default class RangeDatePicker extends React.Component<Props> {
         <View style={styles.content}>
           {!isNightsInDestinationSelected || !isNightsInDestinationVisible ? (
             <RangeDatePickerContent
+              isChoosingPastDatesEnabled={isChoosingPastDatesEnabled ?? false}
```
FilipMessa: Why not use default props for that?

packages/universal-components/src/RangeDatePicker/RangeDatePickerContent.js
```diff
@@ -32,7 +34,12 @@ export default class RangeDatePickerContent extends React.Component<
   constructor(props: Props) {
     super(props);
     this.state = {
-      nextMonths: getMonths(props.numberOfRenderedMonths, props.weekStartsOn),
+      nextMonths: getMonths(
```
FilipMessa: It takes too many parameters, what about refactor to object input type? But nothing big deal just an idea.

jvaclavik:
> And two things?
> What is the motivation for this?

Why it should render dates only from current date without possibility of select date in history? It's nothing very important, but these two props make RangeDatePicker much more universal. 
Originally we found these issue when we wanted to use it for choosing birth date. It's not actual anymore, but I think it still could be useful and it was quite easy fix. 

> And I realize that it doesn't work properly in the storybook.
> When I set the ChoosingPastDatesEnabled to true, nothing happens, I can't choose before date, at first I must take a future date and then it allows me.
> 
> 1. Open the RangeDatePicker in the storybook
> 2. checked the ChoosingPastDatesEnabled
> 3. try to pick up past date

I also noticed that but I was thinking that I would fix this only because of Storybook. It doesn't seem to me very important, because I don't see there any use case for changing this props dynamically. I can fix it if you think it should be like that.

packages/universal-components/src/RangeDatePicker/RangeDatePicker.js
jvaclavik: Because `defaultProps` will not convince flow that it's fine that way 😄. It doesn't work without condition. How should I do it properly without repeating default value on more than one place?

FilipMessa:
> I also noticed that but I was thinking that I would fix this only because of Storybook. It doesn't seem to me very important, because I don't see there any use case for changing this props dynamically. I can fix it if you think it should be like that.

I think that should work because of this is as the react work, if I change props I expect that component will react on that.

JoseDuSV:
> > I also noticed that but I was thinking that I would fix this only because of Storybook. It doesn't seem to me very important, because I don't see there any use case for changing this props dynamically. I can fix it if you think it should be like that.
> 
> I think that should work because of this is as the react work, if I change props I expect that component will react on that.

I'm also for fix. I understand that it is not important for real use case. But on the other hand Storybook is usually your first contact with components when you're exploring the library. And I also thought that something is broken when I played with it for a first time.

packages/universal-components/src/RangeDatePicker/RangeDatePickerTypes.js
```diff
@@ -9,6 +9,8 @@ export type ConfirmType = {|
 |};
 
 export type Props = {|
+  +calendarStartDate?: Date,
+  +isChoosingPastDatesEnabled?: boolean,
```
JoseDuSV: I'm slightly confused by this new functionality. Shouldn't this be handled just by setting `calendarStartDate`?
When I set start date to for example 10. 4. , I would expect that picker will be selectable from this date. Instead of it, the month is rendered, but still not selectable. 

Screenshot 2019-05-30 at 14 59 02

And when I choose `isChoosingPastDatesEnabled` option, whole fourth month is selectable.

Same with the higher month. When I choose 10. 6. as start. Month is again selectable from start.

Screenshot 2019-05-30 at 14 59 26

I'm not sure how important is this feature right now. We can update it to make exact `calendarStartDate` limit possible selection range. Then `isChoosingPastDatesEnabled` will not be necessary because I will be able to set start date exactly to day I want. As addition to it we can also have something like `calendarEndDate` to specify end limit for selection range (can replace `numberOfRenderedMonths` functionality).

O at least rename`calendarStartDate` to something like `calendarStartMonth` (or some object which will have month and year props) to be clear that actual day doesn't really matter.

jvaclavik: Good point. I didn't realise that this name could be confused with functionality of `isChoosingPastDatesEnabled`. It makes sense to rename it. 

Your other proposed solution is a bit more complicated and if we don't need it now (we don't), let's skip it for now.

jvaclavik:
> > > I also noticed that but I was thinking that I would fix this only because of Storybook. It doesn't seem to me very important, because I don't see there any use case for changing this props dynamically. I can fix it if you think it should be like that.
> > 
> > 
> > I think that should work because of this is as the react work, if I change props I expect that component will react on that.
> 
> I'm also for fix. I understand that it is not important for real use case. But on the other hand Storybook is usually your first contact with components when you're exploring the library. And I also thought that something is broken when I played with it for a first time.

OK, your notes makes sense. I will fix it

packages/universal-components/src/RangeDatePicker/RangeDatePickerTypes.js
JoseDuSV: Ok, if this functionality is not crucial right now definitely just some renaming should be enough.
Just to be clear, I'm ok with the `isChoosingPastDatesEnabled`. But it should be somehow clear  that `calendarStartDate` doesn't really define exact starting date/day and only tells which month will be rendered as first one. Thanks

packages/universal-components/src/RangeDatePicker/RangeDatePickerContent.js
jvaclavik: Done

packages/universal-components/src/RangeDatePicker/RangeDatePickerTypes.js
jvaclavik: Done

jvaclavik:
> > > > I also noticed that but I was thinking that I would fix this only because of Storybook. It doesn't seem to me very important, because I don't see there any use case for changing this props dynamically. I can fix it if you think it should be like that.
> > > 
> > > 
> > > I think that should work because of this is as the react work, if I change props I expect that component will react on that.
> > 
> > 
> > I'm also for fix. I understand that it is not important for real use case. But on the other hand Storybook is usually your first contact with components when you're exploring the library. And I also thought that something is broken when I played with it for a first time.
> 
> OK, your notes make sense. I will fix it

Done

FilipMessa:
I am on it 🚀

FilipMessa:
I realize that when I unchecked the IsRangePicker value, the behavior is to as I expect. Will be hard to fix it? 


bug

packages/universal-components/src/RangeDatePicker/RangeDatePicker.js
FilipMessa: BTW: 
If you will be set the `isChoosingPastDatesEnabled` flow type to required, and then set the default props `isChoosingPastDatesEnabled` to false. It will be work, and it will be optional 😄

PR #846, opened 2019-05-28
Refactor the SearchContext

FilipMessa:
**Summary**
- Hydrate the SearchContext from the URL at context creation
- Removed the `setStateFromQueryParams` from components `componentDidMount`, because it shouldn't be necessary.
- Moved the BookingToken from BookingContext to SearchContext, for code simplicity
(need to rethink if the `BookingContext` is necessary)

__________________________________________________________________________________
**Plan for next PRs**
- Move the SearchContexts parser methods to helper file, and write tests.
- Render empty PassengerCards based on count of passengers from SearchContex, similar behavior like on kiwi.com

### DIFF
--- a/apps/core/contexts/bookingContext/BookingContext.js
+++ b/apps/core/contexts/bookingContext/BookingContext.js
@@ -23,20 +23,16 @@ export type PassengerType = {|
 |};
 
 type State = {|
-  bookingToken: ?string,
   passengers: Array<PassengerType>,
   actions: {|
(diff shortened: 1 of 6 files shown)


FilipMessa:
@aalises 

> LGTM! The `setStateFromQueryParams` is not necessary/not used anymore, right? Could we remove it or will be needed somewhere else?

Good point, I guess we could remove it. I will do that.
Thanks

FilipMessa:
@aalises done

Issue #849, opened 2019-05-28
Update Margarita Expo page


JoseDuSV:
If we want to start sharing deployed Margarita version, we should update project informations on Expo app page https://expo.io/@kiwicom-margarita-public/tequila-client .

Some short description with link to landing page should be enough.

Login informations can be found inside 1Pass.

PR #851, opened 2019-05-29
Docs: Add description to Expo app

aalises:
#849

### DIFF
--- a/apps/mobile/app.json
+++ b/apps/mobile/app.json
@@ -1,12 +1,15 @@
 {
   "expo": {
-    "name": "tequila-client",
+    "name": "margarita-mobile",
     "entryPoint": "./apps/mobile/index.js",
-    "description": "This project is really great.",
-    "slug": "tequila-client",
+    "description": "Margarita is an open-source, multi-platform travel app for all major platforms — one code for web, iOS and Android. More information on https://margarita.kiwi.com/",
+    "slug": "margarita-mobile",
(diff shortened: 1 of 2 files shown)


apps/mobile/app.json
```diff
@@ -2,11 +2,14 @@
   "expo": {
     "name": "tequila-client",
     "entryPoint": "./apps/mobile/index.js",
-    "description": "This project is really great.",
+    "description": "Margarita is an Open-Source, Multi-platform travel app for all major platforms — one code for web, iOS and Android. More information on https://margarita.kiwi.com/",
```
jvaclavik: What about this change?

"Margarita is an open source, multi-platform travel app..."

aalises: Sure!

apps/mobile/app.json
```diff
@@ -2,11 +2,14 @@
   "expo": {
     "name": "tequila-client",
```
jvaclavik: Btw I'm also thinking about this name. Why don't we use something like `margarita`?

aalises: Say no more, maybe something like  `margarita-mobile`?

jvaclavik: Yes, it's fine. Pls update also url for Expo link in README.md.

aalises: Done!

PR #854, opened 2019-05-29
Replace range date picker

KarinaDavtyan:
Add TextInputs for inputing day and year, Picker for the month.
Screenshot 2019-06-05 at 13 30 37


apps/core/components/passengerForm/DateInput.js
```diff
@@ -0,0 +1,131 @@
...
+  getMonthOptions = () => {
+    return [...Array(12).keys()].map<Month>(
+      (currentMonth: number): Month => ({
+        value: String(currentMonth),
+        label: format(setMonth(new Date(), currentMonth), 'MMMM'),
+      }),
+    );
+  };
+
+  handleMonthChange = (month: ?string) => {
```
FilipMessa: Is there a possibility to call handleMonthChange without a mount?

KarinaDavtyan: I am not sure how to do that

FilipMessa: Sorry, probably it is misunderstood, I just don't understand why is the `mount` optional. ```(month: ?string)```

KarinaDavtyan: Oh sorry.
It was because of `+onValueChange: (value: ?string) => void,` from `PickerTypes.js`
But honestly I dont understand why it was there. And as there might be reasoning behind it, I kept it like that

FilipMessa: Ah OK, thanks

apps/core/components/passengerForm/DateInput.js
```diff
@@ -0,0 +1,131 @@
...
+  +dayError: string,
+  +yearError: string,
+|};
+
+type Month = {|
+  value: string,
+  label: string,
+|};
+
+class DateInput extends React.Component<Props, State> {
```
jvaclavik: I'm thinking about to have this component as a part of universal components. What do you think @KarinaDavtyan @FilipMessa ?

KarinaDavtyan: Hmm I dont think its worth doing as multiplatform components development coming

FilipMessa: Hm, I think it is good Idea have it in the `universal-components` or inside `/margarita/packages/components/src/`

JoseDuSV: True that it's somewhere between. But because it is "just" composite of multiple inputs, I'm slightly more for keeping it inside `margarita-components`.

Can we just maybe replace current `DateInput` inside `margarita-components` with this one? It existed only for this use case so the old one is no longer necessary.

apps/core/components/passengerForm/DateInput.js
```diff
@@ -0,0 +1,131 @@
...
+import { View } from 'react-native';
+import {
+  StyleSheet,
+  Picker,
+  Text,
+  TextInput,
+} from '@kiwicom/universal-components';
+
+type Props = {|
+  +onBirthDateSubmit: Date => void,
```
JoseDuSV: What about using something more universal like `onDateChange`. I think that there is some possibility that we will use this component for some different purpose than just birth date.

apps/core/components/passengerForm/DateInput.js
KarinaDavtyan: Alright, will replace DateInput with this

apps/core/components/passengerForm/DateInput.js
KarinaDavtyan: True. Thanks!

apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -56,14 +64,33 @@ const initialFormState = {
   lastName: null,
   id: null,
   nationality: null,
+  date: {
+    day: '',
+    month: '',
+    year: '',
+  },
   dateOfBirth: null,
```
FilipMessa: Do we need the dateOfBirth state? On first look, it looks like duplicity with the date state.

apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -111,26 +158,105 @@ class PassengerForm extends React.Component<Props, State> {
     this.setState({ id });
   };
 
+  validateYear = (year: number): boolean => {
+    return year > getYear(new Date()) - 110 && year < getYear(new Date()) - 18;
```
FilipMessa: Passenger can be also a child or infants. 
By the way, can you split this PR on two?
1. component
2. validation

Thanks

KarinaDavtyan: Oops, forgot about this

Will try!

FilipMessa: I don't want to be a pain in the ass so If you want to keep it in one I will be OK with that. I was just confused with the setState at the end of the validation method. So I think that there must be a better way how to handle it.

apps/core/components/passengerForm/PassengerForm.js
KarinaDavtyan: Not really anymore, thanks!

After removing it, had to change `State` a little bit, as couldnt find a way to reuse `PassengerType` without date of birth property, so I just copy pasted to the State and removed it

Do you know any way to avoid it? Was trying to use `$<Shape>` but didnt seem to work
Thanks!

apps/core/components/passengerForm/PassengerForm.js
KarinaDavtyan: removed setState from validation as its not needed anymore

I could totally divide it, this is what I wanted initially. But the approach of doing it changed several times and at some point I couldnt really divide it. Although now it would be fine.
Sorry for the mess 😞

apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -111,26 +164,91 @@ class PassengerForm extends React.Component<Props, State> {
...
+      this.resetError('day');
+    }
+    if (!year) {
+      this.handleInvalidDate('year', 'Please fill in the field');
+      return false;
+    }
+    this.resetError('year');
+
+    const isYearValid = this.validateYear(year);
+    if (!isYearValid) {
```
jvaclavik: There should be different error message: 

Screenshot 2019-06-06 at 11 24 42

packages/components/src/dateInput/DateInput.js
```diff
@@ -1,90 +1,124 @@
...
+        <Text style={styles.text}>Date of birth</Text>
+        <View style={styles.dateFormWrapper}>
+          <View style={styles.dayPickerWrapper}>
+            <TextInput
+              onChangeText={this.handleDayChange}
+              placeholder="DD"
+              autoCorrect={false}
+              value={this.props.date.day}
+              error={this.props.errors.day}
+              confirmLabel="OK"
```
jvaclavik: I think this prop should not be here, because `TextInput` from UC doesn't have it.

packages/components/src/dateInput/DateInput.js
```diff
@@ -1,90 +1,124 @@
...
+            />
+          </View>
+          <View style={styles.yearPickerWrapper}>
+            <TextInput
+              onChangeText={this.handleYearChange}
+              autoCorrect={false}
+              value={this.props.date.year}
+              error={this.props.errors.year}
+              placeholder="YYYY"
+              confirmLabel="OK"
```
jvaclavik: Same here

packages/components/src/dateInput/DateInput.js
```diff
@@ -1,90 +1,124 @@
...
+    display: 'flex',
+    justifyContent: 'space-between',
+    flexDirection: 'row',
+  },
+  text: {
+    marginStart: parseInt(defaultTokens.spaceSmall, 10),
+    marginBottom: parseInt(defaultTokens.spaceXXSmall, 10),
+  },
+  dayPickerWrapper: {
+    flex: 0.2,
```
jvaclavik: What those `0.x` values mean?

packages/components/src/dateInput/DateInput.js
KarinaDavtyan: It was complaining without it, so I added it. You are asking about `confirmLabel` , no?

packages/components/src/dateInput/DateInput.js
KarinaDavtyan: This is for proportionally dividing columns

apps/core/components/passengerForm/PassengerForm.js
KarinaDavtyan: Yes, I should change validation here and it shouldnt be complaining at all
> Passenger can be also a child or infants.

packages/components/src/dateInput/DateInput.js
jvaclavik: Why not to use it without `0.`. It should work with the same ratio principle
Like 
```
dayPickerWrapper:
  flex: 2

monthPickerWrapper: 
  flex: 4

yearPickerWrapper: 
  flex: 3
```
means ratio `2:4:3` which is the same like `0.2:0.4:0.3`

KarinaDavtyan: Screenshot 2019-06-06 at 12 54 32
It starts to look like this without 0's

PR #855, opened 2019-05-29
[WIP] Chore: update flow from v 0.94.0 to 0.98.0

FilipMessa:


FilipMessa:
It needs to investigate why the deploy failed.

FilipMessa:
> So now Ramda `head` does not return `undefined` when inputting an empty array?

Good point, I just updated the flow and flow types and didn't think a lot about that. Thanks, Must double check

apps/graphql/src/apps/itinerary/helpers/Itineraries.js
```diff
@@ -51,15 +51,15 @@ export const getItineraryType = (routes: ?Array<Array<string>>) => {
...
+  return head(head(sectors ?? []).segments ?? []).departure;
 };
 
 export const getItineraryArrival = (
   tripType: ?TripTypes,
   sectors: ?Array<Sector>,
 ): ?RouteStop => {
   return tripType === TRIP_TYPES.RETURN
-    ? last(head(sectors ?? [])?.segments ?? [])?.arrival
+    ? last(head(sectors ?? []).segments ?? [])?.arrival
```
JoseDuSV: Ad Albert mentioned. This part looks weird. It should return undefined according to docs https://ramdajs.com/docs/#head , so there is probably some mistake in new `ramda` type definition.

FilipMessa: Yep, it is the wrong definition in new the ramda type

FilipMessa: The issue raised https://github.com/flow-typed/flow-typed/issues/3355

JoseDuSV: 👍 Thanks. Since we are using our local copy we can probably fix it there directly if we need/want this PR merged.

FilipMessa: OK I will handle that 😸

PR #858, opened 2019-05-30
Refactor: Move the SearchContext's types to separate file.

FilipMessa:


apps/core/contexts/searchContext/SearchContextTypes.js
```diff
@@ -0,0 +1,82 @@
+// @flow
+
+import { type TripType } from '@kiwicom/margarita-config';
+
+export type Passengers = {|
+  adults: number,
```
jvaclavik: I think it should be OK with `+`, what do you think?

FilipMessa: 🚀

packages/config/constants/TripTypes.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-export type TripTypes = 'Return' | 'OneWay' | 'MultiCity';
+export type TripType = 'Return' | 'OneWay' | 'MultiCity';
 export const TRIP_TYPES = {
```
JoseDuSV: When you are on it. We can probably rename this to `TRIP_TYPE` to have it unified. But it's just an idea and definitely not necessary for this PR 😉 .

PR #866, opened 2019-05-31
Feature update highlighting

jvaclavik:
ezgif com-optimize (7)


apps/landingPage/containers/Features.js
```diff
@@ -22,8 +22,11 @@ type State = {|
   +featuresHeight: number,
   +featuresTopPosition: number,
   +featuresRef: ?Element,
+  +scrolledOnFeature: ?number,
 |};
 
+const buffer = 150;
```
FilipMessa: For constant like this, I would be expected UPPERCASE but I don't know how it is done on other places.

jvaclavik: It's without uppercase, so let's keep it for now. We can refactor it later.

PR #872, opened 2019-06-05
Update highlighing on landing page to avoid confusing users

jvaclavik:
According to heatmap, MANY users clicked on highlighted places because those places had the same color as buttons. So now it should be more clear and design is simpler.

Video how the whole landing page looks like can be found here: https://mega.nz/#!gp1mHQLJ!ImLqjtedIppQqVacYk6HTTtGliUZrk6fwMcbKj1AjS8


apps/landingPage/containers/SupportUs.js
```diff
@@ -0,0 +1,68 @@
...
+  margin-top: ${defaultTokens.spaceXLarge};
+`;
+
+const StarContainer = styled.div`
+  padding-bottom: ${defaultTokens.spaceXXXLarge};
+  @media (min-width: ${defaultTokens.widthBreakpointTablet}px) {
+    padding-left: 40px;
+  }
+`;
+const JoinUsContainer = styled.div``;
```
FilipMessa: Do we need this?

jvaclavik: It's more clear in JSX to have this element semantically

PR #873, opened 2019-06-06
Add date validation

### LINKED ISSUE #830: Replace old Date Picker
Range Date Picker should be replaced ⬇️
Screenshot 2019-05-27 at 11 46 55
Month is not converted to the name and it is difficult to navigate through years.

But  first #828 #829 should be solved in order to prepare new Range Date Picker to be appropriate replacement

KarinaDavtyan:
Closes #830

### DIFF
--- a/apps/core/components/passengerForm/PassengerForm.js
+++ b/apps/core/components/passengerForm/PassengerForm.js
@@ -18,6 +18,15 @@ import {
   DateInput,
 } from '@kiwicom/margarita-components';
 import { createFragmentContainer, graphql } from '@kiwicom/margarita-relay';
+import {
+  getYear,
+  setYear,
+  setMonth,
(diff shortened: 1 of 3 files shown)


apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -129,6 +155,76 @@ class PassengerForm extends React.Component<Props, State> {
     });
   };
 
+  validateYear = (year: number): boolean => {
+    return year > getYear(new Date()) - 110 && year <= getYear(new Date());
```
jvaclavik: What about use some const for 110? It could be in this file, but this hardcoded value is hard to read.

KarinaDavtyan: Fixed, called it maxAge

PR #875, opened 2019-06-06
Feat: Add accessibility features to Text component

Kubitini:


packages/universal-components/src/Text/TextTypes.js
```diff
@@ -27,4 +27,9 @@ export type TextType = {|
   +weight?: 'normal' | 'bold',
   +expo?: boolean,
   +theme: Theme,
+  accessible?: boolean,
+  accessibilityLabel?: string,
+  accessibilityLiveRegion?: 'none' | 'polite' | 'assertive',
+  accessibilityRole?: 'header',
```
RobinCsl: There are more possibilities for this. Could you please add them too?
cf. https://github.com/necolas/react-native-web/blob/master/docs/guides/accessibility.md#accessibilityrole

cf. https://facebook.github.io/react-native/docs/accessibility

Kubitini: I've added more options for these feature. Please check it out and let me know whether it can be this way. Also I'm not able to test it on iOS device, could you please check it?

PR #887, opened 2019-06-14
Refactor/merge contexts

FilipMessa:
- Merge BookingContext into the SearchContext
- Prerender PassengerCards


FilipMessa:
PLEASE don't merge if it hasn't at least two approvals

apps/core/contexts/searchContext/helpers.js
```diff
@@ -26,7 +27,7 @@ const PARSER_CONFIG = {
   isNightsInDestinationSelected: 'boolean',
 };
 
-export function parseURLqueryToState(query: Object) {
+export function parseURLqueryToState(query: any) {
```
JoseDuSV: Is that change from `Object` to `any` necessary? Not that `Object` is that much better. But if we cannot use some more exact type it's probably ok.

FilipMessa: Good point, I try to handle the flow and forgotten to change it back.

apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -289,7 +290,7 @@ class PassengerForm extends React.Component<Props, State> {
             />
             <TextInput
               onChangeText={this.handleNameChange}
-              label="Given names"
+              label="First names"
```
jvaclavik: It seems to me that it should be "first name", without "s", what do you think?

jvaclavik: or maybe first name(s)

FilipMessa: 👍

apps/core/contexts/searchContext/helpers.js
FilipMessa: done

apps/core/components/passengerForm/PassengerForm.js
FilipMessa: done

apps/core/contexts/searchContext/helpers.js
```diff
@@ -93,3 +95,48 @@ function booleanParser(boolean: boolean) {
   // @TODO input validation
   return boolean;
 }
+
+function createPassneger(type: 'adult' | 'infant'): PassengerType {
```
jvaclavik: `createPassneger` :D are you a racist?

FilipMessa: :D 🙃

PR #889, opened 2019-06-14
Fix: fixed passenger form validation for passportID

FilipMessa:


apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
```diff
@@ -40,11 +40,14 @@ class ResultDetailPassenger extends React.Component<Props, State> {
       return 'ID and Name are mandatory';
     }
 
-    const existsID = !!this.props.passengers.find(
-      el => el.passportId === passenger.passportId,
-    );
+    const existsID = this.props.passengers.reduce((reduction, p) => {
```
KarinaDavtyan: What do you think if maybe instead of `p` call it `addedPassenger` or better `prevAddedPassenger`?

jvaclavik: I would rename it this way:
```
this.props.passengers.reduce((passengerAccumulator, currentPassenger)
```

`prevAddedPassenger` is not correct according to me

KarinaDavtyan: Yes it might be incorrect in the scope of reduce
Agree that it should be `validatingPassenger` and instead of `p` it could be just `passenger`

FilipMessa: OK, I will rename `passenger` parameter to `validatingPassenger` and `p` to `passenger`, but I keep `reduction` because it is not an accumulator on the end but it is a reduction to boolean.

jvaclavik: You are right, I didn't notice that it's boolean

PR #903, opened 2019-06-19
Add ModalButtons

jvaclavik:
- `ModalButtons` - Preparation for UX changes in PlacePicker
- Cleaner way how to define default props in RangeDatePicker


packages/universal-components/src/TagsInput/components/TagsContainer.js
```diff
@@ -57,5 +57,7 @@ const styles = StyleSheet.create({
   tag: {
     marginHorizontal: 2,
     alignSelf: 'center',
+    paddingHorizontal: 7,
+    paddingVertical: 4,
```
KarinaDavtyan: Could be design token spaceXXSmall

PR #905, opened 2019-06-20
Refactor date input

KarinaDavtyan:


packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,90 @@
...
+export default class DateInput extends React.Component<Props, State> {
+  state = {
+    day: '',
+    month: '0',
+    year: '',
+  };
+
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    if (this.state !== prevState) {
+      if (Object.values(this.state).every(el => el !== '')) {
```
jvaclavik: this.state.day !== '' && this.state.year !== ''

packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,90 @@
...
+  }
+
+  handleDateValidation = () => {
+    const day = parseInt(this.state.day, 10);
+    const month = parseInt(this.state.month, 10);
+    const year = parseInt(this.state.year, 10);
+
+    let date = new Date();
+
+    if (!day) return false;
```
aalises: `if(!day || !year) return false`

packages/components/src/dateInput/DateInput.js
aalises: Or also:
```
const datesNotEmpty = Object.values(this.state).every(el => el !== '');
const validateYear = this.state.year.length === 4;

if(stateNotNull && validateYear){
 ...
}
```

packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,90 @@
...
+  state = {
+    day: '',
+    month: '0',
+    year: '',
+  };
+
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    if (this.state !== prevState) {
+      if (Object.values(this.state).every(el => el !== '')) {
+        if (this.state.year.length === 4) {
```
jvaclavik: Pass prop `minDate, maxDate` and validate input according it. The same use native pickers.

packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,90 @@
...
+
+export default class DateInput extends React.Component<Props, State> {
+  state = {
+    day: '',
+    month: '0',
+    year: '',
+  };
+
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    if (this.state !== prevState) {
```
jvaclavik: 1. use isEqual
2. 3 conditions can be merged to one

packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,90 @@
...
+    const year = parseInt(this.state.year, 10);
+
+    let date = new Date();
+
+    if (!day) return false;
+    if (!year) return false;
+
+    date = setYear(date, year);
+    date = setMonth(date, month);
+    date = setDate(date, day);
```
aalises: Maybe here you mean to be using `setDay` instead of `setDate` ?

packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,90 @@
...
-    if (month) {
-      this.props.onDateChange(month, 'month');
-    }
+    if (month) this.setState({ month });
   };
 
-  handleYearChange = (year: string) => {
-    this.props.onDateChange(year, 'year');
+  handleYearChange = (input: string) => {
+    const year = input.slice(0, 4);
```
jvaclavik: Why do you use this?

aalises: Why are we also cropping the year by doing `slice` (and also checking somewhere else that the length is indeed 4) ? can it be more than 4 digits? I think this should be handled by the component itself (not allow putting more than 4 digits in the year or precisely 4), but I don't know much how it works.

KarinaDavtyan: Input type number is not handling maxLength as normal input, so this is basically a workaround. I couldnt find any other solution so if you have anything in mind please share :)

packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,90 @@
...
+    const day = parseInt(this.state.day, 10);
+    const month = parseInt(this.state.month, 10);
+    const year = parseInt(this.state.year, 10);
+
+    let date = new Date();
+
+    if (!day) return false;
+    if (!year) return false;
+
+    date = setYear(date, year);
```
aalises: Can we move this into another function? e.g `setDate`.

```
const setDate = (year, month, day) => {
  let date = new Date();

  date = setYear(date, year);
  date = setMonth(date, month);
  date = setDate(date, day);

  return date;
}
```

packages/components/src/dateInput/DateInput.js
KarinaDavtyan: It looks better like this but the whole function logic will probably change

apps/core/scenes/resultDetail/resultDetailContent/ResultDetailPassenger.js
```diff
@@ -36,6 +38,22 @@ class ResultDetailPassenger extends React.Component<Props, State> {
     if (!(validatingPassenger.passportId && validatingPassenger.name)) {
       return 'ID and Name are mandatory';
     }
+    console.log(
```
aalises: `console.log` for testing purposes only plz 🙏

KarinaDavtyan: Dont worry I ll remove it, the condition is not working properly yet, so I kept it

packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,90 @@
...
+    date = setMonth(date, month);
+    date = setDate(date, day);
+    return date;
   };
 
-  handleDayChange = (day: string) => {
-    this.props.onDateChange(day, 'day');
+  handleDayChange = (input: string) => {
+    const day = input.slice(0, 2);
```
aalises: instead of using `2`, `4`... can we have them in constants somewhere such as: `maxDayLength` and `maxYearLength`?

KarinaDavtyan: Alright will add them to config

packages/components/src/dateInput/DateInput.js
KarinaDavtyan: I am not sure about it, as I want to do validation outside of component in order to keep error messages (in this case alert modal) consistent

packages/components/src/dateInput/DateInput.js
KarinaDavtyan: Will do! Thanks

packages/components/src/dateInput/DateInput.js
KarinaDavtyan: Its in the `helpers.js` now, called `composeDateFromStrings`. Maybe name its not so good but I think it explains whats happening well

packages/components/src/dateInput/DateInput.js
KarinaDavtyan: I decided to remove all validation from the component itself, its only checks if its empty or not. Although maybe its not necessary either as if the date wouldnt be composed it would be the same error message

packages/components/src/dateInput/DateInput.js
KarinaDavtyan: I removed `year.length === 4` as its passing check in `ResultDetailPassenger.js`

apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -118,24 +93,25 @@ class PassengerForm extends React.Component<Props, State> {
...
-          date,
         };
       }
     }
-
     return state;
   }
 
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    const notEditedOrChanged =
```
aalises: `const { dateOfBirth, getDateRequest } = this.state`

apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -238,32 +132,38 @@ class PassengerForm extends React.Component<Props, State> {
...
+    } = this.state;
+    const newPassenger = {
+      nationality,
+      passportId,
+      dateOfBirth,
+      gender,
+      name,
+      lastName,
+      bags,
+      id: id || generateId(),
```
aalises: Can you put `??` instead of `||`? if the `id` is 0 it will still be treated as null this way...

packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,107 @@
...
+        month: String(getMonth(date)),
+        year: String(getYear(date)),
+        hasPrefilledState: true,
+      };
+    }
+    return state;
+  }
+
+  componentDidUpdate(prevProps: Props) {
+    if (
```
aalises:
```
const { year, month, day } = this.state
const { onDateChange, getDateRequest } = this.props
```

packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,107 @@
...
-    +year: string,
-  },
-  +errors: {
-    +day: string,
-    +year: string,
-  },
+  +onDateChange: Date => void,
+  +date: ?Date,
+  +editMode: boolean,
+  +getDateRequest: boolean,
```
jvaclavik: It's not clear what it does from the name if it's boolean. Boolean should have that `is|should|has...` prefix

packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,107 @@
...
-    +month: string,
-    +year: string,
-  },
-  +errors: {
-    +day: string,
-    +year: string,
-  },
+  +onDateChange: Date => void,
+  +date: ?Date,
+  +editMode: boolean,
```
jvaclavik: Maybe rename it to `isEditModeEnabled`?

packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,107 @@
...
+const initialState = {
+  day: '',
+  month: '0',
+  year: '',
+  hasPrefilledState: false,
+};
+
+export default class DateInput extends React.Component<Props, State> {
+  state = {
+    ...initialState,
```
jvaclavik: Why not to add it directly? You use initialState only once.

packages/components/src/dateInput/helpers.js
```diff
@@ -0,0 +1,15 @@
...
+export const composeDateFromStrings = (
+  day: string,
+  month: string,
+  year: string,
+): Date => {
+  let date = new Date();
+  date = setDate(date, parseInt(day, 10));
+  date = setMonth(date, parseInt(month, 10));
+  date = setYear(date, parseInt(year, 10));
+  return date;
```
jvaclavik: What about something like this:
```
return new Date(parseFloat(year), parseFloat(month), parseFloat(day))
```

apps/core/components/passengerForm/PassengerForm.js
KarinaDavtyan: I didnt do this change, so probably shouldnt modify

packages/components/src/dateInput/DateInput.js
KarinaDavtyan: Agree! What about then `isEditMode` and `isDateRequested`?

packages/components/src/dateInput/DateInput.js
KarinaDavtyan: Sounds good, will change to that

packages/components/src/dateInput/helpers.js
KarinaDavtyan: Didnt know we can do that, will change to that

apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -60,26 +51,17 @@ const nationalityData = [
...
-    year: '',
-  },
-  inputErrors: {
-    day: '',
-    year: '',
-  },
+  dateOfBirth: null,
   bags: null,
   passengerCount: 1,
+  getDateRequest: false,
```
FilipMessa: This name is little confuse me if something starts with get I expect that is a function.

KarinaDavtyan: Hmm what about `toggleDateRequest` or `dateRequestOn`?

apps/core/components/passengerForm/PassengerForm.js
KarinaDavtyan: fixed

packages/components/src/dateInput/DateInput.js
KarinaDavtyan: fixed

packages/components/src/dateInput/DateInput.js
KarinaDavtyan: Agree, will remove it

apps/core/components/passengerForm/PassengerForm.js
KarinaDavtyan: Changed to `isDateRequested`

packages/components/src/dateInput/DateInput.js
```diff
@@ -2,58 +2,90 @@
...
-      }),
-    );
+type State = {|
+  day: string,
+  month: string,
+  year: string,
+|};
+
+const getMonthOptions = () => {
+  return [...Array(12).keys()].map<Month>(
```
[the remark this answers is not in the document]
KarinaDavtyan: It doesnt seem to work without

FilipMessa:
Just last thing, when the PassengerForm is open there is a pre-filled month field with January. So if a user will fill just day and year. The validation error will appear after submit.

KarinaDavtyan:
> Just last thing, when the PassengerForm is open there is a pre-filled month field with January. So if a user will fill just day and year. The validation error will appear after submit.

Didnt notice this error before, didnt happen before somehow. Will try to fix it

KarinaDavtyan:
@FilipMessa should be fixed now

PR #910, opened 2019-06-28
Add default passenger into the SearchContext

FilipMessa:


apps/core/contexts/searchContext/helpers.js
```diff
@@ -110,10 +110,10 @@ function createPassenger(type: 'adult' | 'infant'): PassengerType {
   });
 }
 
-type PassengersCount = {|
+type PassengersCount = {
```
JoseDuSV: Wouldn't it slightly more "safer" to keep `PassengersCount` exact and pass `{ adults: 1, infants: 0 }` to `createPassengers`.
But it's definitely not crucial and I'm ok even with this solution.

FilipMessa: Good point, I didn't want to be too verbose, but you got right. I will change it.

PR #913, opened 2019-07-02
Add ios date input picker

KarinaDavtyan:
Also refactored PassengerForm so the componentDidUpdated is not needed anymore as DateInput.js would lift state once date composed (not on button click as before).

### DIFF
--- a/apps/core/components/passengerForm/PassengerForm.js
+++ b/apps/core/components/passengerForm/PassengerForm.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react';
-import { View, ScrollView } from 'react-native';
+import { View, ScrollView, Platform } from 'react-native';
 import { generateId } from '@kiwicom/margarita-utils';
 import {
   StyleSheet,
@@ -61,7 +61,6 @@ const initialFormState = {
   dateOfBirth: null,
   bags: null,
   passengerCount: 1,
-  isDateRequested: false,
 };
 
 type State = {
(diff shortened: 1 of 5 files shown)


apps/core/components/passengerForm/PassengerForm.js
```diff
@@ -94,24 +92,20 @@ class PassengerForm extends React.Component<Props, State> {
     // If is in edit mode preload passenger data into the form state
     if (editModeOpened) {
       if (props.prefillData) {
+        const dateOfBirth =
```
aalises: maybe you can destructure `const { prefillData } = props.prefillData`, as it's accessed quite a lot

packages/components/src/dateInput/DateInput.js
```diff
@@ -67,13 +65,15 @@ export default class DateInput extends React.Component<Props, State> {
...
 
-  componentDidUpdate(prevProps: Props) {
-    const { isDateRequested, onDateChange } = this.props;
+  componentDidUpdate(prevProps: Props, prevState: State) {
+    const { onDateChange } = this.props;
     const { day, month, year } = this.state;
-    if (isDateRequested !== prevProps.isDateRequested && isDateRequested) {
-      if (day !== '' && year !== '') {
+    if (day !== prevState.day || year !== prevState.year) {
+      if (day.length === 2 && year.length === 4) {
```
aalises: We had constants for that? `MAX_DAY_LENGTH`, etc...

KarinaDavtyan: Yes we have, forgot about them :(

KarinaDavtyan: fixed

apps/core/components/passengerForm/PassengerForm.js
KarinaDavtyan: fixed, thanks!

